PR #15, opened 2015-02-09
[office] Remove jolla-signonui-service dependecy.

VDVsx:
This dependency is not longer needed since sigon is handle by the qml
code, see change: 32fbfefb81844827e5107842b80510bcf85b918c


rburchell:
LGTM

Issue #10, opened 2014-12-12
Deleting multiple documents at the same time does not work


VDVsx:
DESCRIPTION:
============
When deleting multiple documents at the same time (selecting Delete from the context menu while an another document is already being deleted) is not possible. When the first document gets deleted, the document list gets refreshed and all other pending delete operations cancelled.
 
STEPS TO REPRODUCE:
===================
1) Start Documents
2) Long press on a document
3) Select Delete
4) While the remorse timer is active, long press on an another document
5) Select Delete
 
ACTUAL RESULT:
==============
Only the first document gets deleted because the list refreshes after the document gets deleted and other remorse timers get cancelled

EXPECTED RESULT:
================
Both documents get deleted

REPRODUCIBILITY: 100%

anandrkris:
Sailors - Just wanted to share that we already have 3 reports of this issue.
https://together.jolla.com/question/38050/documents-app-buggy-while-deleting/ >>18 votes
https://together.jolla.com/question/79359/documents-only-one-item-can-be-deleted/ >> 5 votes 
https://together.jolla.com/question/86553/delete-multiple-documents-doesnt-work-as-expected/ >> 2 votes

dcaliste:
Hello Rainemak, thanks for you reply. I've looked to the code, and the delete action comes from models/trackerdocumentprovider.cpp:184 that call removeAt() on the underlying model, which links to models/documentlistmodel.cpp:128. This function indeed call beginRemoveRows() and endRemoveRows() before calling removeAt() on the Qlist storing the DocumentListModelEntry.

So, it seems to me that there is no reset of the model. Would you have any suggestion to look at ?

VDVsx:
Fixed in 1.1.8

PR #18, opened 2015-03-17
Proposed solution to solve issue #10

dcaliste:
The purpose of these two commits is to avoid to call clear() on the model when tracker signal a change in the list of documents.

To do this, the searchFinished() function has been modified to update the current list model instead of clearing it and rebuild it. There are three things to do :
* add new entries (already done by addItem()) ;
* update already existing entries (think new document of a different type, replacing an old one) ;
* remove entries not in tracker list anymore.
The second point is changed in addItem() to update when finding an already existing entry. The third is done via a new boolean in the structure DocumentListModelEntry.

Quickly tested on my device, it solve the issue #10 and avoid in addition the flicking of the list when a document is removed / added.

### DIFF
--- a/models/documentlistmodel.cpp
+++ b/models/documentlistmodel.cpp
@@ -29,6 +29,7 @@ struct DocumentListModelEntry
     QDateTime fileRead;
     QString mimeType;
     DocumentListModel::DocumentClass documentClass;
+    bool dirty; // When true, should be removed from list.
 };
 
 class DocumentListModel::Private
@@ -97,15 +98,32 @@ QHash< int, QByteArray > DocumentListModel::roleNames() const
     return d->roles;
 }
 
+void DocumentListModel::setAllItemsDirty(bool status)
+{
+    for(QList<DocumentListModelEntry>::iterator entry = d->entries.begin();
+        entry != d->entries.end(); entry++) {
+        entry->dirty = status;
+    }
+}
+
 void DocumentListModel::addItem(QString name, QString path, QString type, int size, QDateTime lastRead, QString mimeType)
 {
     // We sometimes get duplicate entries... and that's kind of silly.
-    foreach(const DocumentListModelEntry& entry, d->entries) {
-        if(entry.filePath == path)
+    for(QList<DocumentListModelEntry>::iterator entry = d->entries.begin();
+        entry != d->entries.end(); entry++) {
+        if(entry->filePath == path) {
+            entry->dirty = false;
+            entry->fileType = type;
+            entry->fileSize = size;
+            entry->fileRead = lastRead;
+            entry->mimeType = mimeType;
+            entry->documentClass = static_cast<DocumentClass>(mimeTypeToDocumentClass(mimeType));
             return;
+        }
     }
 
     DocumentListModelEntry entry;
+    entry.dirty    = false;
     entry.fileName = name;
     entry.filePath = path;
     entry.fileType = type;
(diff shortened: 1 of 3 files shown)


models/documentlistmodel.cpp
```diff
@@ -97,15 +98,32 @@ QHash< int, QByteArray > DocumentListModel::roleNames() const
...
-        if(entry.filePath == path)
+    for(QList<DocumentListModelEntry>::iterator entry = d->entries.begin();
+        entry != d->entries.end(); entry++) {
+        if(entry->filePath == path) {
+            entry->dirty = false;
+            entry->fileType = type;
+            entry->fileSize = size;
+            entry->fileRead = lastRead;
+            entry->mimeType = mimeType;
+            entry->documentClass = static_cast<DocumentClass>(mimeTypeToDocumentClass(mimeType));
```
VDVsx: Why we need to compare all fields here ? path should be enough since file system can't have the same path twice, or we don't get full path here ?

models/documentlistmodel.cpp
```diff
@@ -125,6 +143,35 @@ void DocumentListModel::addItem(QString name, QString path, QString type, int si
     endInsertRows();
 }
 
+void DocumentListModel::removeItemsDirty()
+{
+    int start;
+    int index = 0, j;
+
+    start = -1;
```
VDVsx: Could be set already above.

models/documentlistmodel.cpp
```diff
@@ -125,6 +143,35 @@ void DocumentListModel::addItem(QString name, QString path, QString type, int si
...
+                start = -1;
+            }
+        }
+    }
+    if(start>=0) {
+        beginRemoveRows(QModelIndex(), start, index - 1);
+        for (j = start; j < index; j++)
+            d->entries.removeAt(j);
+        endRemoveRows();
+    }
```
VDVsx: Why all of this processing ? it looks to me like a microoptimization that wont be used most of the times, or I'm missing something ? there's no guarantee that entries will be removed consequently. IMO something like this would be enough:

    for (int index=0; index <  d->entries.count(); index++) {
        beginRemoveRows(QModelIndex(), index, index);
        d->entries.removeAt(index);
        endRemoveRows();
   }

models/documentlistmodel.cpp
dcaliste: Thanks for reviewing the patch.

Here, I don't check all fields, the other are affectation. It is to handle the second point in my patch comment : i.e. when an entry change type or other field. For instance I've already a file /path/to/bill.pdf and I save a file to the same path, but it's a new bill.pdf, thus fields must be updated.

models/documentlistmodel.cpp
dcaliste: Ok, I'll change. Sorry, I'm not a C++ developer and I often do mistakes.

models/documentlistmodel.cpp
dcaliste: As you wish. I don't know the implementation of beginRemoveRows() and I was afraid of sending signals to the model that would refresh needlessly. I can remove this also, no problem.

models/documentlistmodel.cpp
VDVsx: Ok, makes sense, I overlooked a bit as well :)

models/documentlistmodel.cpp
VDVsx: The way above it will send signals only when a entry will be removed, please note I missed a if there, will update.

dcaliste:
Ok, I've updated the pull request to follow the changes proposed by @VDVsx. Do you think it is ok for merging now ?

VDVsx:
Thanks now looks good to me, could you squash last commit since is not relevant here anymore ? (http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)
Also when doing the squash modify last commit to have a prefix like: [office] YOUR_COMMIT message.
The [...] prefix is needed by our integration to auto-generate changelogs.
I'll give it a quick test as well and then when commits are in order I can merge.

Thanks for the contribution :)

VDVsx:
Tested on device and works fine. so once commits are fixed I'll merge and tag.

dcaliste:
Ok, I've squashed the commits together. Patch may be ready to merge. And thank you again for reviewing and giving advices.

Issue #21, opened 2015-03-23
Crash when opening a broken PDF file


dcaliste:
If a PDF file is invalid (half downloaded from the web, resulting of a wrong rename, generated by `touch foo.pdf`…), tapping on the name results in the application crash.

pvuorela:
I think this should be working a lot better already. Closing.

Issue #1, opened 2014-06-10
Ungraceful error on attempting to open a password protected document


stephgosling:
Reported here: https://together.jolla.com/question/44849/bug-documents-app-error-message/ the text of which is:

I received an email with a password-protected .docx file attached. I didn't at the time realise the file was 'protected' and downloaded/opened the attachment anyway. That failed (which is fine, not expecting that support) but the error was rather nasty. For what it's worth, it wasn't orientation-aware so it's lucky that the 'OK' was visible so I could 'click' it. 

I'm assuming this is in Documents rather than the mail client, if I'm wrong please close :)

![20140606150014](https://cloud.githubusercontent.com/assets/1391834/3236885/edad514c-f0e1-11e3-9fbc-a567a623adda.png)

VDVsx:
Fix in version 1.1.9. thanks to dcaliste.

PR #22, opened 2015-03-23
Notify user on broken PDF files

dcaliste:
Proposed patches to avoid crashes when tapping on a broken PDF file (issue #21).

I've :
- protected all calls to attributes of Poppler::Document when it is a null pointer because of loading failure at pdfjob.cpp:31.
- added a flag in PdfRenderThread that becomes true when file loading results in a null pointer and expose it read only in PdfDocument.
- used this flag in QML to return to document list when tapping on a broken PDF file and notify the user.

What do you think ? I don't like much the fact that the page for the PDF file is pushed and then popped immediately, but I don't know what to show otherwise. On the other side, that would be nice to still have the drawer at hand even for broken PDF files, because we may want to forward them by email for investigation…

### DIFF
--- a/pdf/pdfrenderthread.cpp
+++ b/pdf/pdfrenderthread.cpp
@@ -77,6 +77,7 @@ public:
 
     Thread *thread;
 
+    bool loadFailure;
     Poppler::Document* document;
     PDFTocModel* tocModel;
 
@@ -143,7 +144,7 @@ PDFRenderThread::~PDFRenderThread()
 int PDFRenderThread::pageCount() const
 {
     QMutexLocker locker{ &d->thread->mutex };
-    return d->document->numPages();
+    return (d->document != nullptr) ? d->document->numPages() : 0;
 }
 
 QObject* PDFRenderThread::tocModel() const
@@ -156,6 +157,11 @@ bool PDFRenderThread::isLoaded() const
 {
     QMutexLocker{ &d->thread->mutex };
     return d->document != nullptr;
+}
+bool PDFRenderThread::isFailed() const
+{
+    QMutexLocker{ &d->thread->mutex };
+    return d->loadFailure;
 }
 
 QMultiMap< int, QPair< QRectF, QUrl > > PDFRenderThread::linkTargets() const
@@ -214,8 +220,14 @@ void PDFRenderThreadQueue::processPendingJob()
         return;
 
     PDFJob* job = dequeue();
-    if (job->type() != PDFJob::LoadDocumentJob)
+    switch(job->type()) {
+    case PDFJob::LoadDocumentJob:
+        d->loadFailure = false;
+        break;
+    default:
         job->m_document = d->document;
+        break;
+    }
     locker.unlock();
 
     job->run();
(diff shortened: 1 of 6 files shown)


pdf/pdfrenderthread.cpp
```diff
@@ -143,7 +144,7 @@ PDFRenderThread::~PDFRenderThread()
 int PDFRenderThread::pageCount() const
 {
     QMutexLocker locker{ &d->thread->mutex };
-    return d->document->numPages();
+    return (d->document != nullptr)?d->document->numPages():0;
```
VDVsx: Add extra spaces here between ? and : conditions for better readability.

pdf/pdfrenderthread.cpp
```diff
@@ -214,8 +220,14 @@ void PDFRenderThreadQueue::processPendingJob()
...
-    if (job->type() != PDFJob::LoadDocumentJob)
-        job->m_document = d->document;
+    switch(job->type())
+    {
+        case PDFJob::LoadDocumentJob: {
+            d->loadFailure = false;
+        }
+        default:
+            job->m_document = d->document;
+    }
```
VDVsx: This project is probably not following the Qt conventions in other places, but we aim follow them, see here: http://qt-project.org/wiki/Qt_Coding_Style#e289ee44592e9c32d4212069f0806daf

plugin/PDFDocumentPage.qml
```diff
@@ -55,4 +61,11 @@ DocumentPage {
         interval: 5000;
         onTriggered: linkArea.sourceSize = Qt.size( base.width, pdfCanvas.height );
     }
+
+    Notification {
+        id: notification;
+        category: "x-sailfish.sailfish-office.error";
+        previewBody: "Cannot open PDF file";
+        previewSummary: "Broken PDF file";
+    }
```
VDVsx: This need to be translatable:

//% "Cannot open PDF file"
previewBody:qsTrId(... ;

check other translations in this project to match the same naming.

rpm/sailfish-office.spec
```diff
@@ -44,6 +44,7 @@ Group: System/Base
 %{_datadir}/translations/*.qm
 %{_datadir}/dbus-1/interfaces/
 %{_datadir}/dbus-1/services/
+%{_datadir}/lipstick/notificationcategories/x-sailfish.sailfish-office.error.conf
```
VDVsx: For in app notifications this is not needed, we don't want the notification to show up in events view.

plugin/PDFDocumentPage.qml
```diff
@@ -44,6 +45,11 @@ DocumentPage {
     PDF.Document {
         id: pdfDocument;
         source: base.path;
+        onDocumentLoaded: if (failure) {
+            pageStack.completeAnimation();
+            pageStack.pop();
+            notification.publish();
```
VDVsx: This part I need to test on device to see if behavior can be improved, will take a bit more.

dcaliste: Yep, that's great of you. Currently the page is pushed and then the PDF document is loaded, so I don't know how not to push the page without being too intrusive in the code.

Besides, this may be irrelevant not to push the page because, I'm preparing a kind of "place holder" for PDF without rendering (because of broken file or because the file is valid but currently locked down). This place holder would replace the PDFView Item and display a message that the PDF is either broken or protected. In case of protected file a button would open a dialog with a text input to enter the password (related to some discussion in issue #1). Having this place holder would keep the actions of the drawer available, like sharing the file by bluetooth…

What do you think ?

dcaliste:
I've taken into account your remarks and rebased my two commits hopefully in a better shape :
* Removed the x-sailfish.sailfish-office.error.conf file ;
* Try to follow the Qt coding rules for switches and add spaces for better readability ;
* Use translatable strings for the notification.

Beside, how should I handle the translations ? At least in English ?

VDVsx:
You just need the engineering English like is done now, Jolla translation team does the rest.

plugin/PDFDocumentPage.qml
VDVsx: Sounds good since it covers additional cases.

VDVsx:
I'm merging this already to facilitate your work, then can tag later after next commit that addresses the QML side like discussed above.

dcaliste:
Thank you.

PR #23, opened 2015-03-27
[sailfish-office] Add the ability to unlock password encrypted PDF documents

dcaliste:
I've wrapped thePoppler::Document::unlock() method in Pdf::Document (through a call in background) and add a requestUnLock() method for Pdf::Document objects.

The UI has been updated in accordance to be able to enter password for encrypted PDF files. This is done through a ViewPlaceholder in the PDFView object.

The previously introduced notification for broken PDF has been removed in favour of this new ViewPlaceholder.

This pull request partially address issue #1 for PDF documents.

### DIFF
--- a/pdf/pdfdocument.cpp
+++ b/pdf/pdfdocument.cpp
@@ -43,6 +43,7 @@ PDFDocument::PDFDocument(QObject* parent)
     d->thread = new PDFRenderThread{ this };
     connect( d->thread, &PDFRenderThread::loadFinished, this, &PDFDocument::documentLoaded );
     connect( d->thread, &PDFRenderThread::loadFinished, this, &PDFDocument::pageCountChanged );
+    connect( d->thread, &PDFRenderThread::loadFinished, this, &PDFDocument::loadFinished );
     connect( d->thread, &PDFRenderThread::jobFinished, this, &PDFDocument::jobFinished );
 }
 
@@ -151,6 +152,11 @@ void PDFDocument::requestPageSizes()
     d->thread->queueJob( job );
 }
 
+void PDFDocument::loadFinished()
+{
+  if (d->thread->isFailed())
+    emit documentFailed();
+}
 void PDFDocument::jobFinished(PDFJob* job)
 {
     switch(job->type()) {
--- a/pdf/pdfdocument.cpp
+++ b/pdf/pdfdocument.cpp
@@ -82,6 +82,11 @@ bool PDFDocument::isFailed() const
     return d->thread->isFailed();
 }
 
+bool PDFDocument::isLocked() const
+{
+    return d->thread->isLocked();
+}
+
 PDFDocument::LinkMap PDFDocument::linkTargets() const
 {
     return d->thread->linkTargets();
@@ -120,9 +125,18 @@ void PDFDocument::setSource(const QString& source)
     }
 }
 
+void PDFDocument::requestUnLock(const QString& password)
+{
+    if (!isLocked())
+        return;
+
+    UnLockDocumentJob* job = new UnLockDocumentJob(password);
+    d->thread->queueJob(job);
+}
+
 void PDFDocument::requestPage(int index, int size, QQuickWindow *window )
 {
-    if(!isLoaded())
+    if(!isLoaded() || isLocked())
         return;
 
     RenderPageJob* job = new RenderPageJob{ index, size, window };
@@ -131,21 +145,21 @@ void PDFDocument::requestPage(int index, int size, QQuickWindow *window )
 
 void PDFDocument::prioritizeRequest(int index, int size)
 {
-    if (!isLoaded())
+    if (!isLoaded() || isLocked())
         return;
     d->thread->prioritizeJob(index, size);
 }
 
 void PDFDocument::cancelPageRequest(int index)
 {
-    if (!isLoaded())
+    if (!isLoaded() || isLocked())
         return;
     d->thread->cancelRenderJob(index);
 }
 
 void PDFDocument::requestPageSizes()
 {
-    if(!isLoaded())
+    if(!isLoaded() || isLocked())
         return;
 
     PageSizesJob* job = new PageSizesJob{};
@@ -154,12 +168,19 @@ void PDFDocument::requestPageSizes()
 
 void PDFDocument::loadFinished()
 {
-  if (d->thread->isFailed())
-    emit documentFailed();
+    if (d->thread->isFailed())
+        emit documentFailed();
+    if (d->thread->isLocked())
+        emit documentLocked();
 }
 void PDFDocument::jobFinished(PDFJob* job)
 {
     switch(job->type()) {
+        case PDFJob::UnLockDocumentJob: {
+            emit documentLocked();
+            emit pageCountChanged();
+            break;
+        }
         case PDFJob::RenderPageJob: {
             RenderPageJob* j = static_cast<RenderPageJob*>(job);
             emit pageFinished(j->m_index, j->m_page);
--- a/pdf/pdfjob.cpp
+++ b/pdf/pdfjob.cpp
@@ -35,6 +35,20 @@ void LoadDocumentJob::run()
     }
 }
 
+UnLockDocumentJob::UnLockDocumentJob(const QString& password)
+    : PDFJob(PDFJob::UnLockDocumentJob), m_password(password)
+{
+
+}
+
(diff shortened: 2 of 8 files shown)


pdf/pdfdocument.cpp
```diff
@@ -119,9 +125,18 @@ void PDFDocument::setSource(const QString& source)
     }
 }
 
+void PDFDocument::requestUnLock(const QString& password)
+{
+  if (!isLocked())
+      return;
+
+  UnLockDocumentJob* job = new UnLockDocumentJob{ password };
```
VDVsx: Why these curly brackets here ? normally square ones are used "("

plugin/PDFDocumentPage.qml
```diff
@@ -40,19 +39,60 @@ DocumentPage {
...
+            enabled: pdfDocument.failure || pdfDocument.locked;
+            y: (flickable ? flickable.originY : 0)
+                   + (base.height - height - (passwd.visible ? passwd.height : 0)) / 2
+            text: if (pdfDocument.failure) {
+                //% "Broken file"
+                return qsTrId("sailfish-office-me-broken-pdf");
+            } else {
+                //% "Password protected file"
+                return qsTrId("sailfish-office-me-locked-pdf");
+            }
```
VDVsx: These if/else could be replace with:

pdfDocument.failure ? qsTrId("sailfish-office-me-broken-pdf") : qsTrId("sailfish-office-me-locked-pdf")

Also please avoid using ending punctuation in qml code, is not needed at all, just produces more noise.

pdf/pdfdocument.cpp
dcaliste: Well, I've just copied what's done for the other object creation in the file, see for instance on line 104. What do you think ?

plugin/PDFDocumentPage.qml
dcaliste: Ok, I've removed the trailing ";" in QML lines I've wrote myself. They are still on the original lines of the file to keep a commit just related to the matter. I've also used the inlined "if" for the strings, as you suggested.

pdf/pdfdocument.cpp
VDVsx: Ouch, that's extremely confusing IMO, not sure what the author was trying to achieve here, since all of these objects have defined constructors, anyway let's stick to the common usage and use simple braces like in other Qt code, also let's avoid no meaningful whitespaces around the braces, for example: d->thread->queueJob(job), also see: http://qt-project.org/wiki/Qt_Coding_Style#1d8ea7b3bfc87950e27892e5eac9c6f0

plugin/PDFDocumentPage.qml
```diff
@@ -40,19 +39,49 @@ DocumentPage {
...
 
         onClicked: base.open = !base.open;
+
+        ViewPlaceholder {
+            enabled: pdfDocument.failure || pdfDocument.locked
+            y: (flickable ? flickable.originY : 0) + (base.height - height - (passwd.visible ? passwd.height : 0)) / 2
+            //% "Broken file" : "Locked file"
+            text: pdfDocument.failure ? qsTrId("sailfish-office-me-broken-pdf") : qsTrId("sailfish-office-me-locked-pdf")
+            //% "Cannot read the PDF document" : "Enter password to unlock"
+            hintText: pdfDocument.failure ? qsTrId("sailfish-office-me-broken-pdf-hint") : qsTrId("sailfish-office-me-locked-pdf-hint")
```
VDVsx: Not sure if such syntax for translations works here normally it goes like:
     //% "Broken file"
    text: pdfDocument.failure ? qsTrId("sailfish-office-me-broken-pdf") : 
    //% "Locked file"
    qsTrId("sailfish-office-me-locked-pdf")

dcaliste: Ok, understood, commit is updated. Sorry being newby for these things…

pdf/pdfdocument.cpp
dcaliste: I've used normal braces in object creation and removed unnecessary white spaces. I've also corrected some wrong indentation levels in other places of the patch.

pdf/pdfjob.cpp
```diff
@@ -35,6 +35,20 @@ void LoadDocumentJob::run()
     }
 }
 
+UnLockDocumentJob::UnLockDocumentJob(const QString& password)
+    : PDFJob{ PDFJob::UnLockDocumentJob }, m_password{ password }
```
VDVsx: These should also become normal brackets.

pdf/pdfrenderthread.cpp
```diff
@@ -163,6 +169,11 @@ bool PDFRenderThread::isFailed() const
     QMutexLocker{ &d->thread->mutex };
     return d->loadFailure;
 }
+bool PDFRenderThread::isLocked() const
+{
+    QMutexLocker{ &d->thread->mutex };
```
VDVsx: Same here.

pdf/pdfrenderthread.cpp
```diff
@@ -240,28 +251,32 @@ void PDFRenderThreadQueue::processPendingJob()
...
+                delete d->document;
+            if(d->tocModel) {
+                d->tocModel->deleteLater();
+                d->tocModel = nullptr;
+            }
+    
+            d->document = dj->m_document;
+            if(d->document) {
+                if (!d->document->isLocked()) {
+                    d->tocModel = new PDFTocModel{ d->document };
```
VDVsx: Same.

VDVsx:
Apart from code style, rest LGTM, but did no test yet.

dcaliste:
Thank you @VDVsx I've changed the curly braces to normal ones where you mentioned. Test whenever you can and tell me if there's something wrong (UI wise or in the functionality).

VDVsx:
Password stuff seems to work, however it prints this error when opening doc: [D] Poppler::Debug::qDebugDebugFunction:42 - "Error: Incorrect password" 

Broken file does not seem to work here, blank page is open here: Poppler::Debug::qDebugDebugFunction:42 - "Error (303900): Illegal character '>'"

Probably depends on the kind of corruption ?

dcaliste:
Thank for testing. The message of Poppler for locked document appears when one loads the document without providing password (which is the case when tapping on the document) before later unlock it by providing the password. So that's normal.

For the second error with broken documents, I've only tested with empty PDF file (touch foo.pdf) :( Maybe I should be more diligent. In fact, I've noticed the bug of empty PDF files when I was testing the multiple deletion patch by creating many PDFs that I can delete easily… The current patch for broken files only handles the case when Poppler::Document::load() fails and return a NULL pointer. If I look to the sources (http://bazaar.launchpad.net/~vcs-imports/poppler/master/view/head:/qt5/src/poppler-document.cc line 82), it depends on the function isOk() of Poppler::PDFDoc. If this returns false (and document is not encrypted), then Poppler::Document::load() will return NULL and this case is handled (or should be) by the proposed patch. I guess Poppler::PDFDoc::isOk() returns false only for unloadable documents but not documents with rendering problems…

VDVsx:
Ok, merging this one, since current issues do not really cause a bad failure and were already present in the code before this PR.

anandrkris:
Thanks folks. Love FOSS applications. :heart_eyes: 
Waiting for the app update now. Should we wait for OS update? :wink: 
Even last week I received an important pdf document (*new job offer*) with password and had to wait till I get home to see the details. This is much desired functionality indeed.
https://together.jolla.com/question/83800/jolla-apps-make-updates-independent-of-sailfishos-updates/

VDVsx:
If you don't want to wait for the OS update you can install this package yourself, does not have any external dependencies.

anandrkris:
@VDVsx - Thanks for responding. Will give it a try. Been using Linux for couple of years guess should be able to figure out how to build / install...

dcaliste:
@anandrkris Hello, you can generate the RPM with the MerSDK vbox, that's simple. Then, the rpm has a dependency on a more recent version of sailffish-silica than what is available on 1.1.2.x. It's not really an issue, force the upgrade of sailfish-office with --nodeps option of the rpm command. Then after installation, edit the file /usr/share/sailfish-office/Main.qml and comment line 94 (_defaultLabelFormat attriute has been introduced in more recent sailfish-silica than what we have).

anandrkris:
@dcaliste - Thanks for the details. Adding this to my weekend tasks.

PR #24, opened 2015-04-03
[sailfish-office] Filter the document list with a search field

dcaliste:
This pull request adds the capability to filter the document list to display a subpart of all documents on the phone. The filtering is done on file names only. The model is filtered when matching a subpart of the filename. This is convenient for instance to find a specific document, like when look for electricity bills or whatever.

I've added a QSortFilterProxyModel on the cpp side that works on DocumentListModel object. It is exported as a QML type "FilterModel".

I've added a SearchPageHeader QML component, inspired by the one of the media player (for UI consistency). This component displays a PageHeader or a SearchField. The SearchField is activated through a pulley menu.

What do you think ?

### DIFF
--- a/models/filtermodel.h
+++ b/models/filtermodel.h
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2015 Caliste Damien.
+ * Contact: Damien Caliste <dcaliste@free.fr>
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; version 2 only.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+#ifndef FILTERMODEL_H
+#define FILTERMODEL_H
+
+#include <QSortFilterProxyModel>
+
+#include "documentlistmodel.h"
+
+class FilterModel : public QSortFilterProxyModel
+{
+    Q_OBJECT
+    Q_PROPERTY(DocumentListModel *sourceModel READ sourceModel WRITE setSourceModel NOTIFY sourceModelChanged)
+public:
+    FilterModel(QObject* parent = 0);
+    ~FilterModel();
+
+public:
+    DocumentListModel* sourceModel() const;
+
+public Q_SLOTS:
+    void setSourceModel(DocumentListModel *model);
+
+Q_SIGNALS:
+    void sourceModelChanged();
+};
+
+#endif // FILTERMODEL_H
--- a/qml/FileListPage.qml
+++ b/qml/FileListPage.qml
@@ -23,17 +23,23 @@ import Sailfish.Office.Files 1.0
 
 Page {
     id: page
-    property alias model: listView.model;
+    property alias model: filteredModel.sourceModel;
     property string title: "";
     property QtObject provider;
 
     allowedOrientations: Orientation.All;
+
+    FilterModel {
+        id: filteredModel
+        filterRegExp: RegExp("")
+    }
 
     SilicaListView {
         id: listView;
         anchors.fill: parent
 
         currentIndex: -1;
+        model: filteredModel
 
         children: ScrollDecorator { }
         header: PageHeader {
--- a/qml/FileListPage.qml
+++ b/qml/FileListPage.qml
@@ -25,13 +25,14 @@ Page {
     id: page
     property alias model: filteredModel.sourceModel;
     property string title: "";
+    property string searchText: "";
     property QtObject provider;
 
     allowedOrientations: Orientation.All;
 
     FilterModel {
         id: filteredModel
-        filterRegExp: RegExp("")
+        filterRegExp: RegExp(searchText, "i")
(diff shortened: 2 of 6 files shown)


models/filtermodel.h
```diff
@@ -0,0 +1,44 @@
...
+#include <QSortFilterProxyModel>
+
+#include "documentlistmodel.h"
+
+class FilterModel : public QSortFilterProxyModel
+{
+    Q_OBJECT
+    Q_PROPERTY(DocumentListModel *sourceModel READ sourceModel WRITE setSourceModel NOTIFY sourceModelChanged)
+public:
+    FilterModel( QObject* parent = 0 );
```
VDVsx: these space could go way to be consistent with rest of code in this file.

qml/SearchPageHeader.qml
```diff
@@ -0,0 +1,67 @@
...
+            visible: searchField.opacity === 0.
+        }
+
+        SearchField {
+            id: searchField
+            property bool open: opacity === 1.0
+
+            width: parent.width
+            enabled: false
+            opacity: (enabled) ? 1.0 : 0.0
```
VDVsx: The parenthesis are redundant here.

qml/FileListPage.qml
```diff
@@ -23,26 +23,59 @@ import Sailfish.Office.Files 1.0
...
             //: Application title
             //% "Documents"
             title: qsTrId("sailfish-office-he-apptitle");
-
             // TODO: uncomment once there are more document sources
             // title: page.title;
+
+            //: Document search field placeholder text
+            //% "Search documents"
+            placeholderText: qsTrId("sailfish-office-tf-search-documents")
```
VDVsx: guess this one could go inside the search component itself, I don't see a need to set it here and create the property alias below.

VDVsx:
Filter branch does not compile here, seems due to unrelated changes that got merged here somehow:

/home/valerio/temp_devel/sailfish-office/pdf/pdfrenderthread.cpp: In member function 'void PDFRenderThreadQueue::processPendingJob()':
/home/valerio/temp_devel/sailfish-office/pdf/pdfrenderthread.cpp:262:5: error: jump to case label [-fpermissive]
/home/valerio/temp_devel/sailfish-office/pdf/pdfrenderthread.cpp:244:26: error:   crosses initialization of 'LoadDocumentJob* dj'
/home/valerio/temp_devel/sailfish-office/pdf/pdfrenderthread.cpp:242:11: warning: enumeration value 'RenderPageJob' not handled in switch [-Wswitch]
/home/valerio/temp_devel/sailfish-office/pdf/pdfrenderthread.cpp:242:11: warning: enumeration value 'PageSizesJob' not handled in switch [-Wswitch]
make[2]: *** [pdf/CMakeFiles/sailfishofficepdfplugin.dir/pdfrenderthread.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
[ 88%] Building CXX object pdf/CMakeFiles/sailfishofficepdfplugin.dir/pdfcanvas.cpp.o
make[1]: *** [pdf/CMakeFiles/sailfishofficepdfplugin.dir/all] Error 2
make: *** [all] Error 2
error: Bad exit status from /var/tmp/rpm-tmp.oq3d5G (%build)


You should base your new commits on master and avoid several unrelated things together.

Code above looks good, I would like to test it, would be also nice to have highlight when matching the names.

qml/FileListPage.qml
dcaliste: I don't know. I've followed how this is done in the media player. I guess the SearchHeaderPage component there has a vocation to be public one time, or at least reusable from other QML code, so the labels and place holder texts stays on the user side instead of in the component itself. I've found a similar SearchHeaderPage component in the Sailfish/Silica/Pickers directory for instance. So I've currently duplicated the SearchHeaderPage code, trying to keep it generic in case, at one point, one can switch it with one provided in Silica…

Besides, I can move this text to the component itself if you think it's clearer for now.

dcaliste:
Oups, sorry for the wrong base. I've tried to be tidy by creating branches of my repo to develop various things in separate branches, but didn't fully succeed at one point…

I've rebased the filter branch on master, and it should compile now. Tell me if there's still something wrong.

dcaliste:
For the highlight, I agree that it's a must. I didn't think about it, but now that you mention it, that's mandatory to send a feedback to the user. I'll try to find how this is implemented in the media player or somewhere else…

lukedirtwalker:
@dcaliste for the highlight I have an example in CuteSpot: https://github.com/lukedirtwalker/cutespotify/blob/sailfish/qml/TracklistPage.qml#L68
I think the same example is also in the components gallery ;)

qml/FileListPage.qml
VDVsx: I you prefer the text to be moved, since atm there's no use to have it here and it just generates extra unnecessary code. As for the header stuff that looks fine.

dcaliste:
@lukedirtwalker thanks a lot for the information, it has saved time digging here and there to find this.

dcaliste:
I've added the highlight stuff when searching and did the small corrections @VDVsx suggested. You can test and give me your feedbacks.

qml/SearchPageHeader.qml
```diff
@@ -0,0 +1,70 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+
```
VDVsx: This file is not installed, so a packaged produced with mb tool will not work.

dcaliste: Stupid me, corrected !

VDVsx:
Test and works fine. Merging.

Issue #17, opened 2015-02-27
Inconsistencies when changing orientation


MikErk:
Hello, 

I am just starting to get some insights into sailfish programming and found a bug relating to landscape/portrait orientation. During pdf viewing when I change the orientation it switches to a different page number. In my case I tested with a 120-page PDF: In landscape mode when I am look at page 2, then in portrait mode his already on page 5/6. The gap increases when I go further down the document. I am guessing that it might be that somehow you only look at the current scroller position when the orientation is changed. I have looked in some source but I wasn't able to find the function where this is handled.

I posted a bug info already on together.jolla: https://together.jolla.com/question/83724/portraitlandscape-mode-inconsistent-in-sailfish-office/

anandrkris:
I experienced this issue today. I was reading a PDF document and navigated to in Appendix *(no bookmarks on PDF sadly :cry: )* in portrait mode and had to change orientation as there was lot of text. And the page focus was gone! I hate to download Android apps so would be grateful for fix soon.
Also, am curious why **Jolla-made** apps are released only with OS updates. Can't we have updates through harbor store?

anandrkris:
@dcaliste - Tagging - Just In case you're interested to fix :smile:

sebix:
Isn't this issue fixed now with #25 merged? Can't observe it with 1.1.7.28

dcaliste:
Yes, it should be OK now, but when testing my patch, I've observed similar strange behaviours that I cannot reproduce later. So hopefully I would say, it's closed, but it's up to the maintainer to decide !

PR #25, opened 2015-04-10
[office] Adjust content position when device is rotated and zoom level changes.

dcaliste:
This pull request addresses the issue #17. This is done in the two on{Width,Height}Changed signal handlers of the PDFView component.

It also corrects an other inconsistency in the zoom level when device is rotated :
* open a PDF in portrait mode ;
* rotate the device, the width is adjusted to the new screen width, that's fine.
But :
* return to portrait mode (width is again adjusted, fine) ;
* zoom in a bit, some pixels ;
* rotate the device, the zoom level is conserved and the view is not anymore within [width;2.5 × width]
Any aditional zoom change will jump into this range.
This is due to the fact that the zoom function remove the binding between pdfCanvas.width and base.width and the [width;2.5 × width] constraint is only applied on zoom.

The rest of the pull request correct this by factoring the constraint into a clamp() function that is applied either on zoom or when the base.width is changed.

### DIFF
--- a/plugin/PDFView.qml
+++ b/plugin/PDFView.qml
@@ -30,39 +30,45 @@ SilicaFlickable {
     property alias itemHeight: pdfCanvas.height;
     property alias document: pdfCanvas.document;
 
-    property bool scaled: false;
+    property bool scaled: pdfCanvas.width != width;
 
     signal clicked();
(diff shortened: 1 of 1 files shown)


VDVsx:
LGTM

PR #26, opened 2015-04-14
[office] correct an issue of content jumping when zooming near the end of long PDF files.

dcaliste:
There is an issue with PDF files with many pages when zooming in or out near the end of the file. To reproduce :
* open a PDF file with many pages like http://www.free.fr/adsl/pdf/cgv-forfait-hors-opt-11072007.pdf ;
* jump to the last page and stay at the bottom;
* zoom in and out, from time to time, the view is jumping up and down.

This is due to two factors in the code :
* the zooming factor is not the same for the width and the height. Indeed, the height is cast to an int in pdfcanvas.cpp, so a rounding appears here ;
* the fact of affacting contentX to a new value in PDFView.qml zoom() function makes contentY changes also (I guess because height has changed and the view is adjusting its contentY accordingly to a value that makes the bottom part of the document stays at the bottom of the screen), one cannot use the += affectation that supposes that contentY contains the old value.

In the patch, I've addressed the first issue by using newWidth / oldWidth for width and newHeight / oldHeight for height instead of width factor for height also. The second issue is addressed by storing old values of contentX and contentY before affecting them.

### DIFF
--- a/plugin/PDFView.qml
+++ b/plugin/PDFView.qml
@@ -49,22 +49,28 @@ SilicaFlickable {
 
     function zoom(amount, center) {
         var oldWidth = pdfCanvas.width;
+        var oldHeight = pdfCanvas.height
+        var oldContentX = contentX
+        var oldContentY = contentY
 
         pdfCanvas.width = clamp(pdfCanvas.width * amount);
 
-        var realZoom = pdfCanvas.width / oldWidth;
-        contentX += (center.x * realZoom) - center.x;
-        contentY += (center.y * realZoom) - center.y;
+        /* One cannot use += here because changing contentX will change contentY
+           to adjust to new height, so we use saved values. */
+        contentX = oldContentX + (center.x * pdfCanvas.width / oldWidth) - center.x
+        contentY = oldContentY + (center.y * pdfCanvas.height / oldHeight) - center.y
     }
 
     function adjust() {
         var oldWidth = pdfCanvas.width
         var oldHeight = pdfCanvas.height
+        var oldContentX = contentX
+        var oldContentY = contentY
 
         pdfCanvas.width = scaled ? clamp(pdfCanvas.width) : width
 
-        contentX *= pdfCanvas.width / oldWidth
-        contentY *= pdfCanvas.height / oldHeight
+        contentX = oldContentX * pdfCanvas.width / oldWidth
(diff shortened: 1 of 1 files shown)


VDVsx:
Need rebase now that previous commit went in

dcaliste:
Ok, I've rebased the commit.

Besides, I've noticed a corner case of the previous pull request. In a specific case, it does not correct the issue that it is supposed to solve. This corner case appears when the document is zoomed and the device is rotated. If the zoom level is clamped by the rotation, then the content adjustment with onWidthChanged and onHeightChanged of the PDFCanvas object is not triggered because in the meantime scaled becomes false because of the clamping.

I've corrected this also in this pull request. I can separate the two commits into two pull requests if you prefer but it may be overkill. What do you think ?

VDVsx:
It's fine, but I'm getting some weird issues when rotating and then zoom in and zoom out in portrait, pdf document goes out of screen(background is shown and half of the pdf), can you reproduce that ? might be some problem somewhere since I tried in a devel SW version.

dcaliste:
I'll try some more testing to see if I can reproduce. Currently, I'm testing on a 63 pages pdf file the following steps :
* open in portrait
* scroll to the end (with the scroll button)
* move up and down a bit
* zoom in and out
* rotate, check that top left corner of the view keep its position
* move up and down a bit and zoom in and out again a bit
* rotate again, check that left corner is fine
* …

When zooming in and out, I try :
* no zoom before rotating ;
* zoom below clamp value for landscape orientation and rotate ;
* zoom in range of zoom possibility for landscape orientation and rotate ;
* zoom more than rotation allow when going from landscape to portrait.

I agree that these zooming issues are annoying and should be smashed out before merging. I'll test further.

VDVsx:
Any progress here ? 
I'm afraid I would need to revert previous patch, to make the app promotable, still need to see if it's better with our without it (i.e no regressions caused)

dcaliste:
Sorry, I notice nothing strange since the previous week. I agree to revert as you said so other modifications can be published soon. I'll continue look on this and report when I have new information.

VDVsx:
Just tried and can't reproduce anymore, so might be another bug that was around about multitouch. Will merge and see if testing can reproduce, if still there we can always take previous version.

dcaliste:
Thanks, I agree.

PR #30, opened 2015-05-05
[office] Add a placeholder for the table of content of PDF documents.

dcaliste:
This adds a message for empty index page for PDF documents instead of showing an empty page.

The enabled attribute of the placeholder is not set by binding since the tocmodel object comes from a different thread and QML complains about signaling from a different thread. Thus, enabled attribute is set by javascript by reading the count value when model is set instead.

### DIFF
--- a/plugin/PDFDocumentToCPage.qml
+++ b/plugin/PDFDocumentToCPage.qml
@@ -35,6 +35,17 @@ Page {
         //% "Index"
         header: PageHeader { title: qsTrId( "sailfish-office-he-pdf_index" ); }
 
+        ViewPlaceholder {
+            id: placeholder
+            //% "Document has no table of content"
+            text: qsTrId("sailfish-office-me-no-toc")
+        }
+        // The enabled attribute of the placeholder is not set by binding since
+        // the model object comes from a different thread and QML cannot listen
+        // on signals from a different thread. Thus, the attribute is set by
+        // reading the count value instead of a binding.
+        onModelChanged: placeholder.enabled = !model || (model.count == 0)
+
         delegate: BackgroundItem {
             id: bg;
 
--- a/pdf/pdftocmodel.cpp
+++ b/pdf/pdftocmodel.cpp
@@ -145,3 +145,8 @@ int PDFTocModel::rowCount(const QModelIndex& parent) const
         return 0;
     return d->entries.count();
 }
+
+int PDFTocModel::count() const
+{
+  return d->entries.count();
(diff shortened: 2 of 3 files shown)


plugin/PDFDocumentToCPage.qml
```diff
@@ -35,6 +35,13 @@ Page {
         //% "Index"
         header: PageHeader { title: qsTrId( "sailfish-office-he-pdf_index" ); }
 
+        ViewPlaceholder {
+            id: placeholder
+            //% "Document has no table of content"
+            text: qsTrId("sailfish-office-me-no-toc")
+        }
+        onModelChanged: placeholder.enabled = !model || (model.count == 0)
```
VDVsx: Is this really needed here ? "enabled = !model || (model.count == 0)", should be enough inside above item.

dcaliste:
This is strange, I agree. I put in the PR comment that the enabled attribute of the placeholder is not set by binding since the tocmodel object comes from a different thread and QML complains about signaling from a different thread. Thus, enabled attribute is set by javascript by reading the count value when model is set instead.

Maybe a comment in the QML code would be great to avoid later wondering about this. Or would you prefer to change the C++ part to avoid having the tocmodel object living in the other thread as it is now ? I choose the onModelChanged solution because it doesn't change big parts of the C++ and do the job without any warning in one line, which seems fine to me (but unusual in the way to write it).

plugin/PDFDocumentToCPage.qml
dcaliste: This is strange, I agree. I put in the PR comment that the enabled attribute of the placeholder is not set by binding since the tocmodel object comes from a different thread and QML complains about signaling from a different thread. Thus, enabled attribute is set by javascript by reading the count value when model is set instead.

Maybe a comment in the QML code would be great to avoid later wondering about this. Or would you prefer to change the C++ part to avoid having the tocmodel object living in the other thread as it is now ? I choose the onModelChanged solution because it doesn't change big parts of the C++ and do the job without any warning in one line, which seems fine to me (but unusual in the way to write it).

VDVsx: Ok, I guess this solution is better then, but please add comment here to avoid someone else from modify it. Thx.

dcaliste: Comment added, I agree it was missing :) Thanks for pointing.

VDVsx:
LGTM

Issue #33, opened 2015-05-15
Search looses focus when typing


dcaliste:
Hello,

   The pull request #29 has broken the search tool. Now, each time one press a letter, the list gets the focus and the keyboard disappears. How can one keep the focus on the search field ?

VDVsx:
focus can be forced, in the search field:
focus: parent.focus
// this is needed for case a list changes inserting or removing elements
focusOutBehavior: FocusBehavior.KeepFocus

dcaliste:
Thanks for your advice. I've tried but it doesn't work. I'm still trying to understand what differ from other cases where this is employed, currently have no clue. I'll ask for help if I can't figure it out in the coming days…

dcaliste:
After trying to figure out what's different between my code and the one in the silica component gallery in the SDK, I gave up and try looking for the web : bingo ! https://lists.sailfishos.org/pipermail/devel/2013-December/002251.html

For reference, setting currentItem to -1 in the listview fix the issue.

dcaliste:
Closed with PR #34.

PR #34, opened 2015-05-19
Close issue #33

dcaliste:
This pull request should correct the focus issue #33. It also use an appropriated message when the list is empty because of searching.


VDVsx:
LGTM, could you amend one of the commits with the internal bug number, currently our integration does not support github issues yet: Contributes JB#28915

dcaliste:
Sure, done, thanks for tracking !

PR #35, opened 2015-05-20
Add scrolldecorator to toc page

lukedirtwalker:


VDVsx:
LGTM, but the commit needs to be amended to contain a prefix like: [office]
Check other commit for a example, this is needed for automatic generation of changelogs.

lukedirtwalker:
Updated commit message

PR #36, opened 2015-06-01
Try to address several usability issues, like page navigation or inline searching for PDF documents.

dcaliste:
I would like to address (see https://together.jolla.com/question/90317/help-improve-document-application/ ) these issues for PDF documents :

 * access to current page number ;
 * jump to a given page not in the table of content ;
 * search in documents ;
 * not easily discoverable table of content page as can be seen in the comment complains in the shop.

I propose as a solution to have a toolbar that is displayed when the document is scrolled up and hidden when document is scrolled down, like in the browser. This toolbar contains a page indicator and a search field. The page indicator opens the index subpage where a goto page entry has been added.

### DIFF
--- a/pdf/pdfdocument.cpp
+++ b/pdf/pdfdocument.cpp
@@ -19,6 +19,7 @@
 #include "pdfdocument.h"
 #include "pdfrenderthread.h"
 #include "pdfjob.h"
+#include "pdfsearchmodel.h"
 
 #include <QDebug>
 #include <QUrl>
@@ -28,9 +29,12 @@
 class PDFDocument::Private
 {
 public:
-    Private() : document( nullptr ), completed(false) { }
+    Private() : searching(false), document( nullptr ), completed(false) { }
 
     PDFRenderThread* thread;
+
+    bool searching;
+    PDFSearchModel *searchModel;
 
     Poppler::Document* document;
     QString source;
@@ -45,11 +49,15 @@ PDFDocument::PDFDocument(QObject* parent)
     connect( d->thread, &PDFRenderThread::loadFinished, this, &PDFDocument::pageCountChanged );
     connect( d->thread, &PDFRenderThread::loadFinished, this, &PDFDocument::loadFinished );
     connect( d->thread, &PDFRenderThread::jobFinished, this, &PDFDocument::jobFinished );
+
+    d->searchModel = nullptr;
 }
 
 PDFDocument::~PDFDocument()
 {
     delete d->thread;
+    if (d->searchModel != nullptr)
+      delete d->searchModel;
 }
 
 QString PDFDocument::source() const
@@ -70,6 +78,16 @@ int PDFDocument::pageCount() const
 QObject* PDFDocument::tocModel() const
 {
     return d->thread->tocModel();
+}
+
+QObject* PDFDocument::searchModel() const
+{
+    return d->searchModel;
+}
+
+bool PDFDocument::searching() const
+{
+    return d->searching;
 }
 
 bool PDFDocument::isLoaded() const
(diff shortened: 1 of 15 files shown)


VDVsx:
Thanks for the PR, quite nice additions, I need to show this to the design team, so will take a bit more.

dcaliste:
Thanks. I understand. I would be happy to answer (or correct when necessary) any questions you may have on the implementation choices. With respect to design, well, let's wait for the review !

dcaliste:
Hello, sorry to raise this, but is there any news from the design team on this pull request ? Besides, I understand that their schedule might be quite busy recently and I can wait anyway…

pdf/pdfdocument.cpp
```diff
@@ -166,6 +184,25 @@ void PDFDocument::requestPageSizes()
     d->thread->queueJob( job );
 }
 
+void PDFDocument::search(const QString &search, uint startPage)
+{
+    if(!isLoaded() || isLocked())
```
VDVsx: Nitpick: Space missing after 'if' in order to follow Qt code guidelines.

pdf/pdfjob.cpp
```diff
@@ -74,3 +74,39 @@ void PageSizesJob::run()
...
+  : PDFJob(PDFJob::SearchDocumentJob), m_search(search), startPage(page)
+{
+  
+}
+
+void SearchDocumentJob::run()
+{
+    Q_ASSERT(m_document);
+
+    for( int i = 0; i < m_document->numPages(); ++i )
```
VDVsx: Nitpick: space missing

pdf/pdfsearchmodel.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+}
+
+QHash< int, QByteArray > PDFSearchModel::roleNames() const
+{
+    QHash< int, QByteArray > names;
+    names[Page] = "page";
+    names[Rect] = "rect";
+    return names;
+}
```
VDVsx: normally this is done in the class constructor, then here we just just return the roles, otherwise this is constructed every time this function is called, see for example: https://github.com/sailfishos/sailfish-office/blob/master/models/documentlistmodel.cpp#L49 or more organized example: https://github.com/sailfishos/sailfish-office/blob/master/models/documentlistmodel.cpp#L49

VDVsx:
HI, sorry for taking so long, I was on holidays and looks like no one looked at this. Few comments on the code, in the UI side didn't find anything major that would prevent this from go in, small improvement could be done to the search bar when something is entered there and the string gets hidden by the X and arrow keys, current there's no fading there, so might look misleading to user if for example user types "test12" and see only "test" when searching, some fading would show that more chars are there.

dcaliste:
Thank you for the review. I'll try to correct as you suggest today, otherwise, I'm leaving for vacations next week, so I guess, I'll amend the commits from August the 10th. Thank you, and excited to see SFOS2, later see this in and hopefully still later on some other parts of the UI released as open source.

pdf/pdfsearchmodel.cpp
pvuorela: IIRC qtquick caches the role names.

dcaliste: Oh, I've changed it anyway. It was quite simple to follow documentlistmodel.cpp.

dcaliste:
To achieve the fading effect in the search field, I didn't find any other way that dealing with the _editor attribute of SearchField component. Beginning with an underscore, I'm afraid, it is a private attribute, or supposed so. Here is the way I did it (I put this in the SearchField declaration in PDFDocumentPage.qml) :
    OpacityRampEffect {
                    sourceItem: search._editor
                    direction: search.horizontalAlignment == Text.AlignRight
                    ? OpacityRamp.RightToLeft
                    : OpacityRamp.LeftToRight
                    slope: search._editor.width / search.textRightMargin * 2.
                    offset: 1. - search.textRightMargin / search._editor.width
                }

Can I commit it like that, or it is an issue and there is a clever way to achieve this ?

VDVsx:
Merging as it is, fading and other small UI changes can be done later once we receive more feedback from design team.

dcaliste:
Wonderfull, thank you.

Issue #37, opened 2015-06-03
Search should not be available when there are no documents


VDVsx:
If placeholder text is displayed pulley menu should not be visible.

dcaliste:
Good point, as stated in https://sailfishos.org/develop/docs/silica/sailfish-application-pitfalls.html/ at the end of the page, I've added a visible attribute to the pull down menu.

lukedirtwalker:
@VDVsx can be closed as it is fixed?

PR #38, opened 2015-06-03
[office] fix issue #37, by hiding the pulldown menu when no documents.

dcaliste:
As explained in https://sailfishos.org/develop/docs/silica/sailfish-application-pitfalls.html/ I've added a visible attribute to the pull down menu.


VDVsx:
LGTM

PR #40, opened 2015-07-10
[sailfish-office] Install a separate .desktop file for mime-type hand…

adenexter:
…ling. Contributes to JB#30574

When starting the application from the launcher don't invoke the
openCoordinate service as that will cause the application to raise
a window that was just shown, which circumvents home's logic for
starting an application minimized.


jpetrell:
LGTM

PR #41, opened 2015-07-21
[office] Correct zooming, increase the maximum zoom and disable the z…

blammit:
…oom proxy. Fixes JB#30396

Increase the maximum zoom so that the document can be zoomed to a
greater amount on devices with larger screens.

Also, scaling wasn't working correctly (more noticeable when zooming
out than in); as zoomAroundPoint() applies the given scale amount by
adding it to the current zoom, the amount should be multiplied by the
scale difference rather than just passed directly.

This also disables the zoomProxy for these views, as the proxy display
is sometimes confused in the current version of Calligra -- sometimes
the proxy is displayed upside down when zooming out (presumably because
the scale applied to the proxy becomes negative).

### DIFF
--- a/plugin/SpreadsheetPage.qml
+++ b/plugin/SpreadsheetPage.qml
@@ -53,6 +53,8 @@ DocumentPage {
             id: controller;
             view: v;
             flickable: f;
+            useZoomProxy: false;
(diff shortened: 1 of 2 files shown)


martinjones:
LGTM

Issue #42, opened 2015-08-14
PDF Table of Contents or Hyperlinks


martin-ueding:
I have a lot of LaTeX documents on my phone. They have the LaTeX PDF hyperlinks from the `hyperref` package. However, I cannot click on the links in the table of contents. For some reason the links I have created manually with `\href{mailto:…}{…}` work.

There is no table of contents (or at least I have not found it). Together this makes it really hard to navigate in the document.

I would really appreciate a table of contents view for the opened PDF.

dcaliste:
The table of content (It's the case for Beamer presentations I've checked for instance) is available as an attached page :
* tap the document to show the sharing panel.
* you'll see there a dot on the top right indicating an attached page that you can reach with a right swype.

For inline links, this is currently not implemented.

martin-ueding:
I did not know that the table of contents was hidden there. Thanks for pointing that out! I'd appreciate working hyperlinks but I cannot give you any input on your implementation approach --- I have no idea about this project or Sailfish OS programming.

PR #43, opened 2015-08-26
[office] Add support for internal Goto link in PDF.

dcaliste:
This adds partial support for internal links in PDF documents. It makes the view scrolls to the target page of the link. It should contribute to issue #42. Currently, the inside of the target page itself is not scrolled.

### DIFF
--- a/pdf/pdfcanvas.cpp
+++ b/pdf/pdfcanvas.cpp
@@ -164,12 +164,12 @@ void PDFCanvas::setDocument(PDFDocument* doc)
     }
 }
 
-qreal PDFCanvas::pagePosition(int index) const
+QRectF PDFCanvas::pageRectangle(int index) const
 {
     if( d->pages.count() == 0 )
-        return 0.f;
+        return QRectF();
 
-    return d->pages.value( index ).rect.y();
+    return d->pages.value( index ).rect;
 }
 
 int PDFCanvas::currentPage() const
--- a/pdf/pdfrenderthread.cpp
+++ b/pdf/pdfrenderthread.cpp
@@ -92,12 +92,27 @@ public:
             Poppler::Page* page = document->page(i);
             for(Poppler::Link* link : page->links())
             {
-                if(link->linkType() == Poppler::Link::Browse)
-                {
+                switch (link->linkType()) {
+                case (Poppler::Link::Browse): {
                     Poppler::LinkBrowse* realLink = static_cast<Poppler::LinkBrowse*>(link);
                     QRectF linkArea = link->linkArea();
                     linkTargets.insert( i, QPair< QRectF, QUrl >{ linkArea, realLink->url() } );
+                    break;
                 }
+                case (Poppler::Link::Goto): {
+                    Poppler::LinkGoto* gotoLink = static_cast<Poppler::LinkGoto*>(link);
+                    // Not handling goto link to external file currently.
+                    if (gotoLink->isExternal())
+                        break;
+                    QRectF linkArea = link->linkArea();
+                    QString linkPage = QString("?page=%1").arg(gotoLink->destination().pageNumber());
+                    linkTargets.insert( i, QPair< QRectF, QUrl >{ linkArea, linkPage } );
+                    break;
+                }
+                default:
+                    break;
+                }
+
             }
         }
     }
--- a/pdf/pdfrenderthread.cpp
+++ b/pdf/pdfrenderthread.cpp
@@ -24,6 +24,7 @@
 #include <QMutex>
 #include <QDebug>
 #include <QCoreApplication>
+#include <QUrlQuery>
 
 #include <poppler-qt5.h>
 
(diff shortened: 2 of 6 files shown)


dcaliste:
The second commit in this PR add support for in-page shifts in goto links.

In PDF Goto links, the destination is defined by :
* possibly an external file (not implemented) ;
* a mandatory page id (first commit) ;
* a possible shift inside the page (second commit) ;
* a possible zoom level (not implemented).

pdf/pdfcanvas.cpp
```diff
@@ -164,12 +164,14 @@ void PDFCanvas::setDocument(PDFDocument* doc)
...
-qreal PDFCanvas::pagePosition(int index) const
+qreal PDFCanvas::pagePosition(int index, qreal offset) const
 {
     if( d->pages.count() == 0 )
         return 0.f;
 
-    return d->pages.value( index ).rect.y();
+    const QRectF& rect = d->pages.value( index ).rect;
+    qreal minOffset = 0., maxOffset = 1.;
+    return rect.y() + rect.height() * qBound(minOffset, offset, maxOffset);
```
pvuorela: Minor detail, and fine as is too, but to me offset sounds like a discrete value, pixels for example. Also as qml side calculates contentX, maybe it could be simpler and more consistent, if this method would be called pageRectangle() and qml would decide position inside the page?

plugin/PDFView.qml
```diff
@@ -178,7 +179,13 @@ SilicaFlickable {
         VerticalScrollDecorator { color: Theme.highlightDimmerColor; }
     ]
 
-    function goToPage(pageNumber) {
-        base.contentY = pdfCanvas.pagePosition( pageNumber );
+    function goToPage(pageNumber, top, left) {
+        // Adjust horizontal position if required.
+        if (left !== undefined && left >= 0.) {
+            base.contentX = base.itemWidth * left
```
pvuorela: Tested with first pdf with links I could find from the internet: http://www.rpi.edu/dept/arc/docs/latex-thesis/latextopdf.pdf

On that, clicking "section 6" link on first page, the new position has text touching edges of the screen, which doesn't look nice. How about having some offset left and above new position? Could perhaps also guard document bottom to stay on the screen?

pvuorela:
Nice! Couple comments, but on general good to go from my side.

plugin/PDFView.qml
dcaliste: Yeh, I see. I will modify the commit to add some offsets, you're right.

pdf/pdfcanvas.cpp
dcaliste: I agree to change offset that's not sounding for what it is. Maybe something like "qreal inPage".

For the rest, I may have not understood your point. For me, this value is mandatory in the function because every single page height is not known from QML side and because of spacing between pages, it is a bit cumbersome to use pagePosition(i+1) - pagePosition(i). ContentX is different because it refers to the width which is  known from the QML side.

pvuorela: The later sugestion was replacing qreal pagePosition(int) with QRectF pageRectangle(int). That gets converted into qml rect type and code in PDFView.qml can decide what position between rect.y and (rect.y + rect.height) to set.

dcaliste: Ah, finally I got your point, sorry being slow ! You're absolutely right. I like your suggestion, I'll amend the commit to change it for this.

dcaliste:
I've amended the commits to:
* change the pagePosition() function into pageRectangle() function and the QML handle the shift based on this ;
* add optional offsets when moving to a specific position in a page. They are optional because this same function goToPage() may be used by a later PR to set back the view to a saved position and in this case, we don't want offsets.

pdf/pdfrenderthread.cpp
```diff
@@ -92,12 +93,36 @@ class PDFRenderThreadPrivate
...
                     linkTargets.insert( i, QPair< QRectF, QUrl >{ linkArea, realLink->url() } );
+                    break;
                 }
+                case (Poppler::Link::Goto): {
+                    Poppler::LinkGoto* gotoLink = static_cast<Poppler::LinkGoto*>(link);
+                    // Not handling goto link to external file currently.
+                    if (gotoLink->isExternal())
+                        break;
+                    QRectF linkArea = link->linkArea();
+                    QUrl linkURL = QUrl("?");
```
pvuorela: Minor detail, but I think "?" is unnecessary here.

pvuorela:
The case of scrolling beyond document end is still there. On the pdf I linked can be triggered by having zoom level at minimum, full document width shown. Then clicking "6" on first page (can be quite hard to hit as the link text is very small) -> document jumps to section, application background is shown on the right and bottom half of the screen. Press and small movement makes the document jump back inside screen.

If you want, can already merge this as is and have above fixed separately.

dcaliste:
Oups, sorry, no don't merge. I forgot this remark. I'll correct it before merge.

pdf/pdfrenderthread.cpp
dcaliste: I agree, I remove it.

dcaliste:
Ok, I've added tests to check that a link don't send the view beyond the limits. By the way, this issue was pre-existing, for instance with a PDF document with pages smaller that the screen, going to the last page (with the goto input field), put this page on top of the screen and thus leave much empty space below. So one minor issue less !

dcaliste:
Concerning the usability issue, I've tested with a wiggleFactor = 15.f in pdfcanvas.cpp. It helps a bit but it's not very convincing. One could increase this value more, but I don't think it's a good idea, especially when links are close because the urlAtPoint() algorithm (see pdfcanvas.cpp) currently stops on the first expanded link rectangle that contains the tap. So if we expand link rectangles too much, the first overlapping rectangle will hide all other ones.

Instead, I propose to change a bit the urlAtPoint() algorithm (see pdfcanvas.cpp) to use a distance between non-expanded link rectangles and tap point. We select the link as the min of these distances and also if it is below a (big) threshold given by Theme.itemSizeExtraSmall for instance (to be resolution independant). What do you think ?

pvuorela:
Yea, makes sense to use closest link instead of topmost overlapping.

Also, this looks good to me now. Merging.

PR #44, opened 2015-09-18
[office] Change the current page value according to biggest page shown instead of top one.

dcaliste:
Change slightly the algorithm to get current page :
* previously, page number correspond to the page the top line of the screen belongs to.
* new implementation, page number is the page with the biggest shown surface, if several have the biggest shown surface, the first to match gives the page number.

This PR is necessary for a later one that is in my last-page branch to save current page when quiting a document. It's better to have a more natural number for the pages.

This new way is much more natural and simply implemented with a strictly superior to previously found biggest area. It has only one drawback that is IMHO not very important since it's a corner case : it will give a false page number in the case where several full pages are visible on the  screen but one of them, near the bottom, has a different size and is bigger. To treat easily this corner case would require a lot of ifs in the code and makes it less easy to read.

### DIFF
--- a/pdf/pdfcanvas.cpp
+++ b/pdf/pdfcanvas.cpp
@@ -51,7 +51,7 @@ public:
     Private( PDFCanvas* qq )
         : q{ qq }
         , pageCount{ 0 }
-        , currentPage{ 0 }
+        , currentPage{ 1 }
         , renderWidth{ 0 }
         , document{ nullptr }
         , flickable(0)
@@ -354,7 +354,8 @@ QSGNode* PDFCanvas::updatePaintNode(QSGNode* node, QQuickItem::UpdatePaintNodeDa
     }
 
     QList<QPair<int, int> > priorityRequests;
-    bool currentPageSet = false;
+    int currentPage = d->currentPage;
+    qreal maxVisibleArea = 0.;
 
     for( int i = 0; i < d->pageCount; ++i )
     {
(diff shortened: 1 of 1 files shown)


pvuorela:
Makes sense for having current page the biggest area shown.

One slight additional behavioral change would be PDFDocumentPage / SearchField which wouldn't anymore start search from first visible page if next one has bigger area. Not necessarily wrong though.

Would it make sense to compare visible height instead of area if pages can have different width?

pdf/pdfcanvas.cpp
```diff
@@ -468,6 +472,11 @@ QSGNode* PDFCanvas::updatePaintNode(QSGNode* node, QQuickItem::UpdatePaintNodeDa
 
     d->cleanTextures();
 
+    if (d->currentPage != currentPage) {
+        d->currentPage = currentPage;
```
pvuorela: Opening a PDF I see page first changing into -1 and then the correct value.

dcaliste: Good catch, I'll correct.

dcaliste:
Sorry, maybe I'm wrong. This patch is not modifying the access to the page (search or table of content), but the number that is displayed when moving inside the document. So I don't think this will affect the SearchField, or at least I don't see how.

pvuorela:
https://github.com/sailfishos/sailfish-office/blob/master/plugin/PDFDocumentPage.qml#L134
That appears to ask search to start from current page.

dcaliste:
Ah, OK, catch your point. That's right. IMHO I would leave it like that for the moment for code simplicity. I did the same with the first crude implementation of current page, and submit an improved algorithm when the necessity arised. What do you think ?

pvuorela:
I think it's fine. Evince seemed to have similar search and current page functionality too.

dcaliste:
I believe the -1 issue you reported should be corrected now in the latest version of the PR.

pvuorela:
LGTM.

> This PR is necessary for a later one that is in my last-page branch to save current page when quiting a document.

What kind of plans do you have btw? Suppose restoring the page later? Had a chat with Vappu who had done the initial design long time ago. Her initial thought for such was to restore page when document is opened from main page later. Also agreed that the whole UI should be revisited. Clicking links doesn't for example work too nice with click to switch split screen mode.

dcaliste:
Thanks for merging.

In my last-page branch on gtihub, I have an initial implementation to save to a local sqlite database for each document the page index when quiting the view of the document. So, when the document is open later, it scroll to last seen page. The zoom level and inside page position are not saved currently because I was waiting for the change to the goToPage() function to add left and top arguments. Note that this devel is not ready yet but you can give a look to the commits if you want to have an idea there.

Conserning the issue of tapping right on the links, I agree it's currently quite bad. I've proposed the implementation of the goto links because it was asked by someone here and the existing code to handle external links was quite simple to extend. There is a wingle variable in pdfcanvas.cpp source file which expand the border of each links for tap detection. It is currently set to 4 pixels. Maybe one can put it to a bit more and try to play a bit with it ?

PR #45, opened 2015-09-20
[office] Add a shape visual for each page that has not been yet rendered.

dcaliste:
This is a small modification in the paint node method to display a diffused rectangle shape of a page when the page has not been rendered yet. This gives a minimal visual feedback of the position in the document for instance when scrolling fast, or when  going to the next search match several pages further.

### DIFF
--- a/pdf/pdfcanvas.cpp
+++ b/pdf/pdfcanvas.cpp
@@ -34,12 +34,14 @@ struct PDFPage {
     PDFPage()
         : index{-1}
         , requested{ false }
+        , rectPlaceholder{ false }
         , texture{ nullptr }
     { }
 
     int index;
     QRectF rect;
     bool requested;
+    bool rectPlaceholder;
     QSGTexture *texture;
 
     QList< QPair< QRectF, QUrl > > links;
@@ -80,6 +82,7 @@ public:
     QList< QSizeF > pageSizes;
 
     QColor linkColor;
+    QColor pagePlaceholderColor;
 
     QList<QSGTexture *> texturesToClean;
     QPointer<QQuickWindow> connectedWindow;
@@ -205,6 +208,21 @@ void PDFCanvas::setLinkColor(const QColor& color)
         emit linkColorChanged();
     }
 }
+QColor PDFCanvas::pagePlaceholderColor() const
+{
+    return d->pagePlaceholderColor;
+}
+
+void PDFCanvas::setPagePlaceholderColor(const QColor& color)
+{
+    if( color != d->pagePlaceholderColor )
+    {
+        d->pagePlaceholderColor = color;
+        d->pagePlaceholderColor.setAlphaF( 0.25 );
+        update();
+        emit pagePlaceholderColorChanged();
+    }
+}
 
 void PDFCanvas::layout()
 {
@@ -318,6 +336,7 @@ void PDFCanvas::sceneGraphInvalidated()
             page.texture = 0;
         }
         page.requested = false;
+        page.rectPlaceholder = false;
     }
 }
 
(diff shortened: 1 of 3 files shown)


pdf/pdfcanvas.cpp
```diff
@@ -398,6 +398,10 @@ QSGNode* PDFCanvas::updatePaintNode(QSGNode* node, QQuickItem::UpdatePaintNodeDa
             t = new QSGTransformNode{};
             root->appendChildNode(t);
         }
+        else if( t->childCount() > 0 ) {
+            delete t->firstChild();
+            t->removeAllChildNodes();
```
pvuorela: This would delete and below recreate child nodes when view is scrolled even a bit. Should avoid if it's already the wanted type of node?

pdf/pdfcanvas.cpp
```diff
@@ -449,13 +453,12 @@ QSGNode* PDFCanvas::updatePaintNode(QSGNode* node, QQuickItem::UpdatePaintNodeDa
...
         {
-            if( t->childCount() > 0 )
-            {
-                delete t->firstChild();
-                t->removeAllChildNodes();
-            }
+            QSGSimpleRectNode *bgNode = new QSGSimpleRectNode();
+            t->appendChildNode( bgNode );
+            bgNode->setRect( QRectF(0., 0., page.rect.width(), page.rect.height()) );
+            bgNode->setColor( d->linkColor );
```
pvuorela: Rectangle is not really a link, could define a separate property perhaps?

pdf/pdfcanvas.cpp
dcaliste: Yes, I see. I'll rework this.

pdf/pdfcanvas.cpp
dcaliste: I agree, I was a bit lazy last Sunday and quite satisfied with the visual result because the link color property is set to Theme.highlightColor plus an alpha… I'll add one specific property which is cleaner than supposing that the link color will always be highlightColor !

dcaliste:
I've added a pagePlaceholderColor property. Maybe the name is too long ?

I've changed a bit the paint function to avoid deleting the QGSNode each time we paint. I didn't find how to differentiate QGSSimpleTextureNode and QSGSimpleRectangleNode because they are not inheriting from QObject so qobject_cast() is not available. Instead I've added a flag (rectPlaceholder) that is true when the first child of the QGSNode is indeed a rectangle, to know if we need to delete the node to insert the texture. What do you think ?

pdf/pdfcanvas.cpp
```diff
@@ -34,12 +34,13 @@ struct PDFPage {
...
         : index{-1}
         , requested{ false }
+        , rectPlaceholder{ false }
         , texture{ nullptr }
     { }
 
     int index;
     QRectF rect;
-    bool requested;
+    bool requested, rectPlaceholder;
```
pvuorela: As personal preference, I'd keep member variable declarations separate to ease reading and follow existing convention. Doesn't matter much, though.

pdf/pdfcanvas.cpp
```diff
@@ -226,6 +243,7 @@ void PDFCanvas::layout()
         page.index = i;
         page.rect = QRectF(0, totalHeight, width(), width() * ratio);
         page.links = links.values( i );
+        page.rectPlaceholder = false;
```
pvuorela: Minor detail, but looks unnecessary, like line below too, as ctor sets false already.

pvuorela:
I think the name is fine. Not referenced in many places so clarity wins over few characters.

Couple cosmetic comments, which don't matter much. LGTM.

pdf/pdfcanvas.cpp
dcaliste: You're right. Commit changed.

pdf/pdfcanvas.cpp
dcaliste: I agree, that's right. Commit changed.

dcaliste:
I've corrected the minor details by the way.

pvuorela:
Looking good. Thanks!

PR #46, opened 2015-09-21
Adjust style, redundant code, deprecated property etc

pvuorela:


pvuorela:
Added a bunch of memory leak fixes. Still seeing page texture leaking, and also memory size keeps increasing when reopening document, but will continue investigating.

pvuorela:
@tanty or @dcaliste perhaps?

dcaliste:
Thanks for correcting all these leaks. I was not diligent enough !

I've clone your branch and tried different things this morning without much success up to now :
* run valgrind didn't appear usefull, got a lot of stack errors but without places where it happens, I need to find how to compile with -g maybe. The memory report didn't highlight any big leaks (some kB only) on program end, and most of them due to GLib static stuff that are not freed on quit.
* fallback to use 'top' and indeed notice a leak of ~8MB for a two pages PDF file. So I remove the texture loading (put a && false in pdfcanvas line 383 to disable the pageRequest) and still got some MB of leaks, so I tried to look further. I removed the PDFView instance and the PDF.Document instance from PDFDocumentPage, but still notice a ~2MB of increase each time I tap on a PDF from the document list. Didn't understand why, because mostly it add a QML Page to the PageStack and then release it.

I didn't have time to dig further. I will try in the coming days.

pvuorela:
Some of the memory usage comes from qml being garbage collected language. Garbage colection will not necessarily trigger each time after document is closed. Opening PDF the first time has also some memory usage for compiling qml code.

Have been myself running valgrind and don't anymore spot too much things going wrong. What does still happen is memory usage increasing when application with pdf open is swiped away and then brought back foreground. Investigating why, but it should be a separate issue.

dcaliste:
Well the 8MB I notice are pilling up when leaving / loading the PDF file is done repeatedly. I'm afraid it's not only due to garbage collected data, but I maybe wrong, being mostly accustomed to C (and Fortran). I will try your usecase also and see if I can spot something.

pvuorela:
Swipe + foreground seems to be cover updating snapshot with updatePreview(). Previous becomes garbage and gets freed then some day. Pondering whether to have explicit gc() call or have some c++ instead.

models/documentproviderplugin.h
```diff
@@ -33,8 +33,8 @@ class DocumentProviderPlugin : public QObject
     Q_PROPERTY(QObject *model READ model NOTIFY modelChanged)
     Q_PROPERTY(bool ready READ isReady NOTIFY readyChanged)
     Q_PROPERTY(QUrl page READ page WRITE setPage NOTIFY pageChanged)
-    Q_PROPERTY(QString setupPageUrl READ setupPageUrl NOTIFY titleChanged);
-    Q_PROPERTY(bool needsSetup READ needsSetup NOTIFY needsSetupChanged);
+    Q_PROPERTY(QString setupPageUrl READ setupPageUrl NOTIFY titleChanged)
+    Q_PROPERTY(bool needsSetup READ needsSetup NOTIFY needsSetupChanged)
```
rainemak: I'd assume that all of these could be declared as ```FINAL``` e.i. nobody overrides these properties.

pvuorela: Might be, but should be done separately instead of accumulating more changes for this PR.

pdf/pdfcanvas.cpp
```diff
@@ -113,9 +113,16 @@ PDFCanvas::PDFCanvas(QQuickItem* parent)
 
 PDFCanvas::~PDFCanvas()
 {
-    if(d->resizeTimer)
-        delete d->resizeTimer;
+    for( int i = 0; i < d->pageCount; ++i )
```
rainemak: Should be like...
``` 
for (int ... ++i) {
}
``` 
The rests seems to follow more or less the style you used. So I'm fine with that as well. One could do a commit that fixes all coding style related issues in a commit. Let's do not mix it up over here.

pvuorela: The code in this repository doesn't have consistent style on many aspects. Loop and brace style I didn't yet touch and this loop is now consistent with similar existing one.

rainemak: Is that ```d->pageCount``` really needed. Would the pages.count() be enough (with a temporary variable). Just increases bookkeeping.

pvuorela: First, this alone shouldn't do anything different than sceneGraphInvalidated(). Second, pages.count() would assume all the pages are put to QHash, which I don't think this should do.

plugin/ZoomableThumbnail.qml
```diff
@@ -24,19 +24,17 @@ SilicaFlickable {
     id: base;
 
     property real maxHeight: height;
+    property alias content: thumb.data;
```
rainemak: Extra semi colons

plugin/ZoomableThumbnail.qml
```diff
@@ -24,19 +24,17 @@ SilicaFlickable {
...
+    property bool scaled: false;
```
rainemak: No need to have default value false

plugin/ZoomableThumbnail.qml
rainemak: Nevermind about semi-colons. File seems to be full of them. Worth adding a cleanup task to github repo.

rainemak:
Added a few comments, please check them. LGTM.

pvuorela:
I'll continue on style details later. Merging this.

PR #47, opened 2015-10-05
Remember last page position for PDF documents

dcaliste:
As described in the [TJC post](https://together.jolla.com/question/90317/help-improve-document-application/#lastpage), I propose here an implementation for the feature to remember last page position for PDF documents. It is based on :
* the use of a local SQlite database (see plugins/PDFStorage.js) with a table indexed by file path, containing the page index, the position in the page (in reduced coordinates) and the actual width of the page in pixel. I prefered to store these (more) abstract data, than the actual contentX, contentY and width, height of the PDFCanvas object, to be sure that the last position is properly reproduced when opening a file in landscape while position was stored in portrait (or vice-versa).
* I open a database connection every time a PDF file is taped and loaded to fetch for the possibly stored position data. I don't know if it's clever or not. I prefered this than a global variable owned by the document list and passed to the PDFView instance. Like that it is self contained in PDFView.
* the loading of the stored settings is done on the first emission of the pageLayoutChanged signal and not on Component.completed since, it's only after this signal that page geometries are known.
* the storing to the database is done on Component destruction. But there's a trick here I don't like. If I put the Component.onDestruction handler in the PDFCanvas instance, when triggered, the properties have been already destroyed (_settings attribute is null there). While putting the signal handler in the parent (the PDFView instance), the child is not destroyed yet, and I can do things with _settings. It's fragile IMHO since the Qt documentation explicitly says that the order of destroying is not guaranteed. But I have no other idea. I've tried in fact to react on the page transition, which works fine, but saving is not done when we quit the application directly, which is a no go of course.
* Finally the PR is divided into two commits, one to bring all the machinery for storing the page data, the other one to add a setting for the document application to enable or not this feature (that may be disturbing first to go back on a big zoom level in a middle page while not knowing about this, so it's an opt-in feature).

Sorry, long text, but for this feature that seems so simple (even to implement), it finally requires many explanations since many choices were done.

### DIFF
--- a/plugin/PDFView.qml
+++ b/plugin/PDFView.qml
@@ -19,6 +19,7 @@
 import QtQuick 2.0
 import Sailfish.Silica 1.0
 import Sailfish.Office.PDF 1.0 as PDF
+import org.nemomobile.configuration 1.0
 import QtQuick.LocalStorage 2.0
 import "PDFStorage.js" as PDFStorage
 
@@ -124,7 +125,7 @@ SilicaFlickable {
     // Ensure proper zooming level when device is rotated.
     onWidthChanged: adjust()
 
-    Component.onDestruction: pdfCanvas.savePageSettings()
+    Component.onDestruction: if (rememberPositionConfig.value) pdfCanvas.savePageSettings()
     PDF.Canvas {
         id: pdfCanvas;
 
@@ -135,7 +136,7 @@ SilicaFlickable {
         linkColor: Theme.highlightColor;
         pagePlaceholderColor: Theme.highlightColor;
 
-        onPageLayoutChanged: restorePageSettings()
+        onPageLayoutChanged: if (rememberPositionConfig.value) restorePageSettings()
 
         // Handle save and restore the view settings.
         property var _settings
@@ -231,5 +232,12 @@ SilicaFlickable {
         }
         contentX = Math.max(0, scrollX)
         contentY = Math.max(0, scrollY)
+    }
+
+    ConfigurationValue {
+        id: rememberPositionConfig
+
+        key: "/apps/sailfish-office/settings/rememberPosition"
+        defaultValue: false
     }
 }
--- a/plugin/PDFView.qml
+++ b/plugin/PDFView.qml
@@ -19,6 +19,8 @@
 import QtQuick 2.0
 import Sailfish.Silica 1.0
 import Sailfish.Office.PDF 1.0 as PDF
+import QtQuick.LocalStorage 2.0
+import "PDFStorage.js" as PDFStorage
 
 SilicaFlickable {
     id: base;
@@ -122,6 +124,7 @@ SilicaFlickable {
     // Ensure proper zooming level when device is rotated.
     onWidthChanged: adjust()
 
+    Component.onDestruction: pdfCanvas.savePageSettings()
     PDF.Canvas {
         id: pdfCanvas;
 
@@ -131,6 +134,35 @@ SilicaFlickable {
         flickable: base;
         linkColor: Theme.highlightColor;
         pagePlaceholderColor: Theme.highlightColor;
+
+        onPageLayoutChanged: restorePageSettings()
+
+        // Handle save and restore the view settings.
+        property var _settings
+        function restorePageSettings() {
+            if (!_settings) {
+                _settings = new PDFStorage.Settings(document.source)
+                var last = _settings.getLastPage()
+                if (last[3] > 0) {
+                    width = last[3]
+                    base.adjust()
+                }
+                base.goToPage( last[0], last[1], last[2] )
+            }
+        }
+        function savePageSettings() {
+            if (_settings) {
+                // Find the page on top
+                var i = currentPage - 1
+                var rect = pageRectangle( i )
+                while (rect.y > contentY && i >= 0) {
+                    rect = pageRectangle( --i )
+                }
+                var top  = (contentY - rect.y) / rect.height
+                var left = (contentX - rect.x) / rect.width
+                _settings.setLastPage(i, top, left, width)
+            }
+        }
 
         PinchArea {
             anchors.fill: parent;
@@ -197,7 +229,7 @@ SilicaFlickable {
         if (scrollY > contentHeight - height) {
             scrollY = contentHeight - height
         }
-        contentX = scrollX
-        contentY = scrollY
+        contentX = Math.max(0, scrollX)
+        contentY = Math.max(0, scrollY)
     }
 }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,7 @@ configure_file(config.h.in config.h)
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x -fPIC -pie -rdynamic -Wall")
 
+add_subdirectory(settings)
 add_subdirectory(plugin)
 add_subdirectory(pdf)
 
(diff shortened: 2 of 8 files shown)


pvuorela:
Talked with Vappu/design that there doesn't seem enough need for having a separate setting. Not something too many people should be expected to change. If restoring page with zooming is disturbing, can as well avoid zooming. Should be enough already if current page number is restored.

plugin/PDFView.qml
```diff
@@ -132,6 +136,35 @@ SilicaFlickable {
         linkColor: Theme.highlightColor;
         pagePlaceholderColor: Theme.highlightColor;
 
+        onPageLayoutChanged: if (rememberPositionConfig.value) restorePageSettings()
```
pvuorela: Not sure if pageLayoutChanged is proper for this. Emitted also on resizing and something like that. Should maybe have some documentLoaded signal?

pvuorela: Or would PDFDocumentPage be more natural place save/restore functionality? Rationale: PDFView would stay more as a "stupid" viewer component for content, not needing to care if multiple instances are created of same file or anything like that.

dcaliste: For the pageLayoutChanged signal, the issue is that, on documentLoad signal, the page geometry is not yet computed. Indeed, the signal for the page geometry is pageLayout. But as you mentioned, this is emitted on each resizing (read here, portrait / landscape rotation). So we need a signal that is firstPageLayout. Since, I found emitting such a signal quite ugly because too much oriented for the page-remember feature, I decide to plug the pageRestore function on every pageLayoutChanged *and* in the function do the job only once, thanks to the "if (_settings !== undefined)" statement.

dcaliste: For the place of the code, *i.e.* PDFView or PDFDocumentPage, I hesitated much before deciding for PDFView because the actual values we want to save / restore are indeed there. If I put the save / restore functions in the PDFDocumentPage instance, we'll need accessors for the pageRectangle() function, the pageLayout signal and for the contentX and contentY properties. That finally decided me to put the code in PDFView, even if I agree that a dumber PDFView would have been desirable from a reusability point of view.

Besides, if we consider that's better to have a simpler PDFView object, I can figure out how to move the save / restore stuff to PDFDocumentPage.

dcaliste:
For the setting, I can remove the second commit, so like that, there is only the page-remember code, without the setting ? I can still keep the setting code in a dead branch in case. Do you agree like that ?

pvuorela:
No much harm keeping a dconf option without settings UI. Just needs default value changed.

plugin/PDFView.qml
pvuorela: Now I'm a bit puzzled for what actual case of pageLayoutChanged should restoring be applied to as it's emitted multiple times without order defined too well.

PageRectangle() is not needed if restoring uses only page number, not caring about zoom level. Both versions are fine, I guess.

pvuorela:
Component.onDestruction bothers me also. If properties get destroyed before that, it gets tricky. If nothing else, having storage in C++, available as context property not owned by engine should guarantee it outlives the pdf view components.

dcaliste:
Ok, for the configuration without UI, that's fine like that.

plugin/PDFView.qml
dcaliste: Sorry, maybe I was not clear enough in my explanations.

To restore the page settings, and thus use the goToPage() function, the PDFCanvas object must know each page.rect. Each page.rect are calculated on the C++ side in a separate thread. This calculation is done each time the width of the PDFCanvas object is changed, thus it is done when the PDFCanvas is first set as a child in the QML page hierarchy. If one calls goToPage() before the page.rect has been calculated (for instance when reacting at the signal documentLoaded before the thread finished is work), nothing happens.

My reasoning was the following :
* I need to call goToPage(), thus I need to know when the thread finished to calculate all the page.rect.
* looking in the C++ side, I see that pageLayoutChanged signal is indeed already emitted when the thread finishes its calculation.
* thus I connect to this thread but react only once.

The path is the following :
documentLoaded -> […] -> run thread for page layout -> emit pageLayoutChanged -> adjust page position if any storage exists for this document -> don't react anymore on later pageLayoutChanged signal.

dcaliste: About PageRectangle(), well there are currently three information that are stored :
* page index, with a little adjustement because currentPage is the biggest page shown, but in this patch I wanted to restore the exact screen configuration, thus if we are showing page N but on top a bit of page N-1 is visible, I need to store N-1 in the database ;
* page position, *i.e.* the (left, top) coordinates, so in my previous example, I show the page N-1 but translate the view to a coordinate top= 0.85 for instance.
* zoom level, stored as the width of the PDFCanvas object in pixel.

I can drop the last information, but I think that having (index, top, left) is quite fine and easier for visual memory to associate state when leaving the document with the state when returning to the document. Additionally, having (index, top, left) fits well with the current state of the goToPage() function which waits for these info indeed.

And finally to return to your question, PageRectangle() is needed to obtain this tripplet.

dcaliste:
I aggre with you, that it may be better to have the storage stuff done at the C++ level but :
* it was quite simple to use the storage API from JS and I don't know how to do this in C++, but I can learn !
* the stored information requires to obtain data from two QML objects : the PDFView for the contentX and contentY properties and the PDFCanvas for currentPage and various page.rect data.

So I don't know much about what to do next about this onDestruction issue.

dcaliste:
By the way, thank you for taking time reviewing this and all the rest and having discussions that improves the result !

plugin/PDFView.qml
pvuorela: What confused me was layout() called in multiple places, first happening in geometry changed handler, but looking closer it seems that call to layout() just returns on no page sizes existing. Requests sizes, but I'd assume that gets ignored due to document not loaded.

pvuorela:
Indeed the javascript interface is nicely simple. There exists QSql for c++, but it's a bit more complicated.

Another alternative is also restricting lifetime even more. If settings is not kept as property, it cannot be destroyed before parent. Instead both loading and onDestruction could create a new shortly lived instance of LocalStorage. How feasible that is depends on how much time the construction takes on worst case. Few milliseconds is ok.

dcaliste:
I'm afraid, it's not possible to use the short life solution, because, one needs the file path to be used as index in the database, but this file path is taken from document.source, so from a property also, which is null at the time onDestruction() is called (just tested).

Concerning the implementation on the C++ side, one needs contentX and contentY which are properties of PDFView, not PDFCanvas, but the PDFView object is reachable from the C++ side thanks to the flickable Q_Property of pdfcanvas object. But as an afterthought, I guess that if destruction order is not guaranteed on the QML side, the flickable property is not guaranteed to be still not null at the time ~pdfcanvas() is called.

Concidering all of this, as a final solution that should work and makes the stuff robust, it would be to save the rect visibleArea that is calculated at pdfcanvas.cpp:360 and use this in ~pdfcanvas() to save relevant info to database from C++ side. I'll try this in the coming days.

pvuorela:
One alternative more: instead of saving at any destruction point do that when app becomes inactive or a pdf document page is popped from page stack.

dcaliste:
Except if there is a signal I don't know about, I tried already to store position when the page becomes inactive (so before being popped…). It worked well, except that it's not saving when quiting by closing with a top swipe, because the page doesn't become inactive, it's just destroyed. So it's a bit strange when being accustomed to page saving but it's not working when quiting…

pvuorela:
There's Qt.application with state and aboutToQuit signal.

dcaliste:
I'm closing this one, in favour of PR #48. Feel free to contradict and prefer the QML + JS implementation, I'll correct then what we have discussed already.

PR #48, opened 2015-10-08
[office] Implement persitent storage of page location for PDF documents (C++).

dcaliste:
This is a revamp PR of the previous one (see #47). Now the implementation is done at the C++ level:
* pdfcanvas.cpp has been modified with a boolean property to activate or not the page remember mechanism (this property is set from QML with a ConfigurationValue, but no UI).
* when pdfcanvas.cpp receives the documentLoaded signal, it opens a database connection to read if there are any position information stored for this document. Retrieved values are stored for later use, when the pageSizesFinished signal will be received. The data base connection is done earlier than when the pageSizesFinished is received to overlap with the calculation of page sizes done in a separate thread.
* when pdfcanvas.cpp receives the pageSizesFinished signal, if any remember values were retrieve it emits the new requestPageWidth / requestPagePosition signals that are catched by QML to call the adjust() and goToPage() function.
* in ~PDFCanvas(), page position are stored in the database. It requires here to have stored the filename source() of document, because document object is already destroyed when ~PDFCanvas() is called.

Hopefully, this implementation will be cleaner than the previous one, at least it's not suffering from the multi-destruction issue since all information that need to be stored are located in the PDFCanvas object.

Any remarks or asks for improvements are welcomed.


dcaliste:
Just have rebased on master, and correct space indentation in pdf/pdfcanvas.cpp.

pdf/pdfcanvas.cpp
```diff
@@ -78,6 +89,12 @@ class PDFCanvas::Private
     float spacing;
 
     QRectF visibleArea;
+    bool rememberPagePosition;
+    QString dSource;
```
pvuorela: Better not abbreviate names unless really necessary. Then again is the whole member variable needed when it can be fetched with document pointer?

pdf/pdfcanvas.cpp
```diff
@@ -535,5 +582,123 @@ void PDFCanvas::resizeTimeout()
...
+    if (d->rememberPage > 0 || d->rememberTop > 0. || d->rememberLeft > 0.) {
+      /* We wait for the layout to be done here, so pageRectangle() will
+         be available at QML level. */
+      emit requestPagePosition(d->rememberPage, d->rememberTop, d->rememberLeft);
+    }
+}
+
+/* Part related to access the local storage, as available from QML side. */
+static const QString dbConnection{"PdfSettings"};
+static QString _databasesPath()
```
pvuorela: Underscore prefix with C++ doesn't seem like a good idea.

pdf/pdfcanvas.cpp
dcaliste: Ok, I'll put full name. For the copy of document->source() in dSource, it's something I don't know in C++. Explanations :
* PDFCanvas::Private has a pointer to a document. It is set by PDFCanvas::setDocument() by just a pointer assignment (d->document = document).
* when ~PDFCanvas() is called, the actual document instance it points to has already been destroyed, so calling document->source() segfaults there.

In C with glib I would have count ref document (g_object_ref(document)) in the setDocument() method, to be sure that document is available during the full life of my PDFCanvas object. I would have called g_object_unref() then in the destroy method of PDFCanvas. I don't know how to do this in C++, *i.e.* I don't know how to add a reference to an object to be sure that we can call its methods whenever we want to.

pdf/pdfcanvas.cpp
dcaliste: Ok, I didn't know about this. What do you suggest as a naming scheme for local static funtions ?

pdf/pdfcanvas.cpp
```diff
@@ -535,5 +582,123 @@ void PDFCanvas::resizeTimeout()
 void PDFCanvas::pageSizesFinished(const QList< QSizeF >& sizes)
 {
     d->pageSizes = sizes;
+    if (d->rememberWidth > 0) {
+      emit requestPageWidth(d->rememberWidth);
```
pvuorela: Inconsistent indentation here and below.

pdf/pdfcanvas.cpp
```diff
@@ -535,5 +582,123 @@ void PDFCanvas::resizeTimeout()
...
+    }
+}
+
+/* Part related to access the local storage, as available from QML side. */
+static const QString dbConnection{"PdfSettings"};
+static QString _databasesPath()
+{
+    return QStandardPaths::writableLocation(QStandardPaths::DataLocation) +
+    QDir::separator() + QLatin1String("QML") + QDir::separator() + 
+    QLatin1String("OfflineStorage") + QDir::separator() + QLatin1String("Databases");
```
pvuorela: Indenting line continuations would be more readable.

pdf/pdfcanvas.cpp
pvuorela: Doesn't need anything special. Just drop prefix.

pdf/pdfcanvas.cpp
pvuorela: Thinking if it would be cleaner for this reason too, if instead of dtor the saving would be explicitly asked on inactive application or popped page.

pdf/pdfcanvas.cpp
```diff
@@ -187,6 +211,19 @@ int PDFCanvas::currentPage() const
...
 
+bool PDFCanvas::rememberPagePosition() const
+{
+    return d->rememberPagePosition;
+}
+
+void PDFCanvas::setRememberPagePosition(bool status)
+{
+    if (d->rememberPagePosition != status) {
+        d->rememberPagePosition = status;
```
pvuorela: This relies on timing a bit. Since setting the property just stores the value, it needs to be set before loading is finished. On worst case the order gets wrong and nothing is loaded.

pdf/pdfcanvas.cpp
```diff
@@ -535,5 +582,123 @@ void PDFCanvas::resizeTimeout()
...
+static bool _openDatabaseSync(const QString &dbname, const QString &dbversion,
+                              const QString &dbdescription, unsigned int estimatedsize)
+{
+    _initDatabasesPath();
+
+    QCryptographicHash md5(QCryptographicHash::Md5);
+    md5.addData(dbname.toUtf8());
+    QString dbid(QLatin1String(md5.result().toHex()));
+
+    QString basename = _databasesPath() + QDir::separator() + dbid;
```
pvuorela: Why md5 instead of just using plain name?

pdf/pdfcanvas.cpp
```diff
@@ -535,5 +582,123 @@ void PDFCanvas::resizeTimeout()
...
+
+    QSettings ini(basename + QLatin1String(".ini"), QSettings::IniFormat);
```
pvuorela: Configuration for configuration file seems a bit excessive :)

pdf/pdfcanvas.cpp
```diff
@@ -535,5 +582,123 @@ void PDFCanvas::resizeTimeout()
...
+    _ensureDbConnection();
+    QSqlDatabase db = QSqlDatabase::database(dbConnection);
+    _ensureTable(db);
+
+    QSqlQuery query = QSqlQuery(db);
+    query.prepare(QLatin1String("SELECT page, top, left, width FROM LastViewSettings WHERE file = ?"));
+    query.addBindValue(document->source());
+
+    if (query.exec() && query.next()) {
+        rememberPage  = query.value(0).toInt();
```
pvuorela: Pending could perhaps be better word for these?

pdf/pdfcanvas.cpp
dcaliste: I agree, didn't notice, sorry.

pdf/pdfcanvas.cpp
dcaliste: I've just copied everything from the LocalStorage QML source code from Qt. The idea was to have same database file, viewed from C++ and viewed from JS, in case later we want to address it in JS also for whatever reason. Do you think it's a bad idea ? It will indeed simplify the code, but we will loose JS compatibility.

pdf/pdfcanvas.cpp
dcaliste: Same same here, it's copied from the LocalStorage plugin of Qt. It can be removed if you think so. It's just that LocalStorage from JS always creates two files : the SQlite file and a ini file that store some meta information of the database.

pdf/pdfcanvas.cpp
dcaliste: Do you mean pendingPage instead of rememberPage ? Why not I can change.

pdf/pdfcanvas.cpp
dcaliste: I agree, but since the rememberPagePosition property is set when the PDFCanvas is built from QML, this worst case is currently not likely in the current PDFDocumentPage QML implementation. Indeed, when looking at pdfdocument.cpp source, one can see that loading is deferred until the PDF.Document QML object is completed. So it seems to me that when the document is loaded, the PDFCanvas object will already be completed also and its rememberPagePosition property already set.

I may be wrong though, what do you think ?

pdf/pdfcanvas.cpp
pvuorela: Firstly that's a bad idea license-wise. File here is GPLv2, Qt provides LGPL and GPLv3.

JS compatibility is only partially provided, QML side has some chance of adapting to new db scheme internally.

dcaliste: Alright, so I can change the name of the database file, and remove the .ini stuff.

pvuorela: What do you think about that aboutToQuit + page popping thing? With such change the older qml version could also suffice and it was simpler.

dcaliste: Ok, I'll try this in the old JS+QML approach. I'm not specifically in favour of one or the other approach. Anyway, it's not wasted time, I've learned how to do SQlite stuff from C++ and I can reuse this later.

dcaliste:
Superceded by #49 .

PR #49, opened 2015-10-16
[office] Implement persitent storage of page location for PDF documents (QML + JS).

dcaliste:
I've taken back the QML + JS implementation of the remember page position PR. I've corrected what has been discussed in #47:
* I've not included the setting page to turn this on and off. It's still dependent on a dconf key, but it's true by default.
* I've moved the handling of SQL storing / restoring from PDFView to the main PDFDocumentPage object. Like that, the PDFView is kept simple stupid (and possibly reusable for simple PDF display).
* The storing is done on page status becoming inactive and on Qt.application sending aboutToQuit. It seems to work like that.
* The restoring is done on a new signal from PDFView that is emitted once when the page sizes of the PDF document have been computed.

Sorry for this feature implementation taking so much review time from you. I appreciate your remarks and I think that the result looks better. What do you think ?

### DIFF
--- a/plugin/CMakeLists.txt
+++ b/plugin/CMakeLists.txt
@@ -16,6 +16,7 @@ install(FILES
     PDFDocumentPage.qml
     PDFDocumentToCPage.qml
     PDFView.qml
+    PDFStorage.js
     PresentationPage.qml
     PresentationThumbnailPage.qml
     SpreadsheetListPage.qml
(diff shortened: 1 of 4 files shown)


pvuorela:
I like this, doesn't depend on any unspecified behavior and is relatively simple.

Could you still adjust the order of content a bit by grouping properties, functions and child components etc. http://doc.qt.io/qt-5/qml-codingconventions.html
Granted that existing code doesn't always go like that.

dcaliste:
Ok, thank you for the review. I've moved the various parts trying to follow the code conventions you mentioned. The only exception is for getPagePosition() in PDFView.qml because this function is exactly the inverse of the goToPage() function, that is currently located at the end of the file, thus I prefer to keep them together.

pvuorela:
Merging. Thanks!

dcaliste:
Thank you.

PR #54, opened 2015-11-02
Enhanced file search UI 

pvuorela:


pvuorela:
@rainemak, @dcaliste

qml/FileListPage.qml
```diff
@@ -26,11 +26,21 @@ Page {
...
-    property string searchText
+    property string searchText: searchField.text
+    property bool searchEnabled
     property QtObject provider
 
     allowedOrientations: Orientation.All
 
+    onSearchEnabledChanged: {
+        if (pageStack.currentPage.status == PageStatus.Active) {
+            (searchEnabled ? searchField : page).focus = true
```
dcaliste: Tested with current 2.0.0.10, after choosing "show search" from the pulley, the searchfield appears, but not the virtual keyboard. I've checked that indeed searchField.focus is set to true on these lines, but maybe it's a forceActiveFocus() that should be called or something ?

dcaliste:
One minor additional note, maybe not related to the PR, but the pulley doesn't make any sound when closing on "show search" selection, while is emiting a closing sound when closing on "hide search" selection. I have no clue why it's like that !

pvuorela:
Changed to use forceActiveFocus().

Sound on closing is due to pulley menu hitting (multiple times, actually) onFinalPositionReached resulting from height transition on search field. That should be fixed in Silica.

qml/FileListPage.qml
```diff
@@ -43,48 +57,99 @@ Page {
...
+                                             qsTrId("sailfish-office-me-show_search")
+                                             //% "Hide search"
+                                           : qsTrId("sailfish-office-me-hide_search")
+                onClicked: page.searchEnabled = !page.searchEnabled
             }
         }
-        
-        ViewPlaceholder {
+
+        InfoLabel {
```
dcaliste: Just out of curiosity, why is it better to use directly a InfoLabel here, instead of a ViewPlaceHolder ? Is it for the custom 'y' property ?

dcaliste:
Thanks for the forceActiveFocus(). The overall PR is fine with me.

qml/FileListPage.qml
pvuorela: ViewPlaceHolder has a click animation which doesn't make sense here. Pulldown menu doesn't include any item that allows to create more content. Plus on search case even that wouldn't make sense.

Custom y value is another thing. No existing documents could have the label in the middle, but for search case it's better to have search indication closer to the search field.

dcaliste: Thanks, I see the first point, it's clear.

qml/FileListPage.qml
```diff
@@ -43,48 +57,99 @@ Page {
...
+            ignoreUnknownSignals: true
+            onContentYChanged: {
+                if (listView.contentY > (Screen.height / 2)) {
+                    searchField.focus = false
+                }
             }
         }
 
         PullDownMenu {
-            visible: listView.count > 0
```
dcaliste: Oh, sorry, forgot about this remark : should the pulley menu be visible (I mean the search pulley entry) when there is no document (I mean no document on device)? This line was there to ensure this at first. But, I've just noticed, it is broken, since the pulley disappears when there is no document in the list (after search filtering). So removing the 'visible' binding as you do in the PR may be a safe fallback for now.

pvuorela: Menu could indeed be omitted if device doesn't have any document files, but having a special case for that didn't seem important enough. As you said, the existing value had to be removed as it was hiding menu during filtering.

qml/FileListPage.qml
```diff
@@ -26,11 +26,25 @@ Page {
...
+            if (searchEnabled) {
+                searchField.forceActiveFocus()
+            } else {
+                searchField.focus = false
+            }
+        }
+        if (!searchEnabled) {
+            searchField.text = ""
+        }
+    }
```
rainemak: Just wondering... Maybe part of above could be turned into something like

```
SearchField {
    focus: pageStack.currentPage.status == PageStatus.Active && page.searchEnabled
}
```

qml/FileListPage.qml
```diff
@@ -43,48 +57,99 @@ Page {
...
+
+                // We prefer lowercase
+                inputMethodHints: Qt.ImhNoAutoUppercase | Qt.ImhPreferLowercase | Qt.ImhNoPredictiveText
+                EnterKey.iconSource: "image://theme/icon-m-enter-close"
+                EnterKey.onClicked: focus = false
+
+                Behavior on opacity { FadeAnimation { duration: 150 } }
+                Behavior on height {
+                    NumberAnimation {
+                        duration: 150
```
rainemak: Did you check durations with design?

qml/FileListPage.qml
rainemak: Ah, would require at least focus out chance to line 91.

rainemak: Ignore... Seems that it's not that easy to change this to be a binding.

qml/FileListPage.qml
pvuorela: Same as elsewhere. Shown the end up.

rainemak:
LGTM

PR #55, opened 2015-11-03
[office] Implement partial page rendering to allow unlimited zoom.

dcaliste:
There is a TJC question about being able to increase the zoom level for PDFs : https://together.jolla.com/question/1763/unlimited-zoom-needed-in-document-viewer/

The limit is currently hardcoded in PDFView.qml, and set to 2.5. I understand that this limit is used to avoid having too big textures and limit memory consumption. But that would be convenient to be able to change this value anyway, knowing what we're doing. There is even a patch for patch manager to do this : https://openrepos.net/content/ancelad/patch-gallery-image-zoom . The patch, actually, uses a dconf key value instead of the hardcoded 2.5. So like that, one can create a GUI or even change via command line the setting.

The PR, I propose, just changes the hardcoded value for a dconf key (like the idea in the patch). The dconf key is loaded in PDFDocumentPage.qml (like the dconf key for the remember page position) so PDFView.qml is kept simple.

What do you think, this PR may conciliate the memory consumption limit for everyone and the need of some to easily raises the limit in some cases (in a more robust way than a patch) ?

### DIFF
--- a/pdf/pdfcanvas.cpp
+++ b/pdf/pdfcanvas.cpp
@@ -18,6 +18,7 @@
 
 #include "pdfcanvas.h"
 
+#include <QtMath>
 #include <QtCore/QTimer>
 #include <QtCore/QPointer>
 #include <QtGui/QPainter>
@@ -34,14 +35,17 @@ struct PDFPage {
     PDFPage()
         : index(-1)
         , requested(false)
-        , rectPlaceholder(false)
+        , renderWidth(0)
         , texture(nullptr)
     { }
 
     int index;
     QRectF rect;
+
     bool requested;
-    bool rectPlaceholder;
+
+    int renderWidth;
+    QRect textureArea;
     QSGTexture *texture;
 
     QList<QPair<QRectF, QUrl> > links;
@@ -89,6 +93,14 @@ public:
     QList<QSGTexture *> texturesToClean;
     QPointer<QQuickWindow> connectedWindow;
 
+    void cleanPageTexturesLater(PDFPage &page)
+    {
+        if (page.texture) {
+            texturesToClean << page.texture;
+            page.texture = nullptr;
+        }
+    }
+
     void cleanTextures()
     {
         foreach (QSGTexture *texture, texturesToClean)
@@ -114,7 +126,7 @@ PDFCanvas::~PDFCanvas()
     for (int i = 0; i < d->pageCount; ++i) {
         PDFPage &page = d->pages[i];
         delete page.texture;
-        page.texture = 0;
+        page.texture = nullptr;
     }
 
     delete d->resizeTimer;
@@ -135,6 +147,7 @@ void PDFCanvas::setFlickable(QQuickItem *f)
         d->flickable = f;
         connect(d->flickable, SIGNAL(contentXChanged()), this, SLOT(update()));
         connect(d->flickable, SIGNAL(contentYChanged()), this, SLOT(update()));
+        connect(d->flickable, SIGNAL(widthChanged()), this, SLOT(update()));
 
         emit flickableChanged();
     }
@@ -256,8 +269,9 @@ void PDFCanvas::layout()
         page.links = links.values(i);
         page.requested = false; // We're cancelling all requests below
         if (d->pages.contains(i)) {
+            page.renderWidth = d->pages.value(i).renderWidth;
+            page.textureArea = d->pages.value(i).textureArea;
             page.texture = d->pages.value(i).texture;
-            page.rectPlaceholder = d->pages.value(i).rectPlaceholder;
         }
         d->pages.insert(i, page);
 
@@ -348,12 +362,15 @@ QPointF PDFCanvas::fromPageToItem(int index, const QPointF &point) const
                    point.y() * page.rect.height() + page.rect.y());
 }
 
-void PDFCanvas::pageFinished(int id, QSGTexture *texture)
+void PDFCanvas::pageFinished(int id, int pageRenderWidth,
+                             QRect subpart, QSGTexture *texture)
 {
     PDFPage &page = d->pages[id];
 
-    if (page.texture)
-        d->texturesToClean << page.texture;
+    d->cleanPageTexturesLater(page);
+
+    page.renderWidth = pageRenderWidth;
+    page.textureArea = subpart;
     page.texture = texture;
     page.requested = false;
 
@@ -375,12 +392,9 @@ void PDFCanvas::sceneGraphInvalidated()
     d->cleanTextures();
     for (int i = 0; i < d->pageCount; ++i) {
         PDFPage &page = d->pages[i];
-        if (page.texture) {
-            delete page.texture;
-            page.texture = 0;
-        }
+        delete page.texture;
+        page.texture = nullptr;
         page.requested = false;
-        page.rectPlaceholder = false;
     }
 }
 
@@ -409,13 +423,19 @@ QSGNode* PDFCanvas::updatePaintNode(QSGNode *node, QQuickItem::UpdatePaintNodeDa
         visibleArea.width() * 5,
         visibleArea.height() * 5,
     };
+    QRect textureLimit = {
+        0, 0,
+        2.5 * qMin(window()->width(), window()->height()),
+        2.5 * qMin(window()->width(), window()->height())
+    };
+    float renderingRatio = float(d->renderWidth) / width();
 
     QSGNode *root = static_cast<QSGNode*>(node);
     if (!root) {
         root = new QSGNode;
     }
 
-    QList<QPair<int, int> > priorityRequests;
+    QList<QPair<int, QPair<int, QRect> > > priorityRequests;
     int currentPage = d->currentPage;
     qreal maxVisibleArea = 0.;
 
@@ -425,21 +445,43 @@ QSGNode* PDFCanvas::updatePaintNode(QSGNode *node, QQuickItem::UpdatePaintNodeDa
         bool loadPage = page.rect.intersects(loadedArea);
         bool showPage = page.rect.intersects(visibleArea);
 
-        if (loadPage) {
-            if ((!page.texture)
-                || (page.texture && page.texture->textureSize().width() != d->renderWidth)) {
-                if (page.requested && showPage) {
-                    priorityRequests << QPair<int, int>(i, d->renderWidth);
-                } else if (!page.requested) {
-                    d->document->requestPage(i, d->renderWidth, window());
-                    page.requested = true;
-                }
+        // Current rendering in pixels is done with a width of
+        // d->renderWidth which can be different than actual width()
+        // when zooming.
+        QRect pageRect = {
+            0, 0, d->renderWidth, int(page.rect.height() * renderingRatio)
+        };
+        QRect showableArea = {
+            int(renderingRatio * (visibleArea.x() - float(window()->width() / 4.) - page.rect.x())),
+            int(renderingRatio * (visibleArea.y() - float(window()->height() / 4.) - page.rect.y())),
+            int(renderingRatio * (visibleArea.width() + float(window()->width() / 2.))),
+            int(renderingRatio * (visibleArea.height() + float(window()->height() / 2.)))
+        };
+        showableArea = showableArea.intersected(pageRect);
+        // Limit showableArea with textureLimit to avoid looping request
+        // for something too big.
+        textureLimit.moveCenter(showableArea.center());
+        showableArea = showableArea.intersected(textureLimit);
+
+        if (showPage && (page.texture == nullptr
+                         || page.renderWidth != d->renderWidth
+                         || (page.renderWidth == int(width()) &&
+                             !page.textureArea.contains(showableArea)))) {
+            if (!page.requested) {
+                d->document->requestPage(i, d->renderWidth, window(), textureLimit);
+                page.requested = true;
             }
-        } else {
-            if (page.texture) {
-                d->texturesToClean << page.texture;
-                page.texture = 0;
+            priorityRequests << QPair<int, QPair<int, QRect> >(i, QPair<int, QRect>(d->renderWidth, textureLimit));
+        } else if (loadPage && !showPage
+                   && !page.requested && (page.renderWidth != d->renderWidth)) {
+            textureLimit.moveTo(0, 0);
+            // We preload full page only if they can fit into texture.
+            if (textureLimit.contains(pageRect)) {
+                d->document->requestPage(i, d->renderWidth, window());
+                page.requested = true;
             }
+        } else if (!loadPage) {
+            d->cleanPageTexturesLater(page);
 
             // Scrolled beyond where this page is needed, skip it.
             if (page.requested) {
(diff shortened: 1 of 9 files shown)


pvuorela:
For alternative, maybe it could be more beneficial to have dynamic maximum zoom level, calculated from page size? I.e. it would be possible to zoom A2 sized document mentioned on t.j.c. more than an A4 one.

Memory consumption can be limited if only part of the page is rendered, I'd guess.

dcaliste:
I've tried to do something with the dynamical maximum zoom level, implying as less modifications as possible.

The second commit ensures that for the biggest page of the document, one can zoom to a resolution that is the same than the possible zooming resolution with an A4 sheet. The zooming possibility for A4 pages and smaller is still governed by the maxZoomLevel as introduced earlier (or hardcoded 2.5).

I'm not quite confident with the quality of the PR, what do you think ? Is it not bringing to much bloat in PDFView.qml/zoom() function ?

I prefer to keep the maxZoomLevel config, because you'll always have people wanting to zoom more on a given document that the hardcoded 2.5.

About the page partial rendering, I don't know if Poppler can render subparts of pages. Additionally, it would mean some work to pdfcanvas.cpp to handle several tiles per page. Besides, that work may be of some interest. I'm lightly looking at LibreOfficeKit and its tile rendering capabilities. It may be of a possible choice beside Calligra to render office document (and later edit them maybe??).

pdf/pdfcanvas.h
```diff
@@ -33,6 +33,7 @@ class PDFCanvas : public QQuickItem
     Q_PROPERTY( QColor linkColor READ linkColor WRITE setLinkColor NOTIFY linkColorChanged )
     Q_PROPERTY( QColor pagePlaceholderColor READ pagePlaceholderColor WRITE setPagePlaceholderColor NOTIFY pagePlaceholderColorChanged )
     Q_PROPERTY( int currentPage READ currentPage NOTIFY currentPageChanged )
+    Q_PROPERTY(QSizeF maxPageSize READ maxPageSize NOTIFY pageSizesReady)
```
pvuorela: Signal can be emitted when maxSize didn't change. Small detail, but could make sense to have a separate signal for that.

plugin/PDFView.qml
```diff
@@ -42,8 +43,12 @@ SilicaFlickable {
             return width
         }
 
-        if (value > width * 2.5) {
-            return width * 2.5
+        // We guarantee that one can zoom at least to maxZoomLevel times
+        // for an A4 width page (21 cm = 595.56 1/72th inch).
+        var maxWidth = width * maxZoomLevel *
+                       Math.max(pdfCanvas.maxPageSize.width / 595.56, 1.)
+        if (value > maxWidth) {
```
pvuorela: Cosmetics, but all rest can be replaced with return Math.min(value, maxWidth)

pvuorela:
Testing this with a big pdf shows that the memory consumption problem is real. With tablet and no zooming I see acceptable 84MB, zooming to current maximum eats 217MB in total, and then with this PR zooming more it jumps quickly to 800MB and then failing for memory allocation. Unfortunately I don't think that's acceptable and rendering side needs to compensate memory usage somehow.

> I'm not quite confident with the quality of the PR, what do you think ? Is it not bringing to much bloat in PDFView.qml/zoom() function ?

Nah, as per changed lines this is quite small. 

> I prefer to keep the maxZoomLevel config, because you'll always have people wanting to zoom more on a given document that the hardcoded 2.5.

There are always people wanting to do something strange. If the percentage is small enough, it doesn't matter too much. Can live with it, but would consider it an "unstable" setting, subject to be removed at any point.

> About the page partial rendering, I don't know if Poppler can render subparts of pages. Additionally, it would mean some work to pdfcanvas.cpp to handle several tiles per page. Besides, that work may be of some interest. I'm lightly looking at LibreOfficeKit and its tile rendering capabilities. It may be of a possible choice beside Calligra to render office document (and later edit them maybe??).

Had something like tile-based rendering in mind. Maybe switching mode when zoom level is big enough. 

Poppler has parameters to render only part of the page:
http://people.freedesktop.org/~aacid/docs/qt5/classPoppler_1_1Page.html#a43bb866fa1f848156799eb8d9e38d0aa

pvuorela:
> Had something like tile-based rendering in mind. Maybe switching mode when zoom level is big enough. 

Or maybe just restricting rendered part of a page when zoomed in.

dcaliste:
I thought about it over night. I don't have clear conclusions yet and I will need to test several things, but :

Currently the textures for the loaded area is 5 time the visible area. So ideally, we should be able to have a maximum of 5×5×screen size in memory as (tiled) textures, for the phone it means 960×540×4(RGBA)×25 ~ 51MB.

I don't know where all the additional MB, we're seing currently, are coming from. Currently, we are rendering per page, so, for PDF with large pages, the 5 time area, may expand slightly on the previous and next page, even when zoom in, requiring to render 3 time more data in memory, but it still means 150MB, so much less than measured. On tablet, the pixel density is bigger, that may justify the > 800MB consumption.

Since we load 5 time the visible area, it means that we always need to render full page width for zoom level up to 3 (worst case, on a border). So tile rendering is not necessary for width direction for zoom levels lower than 3. On height direction, tile rendering can be beneficial from any zoom level.

Final question for me: what would be the best approach ? Either implement height tiling to avoid the additional 3 time of the simple above calculation in the worst case, or implement dynamic range, limiting the 5 time rendering area with a memory limit for instance.

Need to play a bit more !

pdf/pdfcanvas.h
dcaliste: I fact no, if I'm not mistaken, the signal corresponds to the property. Maybe the naming scheme is bad. The signal "pageSizesReady" is emitted only when the thread finishes to get the physical sizes of every pages. So the maximum of all the sizes is indeed known at that time only.

In fact, the pageSizesReady signal, is the one that we introduced for the remember page position PR. It is just that now there is a property attached to it, and not only a signal. Thus maybe the naming issue.

pvuorela:
> Currently the textures for the loaded area is 5 time the visible area. So ideally, we should be able to have a maximum of 5×5×screen size in memory as (tiled) textures, for the phone it means 960×540×4(RGBA)×25 ~ 51MB.

5x5 rendering area seems like an overkill. On FullHD that'll eat 200MB which is quite much. Not considering zooming, having previous and next page loaded might be enough already. Rendering adjacent out of screen pages with lower resolution might be something to consider too.

> Final question for me: what would be the best approach ? Either implement height tiling to avoid the additional 3 time of the simple above calculation in the worst case, or implement dynamic range, limiting the 5 time rendering area with a memory limit for instance.

But limit on three wouldn't anyway fix the zoom issue on A2 pdfs and such that were the main driver here?

One additional unknown is how well Poppler really works with box parameters. Documentation warns "The parameter (x, y, w, h) are not well-tested. Unusual or meaningless parameters may lead to rather unexpected results." On worst case that can mean tile borders visible, though depends on what is an "unusual" value.

To rephrase my previous comment, one another option could be having a sliding window per page, say twice the screen resolution (max when zoomed). Content of that would be re-rendered when view moved enough. Worst case memory consumption 8 times screen resolution, less if window size is restricted more.

dcaliste:
Thanks for the comments. I'll continue thinking on this in the coming weeks.

dcaliste:
Several months have passed since we discuss how to implement the page partial rendering but I finally had time to gave a try for these ideas. Previously, the page rendering was done like that:
- each page that intersects with the visible area are rendered full size;
- each page that intersects with a buffer around the visible area are stored full size.
The buffer around the visible area accounts for twice the screen size in every directions. It means that for a PDF which renders a page per visible area has two pages in advance when scrolling.
- pages are rerendered when zoom level change and flush from memory when out of the buffer zone.

I've changed this behaviour a bit:
- each page that intersects with the visible area are rendered partial size (partial is defined as the smallest area between the page at a given zoom level and twice the screen size, called textureLimit in the code), this ensure that a low zoom level, the full page is loaded, while at high zoom levels, only the visible portion of the page plus a small buffer of half the screen size in every directions is loaded.
- each page that is in the loaded area, but not shown are stored only if the full page can fit into the textureLimit.
- the textures are rerendered when the zoom level change, and now also when the texture of the visible page does not contain the screen area anymore (plus a buffer of screen size / 4 in every directions to avoid too much to see the texture edge when scrolling).

As a result, the scrolling should be identical as before for low zoom levels (2 pages are cached in advance) but for higher zoom levels one can see often the texture edge because of the texture limit that is imposed on a single texture (I've kept the one texture per page convention). On the memory part, one can zoom as much as possible (with a soft arbitrary limit at 10 now) with the memory remaining constant.

I've changed quite a lot ```pdfcanvas.cpp:updatePaintNode()``` method, from the logic of rendering request and from the layouting point of view. I think that the PR requires much tests before deciding if it solves the memory and texture size issues on zoom. Tell me your opinion, I would be suprise if there is no side effect to the current implementation.

pvuorela:
Zoooming in and out I'm getting opengl errors together with black textures like before zoom level was restricted a while ago. Adding debug to print resulting QImage sizes I see that zooming out fast results in quite big textures.

pdf/pdfrenderthread.cpp
```diff
@@ -279,19 +279,21 @@ void PDFRenderThread::cancelRenderJob(int index)
...
         if (j->type() == PDFJob::RenderPageJob) {
             RenderPageJob *rj = static_cast<RenderPageJob *>(j);
             if (rj->m_index == index && rj->renderWidth() == size) {
-                if (it != d->thread->jobQueue->begin()) { // If it is already at the front, just abort..
-                    d->thread->jobQueue->erase(it);
-                    d->thread->jobQueue->push_front(j);
-                }
-                return;
+                it = d->thread->jobQueue->erase(it);
+                last = rj;
```
pvuorela: Seems this is no longer "prioritizeJob" but rather prioritize and remove duplicates with the assumption that last instance is best one. Could be better if duplicates weren't added to queue at all.

pdf/pdfcanvas.cpp
```diff
@@ -348,12 +357,14 @@ QPointF PDFCanvas::fromPageToItem(int index, const QPointF &point) const
                    point.y() * page.rect.height() + page.rect.y());
 }
 
-void PDFCanvas::pageFinished(int id, QSGTexture *texture)
+void PDFCanvas::pageFinished(int id, int resolution, QRectF subpart, QSGTexture *texture)
```
pvuorela: Resolution is a bit ambiguous word, maybe just renderWidth?

pdf/pdfcanvas.cpp
```diff
@@ -425,21 +433,45 @@ QSGNode* PDFCanvas::updatePaintNode(QSGNode *node, QQuickItem::UpdatePaintNodeDa
...
+            visibleArea.width() + float(window()->width() / 2.),
+            visibleArea.height() + float(window()->height() / 2.)
+        };
+        showableArea = page.rect.intersected(showableArea);
+        // remove one pixel margin to avoid rounding errors.
+        showableArea = QRectF((showableArea.x() + 1) / page.rect.width(),
+                              (showableArea.y() - page.rect.y() + 1) / page.rect.height(),
+                              (showableArea.width() - 2) / page.rect.width(),
+                              (showableArea.height() - 2) / page.rect.height());
+
```
pvuorela: These reduced coordinates are causing me headache :) Page.rect is pixels, .textureArea is reduced, visibleArea and loadedArea are pixels, textureLimit and showableArea are reduced. Do wonder if it would be simpler just playing with pixels on canvas side.

pdf/pdfcanvas.cpp
```diff
@@ -469,55 +501,61 @@ QSGNode* PDFCanvas::updatePaintNode(QSGNode *node, QQuickItem::UpdatePaintNodeDa
...
+            // |  |-n
+            // |  | |- link1
+            // |  | |- link2...
+            QSGSimpleRectNode *bg = static_cast<QSGSimpleRectNode*>(t->firstChild());
+            if (!bg) {
+                bg = new QSGSimpleRectNode;
+                t->appendChildNode(bg);
             }
+            bg->setRect(0., 0., page.rect.width(), page.rect.height());
+            bg->setColor(d->pagePlaceholderColor);
```
pvuorela: Shouldn't change after construction, so could be do only once?

dcaliste: I agree, I've moved it some lines above, when bg object is created. I'm just wondering when ambiance is changed if there can be a case when the page place holder is visible and thus not in the right colour. The only case I imagine is ambiance changed by a timer and the page place holder is visible. But this is very rare situation…

pdf/pdfcanvas.cpp
dcaliste: You're right. I kept ```pageRenderWidth``` to emphase that this is the total page render width and not the texture render width.

pdf/pdfcanvas.cpp
dcaliste: I think you're right. At the beginning , I wanted to keep the textureArea in reduced, because I was thinking for the annotation partial rendering, where we go patching a piece of the page, and we want this patch to be applied on different zoom levels. But that's wrong in fact. When zooming, the patches will be removed and the full page re-rendered.

So the textureArea can be kept in pixel, synchronized with the page width, less headache, I agree. I'll change this when I have a bit of free time.

pdf/pdfrenderthread.cpp
dcaliste: You're right and this prioritize stuff is still not optimal. Currently for instance, when scrolling fast to the page, let say 5, the rendering queue is filled up with render(3), render(4), render(5), re,der(6), render(7) ans the 5th page is not in the top of the list.

I will try to remove this prioritize() function in favor of a priotirtize argument to the requestPage(). Then renderPage(i) do the following:
* if page i is already in the queue, waiting, then its subpart is updated with the new one; otherwise it is appended as a new request.
* if the priority argument is set, then the job in the queue is moved to the top.

pdf/pdfrenderthread.cpp
```diff
@@ -276,22 +276,42 @@ void PDFRenderThread::cancelRenderJob(int index)
...
+        if (j->type() == PDFJob::RenderPageJob &&
+            static_cast<RenderPageJob *>(j)->m_index == index) {
+            RenderPageJob *rj = static_cast<RenderPageJob *>(j);
+            rj->changeRenderWidth(size);
+            rj->m_subpart = subpart;
+            return;
+        }
+        ++it;
+    }
+    locker.unlock();
```
pvuorela: Can skip, QMutexLocker automatically unlocks.

pdf/pdfrenderthread.cpp
dcaliste: Ok, I've finally kept the prioritize() function as it is, with an additional parameter subpart, that allow to update and prioritize the request at the same time.

pdf/pdfcanvas.cpp
dcaliste: Ok, done, page.textureArea is now a QRect (so storing values in integer pixels).

pdf/pdfrenderthread.cpp
dcaliste: Sorry, this was WIP, not ready yet.

dcaliste:
Ok, the last push now should correct the texture errors as existing in the previous PR state.

You can test and review and give me your opinion.

pvuorela:
Looks lot cleaner now. Thanks for polishing. Haven't either gotten those OpenGL errors anymore.

Tried adding some debug to follow what gets rendered. At RenderPageJob::run():
qDebug() << "+++ rendering page" << m_index << m_width << m_subpart;
-> 
[D] RenderPageJob::run:65 - +++ rendering page 0 720 QRect(-360,-814 1440x2560) 
[D] RenderPageJob::run:65 - +++ rendering page 0 720 QRect(-360,-814 1440x2560) 

And if zooming, closing and reopening:
[D] RenderPageJob::run:65 - +++ rendering page 0 720 QRect(-507,-845 1440x2560) 
[D] RenderPageJob::run:65 - +++ rendering page 0 720 QRect(-507,-845 1440x2560) 
[D] RenderPageJob::run:65 - +++ rendering page 0 6591 QRect(1237,2703 1440x2560) 
[D] RenderPageJob::run:65 - +++ rendering page 0 6591 QRect(1237,2703 1440x2560) 

Seems to do some unnecessary work.

Also couple rendering issues. First, zoomed portrait, turned landscape -> part of document on right side is missing. Second, trying above more on different zoom levels, I ended up having device in landscape and rendering the same sub rect of first page in loop, over and over again.

dcaliste:
> Seems to do some unnecessary work.

It was pre-existing to this PR, in my opinion. After check, I've understand that the pageSizeRequest is done twice, because the geometry change (set width) and the document loaded signals ask for layouting, which in turn call the pageSizeRequest twice. Then after the page size is known, the layout is called twice, quickly enough for the page not to be rendered yet, but since layout set back the requested to false, the request is done twice with the same size.

Well, the second commit appended to this PR should cure this behaviour.

dcaliste:
> First, zoomed portrait, turned landscape -> part of document on right side is missing.

The update() was not called when the flickable width was changed, and the content was not changing its width. I've added a connection to the flickable.

> Second, trying above more on different zoom levels, I ended up having device in landscape and rendering the same sub rect of first page in loop, over and over again.

The landscape was requesting a texture greater than the size authorized by textureLimit, so the request was always looping. I've added an .intersected() call to the showableArea with the textureLimit so the showableArea ('''i.e.''' the minimum size to be requested) in always bounded by the texture limit.

dcaliste:
Sorry, I noticed a load issue yesterday evening that arises from the correction of the pageSizeRequest correction. It's corrected now.

dcaliste:
As an after thought, I've added a third commit to use a configuration key to set the maximum zoom level, so less values are hard coded. I think, it's wise to default to a not too big value, so one can not arrive to a situation where the screen is full white because one zoom accidentaly too much in an area between text. But a configuration key can be changed by advanced users, or by a UI available in the store or elsewhere, to match their needs.

pvuorela:
Orientation change not re-rendering fixed and not able to get rendering loop either. Good. 

Don't still see configuration too important. User himself zoomed into white area is able to zoom out as well. What might matter is exiting document while zoomed so reopening next time can have document wherever. Could have separate zoom restriction on reopening if that really matters. Or simply don't care. Having said that, guess can live with the dconf key too.

Getting close to merging, though noticed one more detail. Opening pdf fully zoomed out:
[D] RenderPageJob::run:65 - +++ rendering page 0 720 QRect(-540,-391 1800x1800) 
[D] RenderPageJob::run:65 - +++ rendering page 1 720 QRect(-540,-625 1800x1800) 
[D] RenderPageJob::run:65 - +++ rendering page 2 720 QRect(0,0 0x0) 
[D] RenderPageJob::run:65 - +++ rendering page 3 720 QRect(0,0 0x0) 

Slowly moving down, first this:
[D] RenderPageJob::run:65 - +++ rendering page 4 720 QRect(0,0 0x0) 
And then:
[D] RenderPageJob::run:65 - +++ rendering page 2 720 QRect(-540,-739 1800x1800) 

So the third (2) page first gets rendered without subrect and same area is rerendered with the subrect. Continuing further down first one full page farther away is rendered and then the one coming to view is rerendered.

dcaliste:
> So the third (2) page first gets rendered without subrect and same area is rerendered with the subrect.

Good catch. The issue was that when requesting full page rendering, the textureArea rectangle was left empty. But textureArea is used when the page is visible to know if the current texture contains the visible part of the page… Corrected by exporting correctly the textureArea when asking for full page rendering in pdfjob.cpp.

For the maximum zoom level and the configuration option, I've no definite position. I like to have the configuration key, in case. For me, the current maximum zoom level of 10 is fine (_i.e._ allow to zoom well on sometimes tiny QR codes for e-tickets) but a lower value would be fine also (while now it should have no impact on memory consumption).

pvuorela:
I think this should be good to go. Got one rerendering when scrolling downwards, but that's not too much. Big thanks!

dcaliste:
Ok, thank you for the review. I can give a look to the rerendering issue in one week. I'm now on vacation without my computer.

PR #56, opened 2015-11-04
Adjusted pdf error and locking state UIs 

pvuorela:
ViewPlaceHolder didn't really work for these. Now avoiding whole split view if pdf is not loaded.

Some other cleanups and fixes also included.

### DIFF
--- a/plugin/PDFDocumentPage.qml
+++ b/plugin/PDFDocumentPage.qml
@@ -28,6 +28,10 @@ DocumentPage {
 
     property var _settings // Handle save and restore the view settings using PDFStorage
 
+    busy: (!pdfDocument.loaded && !pdfDocument.failure) || pdfDocument.searching
+    source: pdfDocument.source
+    indexCount: pdfDocument.pageCount
+
     function savePageSettings() {
         if (!rememberPositionConfig.value) { return }
         
@@ -215,10 +219,6 @@ DocumentPage {
         id: pdfDocument
         source: base.path
     }
-
-    busy: (!pdfDocument.loaded && !pdfDocument.failure) || pdfDocument.searching
-    source: pdfDocument.source
-    indexCount: pdfDocument.pageCount
 
     ConfigurationValue {
         id: rememberPositionConfig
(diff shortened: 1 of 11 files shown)


pvuorela:
@rainemak, @dcaliste

dcaliste:
PDFDocumentPage.qml is so much cleaner like that and easier to read. Using the back navigation is also an infinitely better idea than the tap to obtain the split view, much cleaner. Having the forceActiveFocus() is nice also. Thanks.

I've tried with locked down documents, it's working as expected. We're still missing a way to warn the user that the entered password was wrong, but it was lacking since my first implementation, so it's the matter of another PR.

Fine with me for this PR (I take good note of using brackets on multi-line JS in QML for next times !).

plugin/PDFDocumentPage.qml
```diff
@@ -55,67 +62,41 @@ DocumentPage {
...
 
+    Binding {
+        target: base
+        property: "forwardNavigation"
+        value: false
+        when: (pdfDocument.failure || pdfDocument.locked)
+    }
+
+    Loader {
+        parent: base
```
rainemak: Why do you have explicit parent over here?

plugin/ToolBar.qml
```diff
@@ -55,21 +57,22 @@ PanelBackground {
...
+            if (!flickable.movingVertically)
+                return
 
-        _dragUp = !autoHide || (flickable.contentY < _previousContentY)
-        if (_dragUp) autoHideTimer.restart()
+            _dragUp = !autoHide || (flickable.contentY < _previousContentY)
 
-        _previousContentY = flickable.contentY
+            if (_dragUp)
+                autoHideTimer.restart()
```
rainemak: I'd be tempted to add curly brackets here for the if-statement.

plugin/PDFDocumentPage.qml
pvuorela: Because DocumentPage has default property that otherwise parents to container item.

plugin/ToolBar.qml
pvuorela: Not considering relevant.

plugin/PDFDocumentPage.qml
rainemak: Sure. I just read Page :)

rainemak:
LGTM

PR #57, opened 2015-11-12
Polish PDF search functionality

pvuorela:
- Refactored Toolbar not to force PDF content clipping
- Added enter key icon for PDF search field

### DIFF
--- a/plugin/ToolBar.qml
+++ b/plugin/ToolBar.qml
@@ -22,49 +22,38 @@ import Sailfish.Silica 1.0
 PanelBackground {
     id: toolbar
 
-    property var flickable: undefined
-    property int parentHeight
-    property bool hidden
-    property bool autoHide: true
+    property Item flickable
+    property bool forceHidden
+    property bool autoShowHide: true
+    property int offset: _active && !forceHidden ? height : 0
 
+    property bool _active
     property int _previousContentY
-    property bool _dragUp
 
-    anchors.top: flickable.bottom
-
-    onAutoHideChanged: if (autoHide && !autoHideTimer.running) autoHideTimer.start()
-
-    states: [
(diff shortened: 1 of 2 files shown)


pvuorela:
@rainemak & @dcaliste

plugin/ToolBar.qml
```diff
@@ -22,49 +22,39 @@ import Sailfish.Silica 1.0
 PanelBackground {
     id: toolbar
 
-    property var flickable: undefined
-    property int parentHeight
-    property bool hidden
+    property Item flickable
+    property bool forceHidden
     property bool autoHide: true
+    property bool trackMovement: true
```
dcaliste: Didn't test, but from reading it seems to me that trackMovement is redundant with autoHide because in PDFDocumentPage.qml trackMovement is bind to autoHide. But besides, in a generic case the two properties still have sense : we want to appear so trackMovement is true, but don't want to disappear after, so autohide is false.

Maybe this case is too generic and Toolbar.qml may be even further simplified ?

dcaliste:
Besides the above note, the simplification in Toolbar.qml is much welcomed. Congrats.

plugin/ToolBar.qml
pvuorela: Yes, this is redundant. Introduced because I couldn't come up with a single word that would cover both cases. Hmm, stateChangesEnabled?

dcaliste: What about autoShowHide ? Too ugly maybe, but contains auto which suggests well the behaviour.

pvuorela: Good enough, sold.

pvuorela:
Combined redundant properties.

PR #59, opened 2015-11-13
Refactor and implement error state for cover

pvuorela:


pvuorela:
@rainemak & @dcaliste

qml/CoverFileItem.qml
```diff
@@ -0,0 +1,56 @@
...
+    Image {
+        id: icon
+
+        anchors {
+            left: parent.left
+            leftMargin: Theme.paddingLarge
+            verticalCenter: root.multiLine ? undefined : parent.verticalCenter
+        }
+        source: root.iconSource !== "" ? root.iconSource
+                                       : "image://theme/icon-l-document"
```
dcaliste: Should we profit of this refactoring, to change icon-l-document to the -m- version ? I have some postscript files that are listed by tracker as documents, so that appear in the list and the large icon is quite out of place compared to the other mime type icons.

qml/FileListPage.qml
```diff
@@ -162,15 +162,9 @@ Page {
...
-                states: State {
-                    when: icon.fileMimeType === ""
-                    PropertyChanges {
-                        target: icon
-                        source: "image://theme/icon-l-document?"
-                                + (listItem.highlighted ? Theme.highlightColor : Theme.primaryColor)
-                    }
-                }
+                source: fileMimeType !== "" ? fileMimeType
+                                            : ("image://theme/icon-l-document?"
```
dcaliste: Here it's another place with the -l- icon.

qml/CoverFileItem.qml
pvuorela: Uh, doesn't look good indeed. However, that seems to be case only in main page file list, not in cover, so I don't think it belongs in this PR. I'll make a request for missing postscript icon.

Also we are currently missing scalability support for these icons. Commonly Sailfish apps use icons from theme which are scaled according to device screen, but here non-scaled png files are used.

dcaliste: Yes, the size -l- issue is visible only in the main page, not in cover since cover resize icons (notice this after). The comment was position on the first -l- I've found, so quite misplaced !

Not having a post-script icon is not much of an issue currently since post-script are not handled anyway… Mentioning them was just to give a use case for when the icon-l-document is actually shown.

For the scalable icons, one should find the SVG sources of the current PNG files and put them in the images/ folder. Then generate whatever sizes requested by scalability and use the new SAILFISHAPP_ICONS variable in the pro file. Or is there a possibility to add images to an icon theme ? What is the best way to go ? Anyway, first one need the SVG sources…

pvuorela: Can be added to theme, but not sure if these belong there. Been requesting new SVG file and once I get some, I can proceed. Separate issue anyway.

dcaliste: Sure SVG files are another matter.

Don't you think that you can change -l- to -m- for the icon-l-document theme file in FileListPage.qml, so all icon sizes will be consistent (before becoming scalable later) ? After all, the commit https://github.com/pvuorela/sailfish-office/commit/fdc45f973103dd7e9edcc5d4368673eb26723f68 is already not dealing with cover stuff :)

pvuorela:
There, now smaller icon in main list.

plugin/PDFDocumentPage.qml
```diff
@@ -93,6 +94,11 @@ DocumentPage {
     PDFView {
         id: view
 
+        // for cover state
+        property bool contentAvailable: pdfDocument.loaded && !(pdfDocument.failure || pdfDocument.locked)
+        property alias title: base.title
+        property alias mimeType: base.mimeType
```
rainemak: These all look like readonly properties.

qml/CoverFileItem.qml
```diff
@@ -0,0 +1,56 @@
...
+        anchors {
+            left: icon.right
+            leftMargin: Theme.paddingMedium
+            verticalCenter: root.multiLine ? undefined : parent.verticalCenter
+            right: parent.right
+            rightMargin: Theme.paddingLarge
+        }
+
+        truncationMode: root.multiLine ? TruncationMode.None : TruncationMode.Fade
+        wrapMode: root.multiLine ? Text.WrapAnywhere : Text.NoWrap
```
rainemak: No sure how have we treated multiLine cover texts but I'd try to with "Text.Wrap" when multiLine enabled.

qml/CoverFileItem.qml
```diff
@@ -0,0 +1,56 @@
...
+
+        anchors {
+            left: parent.left
+            leftMargin: Theme.paddingLarge
+            verticalCenter: root.multiLine ? undefined : parent.verticalCenter
+        }
+        source: root.iconSource !== "" ? root.iconSource
+                                       : "image://theme/icon-l-document"
+
+        sourceSize { width: Theme.iconSizeSmall; height: Theme.iconSizeSmall }
```
rainemak: Change group to expand over multiple lines ...
```
sourceSize {
    width: Theme.iconSizeSmall
    height: Theme.iconSizeSmall
}
```

qml/CoverFileItem.qml
pvuorela: No, this is file name so it's more unlikely to have spaces to wrap at.

qml/CoverFileItem.qml
pvuorela: This is existing code. I can change that on a later PR, though.

qml/CoverFileItem.qml
rainemak: Right.

rainemak:
LGTM

PR #60, opened 2015-11-16
[sailfish-office] Avoid unnecessary import of Calligra. Fixes JB#33412

pvuorela:
Compiling all subpages at startup and needlessly importing Calligra
affects memory consumption and startup time. Now doing only what's
needed.


pvuorela:
@rainemak & @dcaliste

rainemak:
LGTM

PR #61, opened 2015-11-18
refactor and avoid flashing main page on startup

pvuorela:


pvuorela:
@rainemak & @dcaliste 

Found a crash too: pdf open and clicking another on transfers. But that already exists, so not a regression.

pvuorela:
Forgot to add dbus service to .desktop file.

pvuorela:
There.

main.cpp
```diff
@@ -115,16 +98,29 @@ Q_DECL_EXPORT int main(int argc, char *argv[])
...
-    context->setContextProperty(QStringLiteral("applicationWindow"), view.data());
-
     //% "Documents"
     Q_UNUSED(QT_TRID_NOOP("sailfish-office-ap-name"))
 
+    bool preStart = false;
+    QString fileName;
+
+    for (int i = 1; i < argc; ++i) {
+        if (QString(argv[i]) == "-prestart") {
```
rainemak: Use ```QStringLiteral("-prestart")``` in the comparison (http://doc.qt.io/qt-5/qstring.html#QStringLiteral).

I'd prefer using app->arguments() QStringList but this does the job as well.

qml/Main.qml
```diff
@@ -26,7 +26,90 @@ ApplicationWindow
     id: window
 
     property Item documentItem
-    property QtObject fileListModel
+    property QtObject fileListModel: trackerProvider.model
+    property Page _mainPage
```
rainemak: Imo not needed to be a private property.

qml/Main.qml
```diff
@@ -26,7 +26,90 @@ ApplicationWindow
...
+
+        default:
+            console.log("Warning: Unrecognised file type for file " + fileInfo.fullPath)
+        }
+
+        if (handler != "") {
+            pageStack.push(handler,
+                           { title: fileInfo.fileName, path: fileInfo.fullPath, mimeType: fileInfo.mimeType },
+                           PageStackAction.Immediate)
+        }
```
rainemak: What happens if there is no handler? Maybe something should be shown for the user to indicate that this file type is unknown.

pvuorela: Could, but need then to decide UI, have new strings and whatnot. Can create a task.

qml/Main.qml
pvuorela: I think I rather keep for now as rest are used from outside.

pvuorela:
Added string literal

rainemak:
App startup with a file takes quite a bit of hit with this as the first frame rendered only after everything is ready. The cost is calculated in seconds, when using qml profiler.

Normally first delegates of the FileListPage are created ~2.5secs after start. With this PR we are seeing the first frame >4secs after start.

Would it be possible to prepare the content the first page asynchronously in case we're launching office with a document file?

qml/Main.qml
rainemak: Thanks.

pvuorela:
As commented offline, I think file open startup optimizations should be handled separately.

rainemak:
LGTM

PR #62, opened 2015-11-18
Add QML_DEBUGGING_ENABLED env variable to enabling qml debugging / profiling

rainemak:
@pvuorela smallie for easier qml profiler enabling.


pvuorela:
LGTM

PR #63, opened 2015-11-25
Fix crash + cosmetics

pvuorela:
See commits.


pvuorela:
@rainemak & @dcaliste

dcaliste:
Good catch for the layout bug. And happy to see you here. Good luck for the times to come.

rainemak:
Layouting / rendering works now as it should when starting from an external file. LGTM.

PR #64, opened 2015-12-18
[sailfish-office] Fix pdf document image grabbing. Contributes to JB#31599

rainemak:
This requires a small change on Sailfish Silica. Thus, left the ```updateCover``` but in practice that could be removed. If everybody agrees, I'd like to drop it already now.

Idea here is that content is grab when peeking starts (cover becomes visible) and released when cover is invisible and application window is visible (in practice at foreground). The releasing part guarantees that the old captured document doesn't flicker on the cover.

In addition, grabbing is done only if application window is visible (main window). Currently you can see errors printed from QQuickItemGrabResultPrivate::create we're trying to grab when main window is already hidden.

### DIFF
--- a/qml/CoverPage.qml
+++ b/qml/CoverPage.qml
@@ -22,9 +22,17 @@ import Sailfish.Silica 1.0
 CoverBackground {
     id: root
 
-    property bool update: status === Cover.Active || Qt.application.active
-
-    onUpdateChanged: previewImage.updatePreview()
+    // While peeking both main window and cover window are visible at the
+    // same time. Thus, we can capture from main window when cover
+    // becomes visible.
+    // Note: this may change in future.
+    onVisibleChanged: {
+        if (visible) {
+            previewImage.updatePreview()
+        } else if (window.visible) {
+            previewImage.source = ""
+        }
+    }
 
     CoverPlaceholder {
         //: Cover placeholder shown when there are no documents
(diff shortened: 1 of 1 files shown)


dcaliste:
Cannot really test now without the modification in Silica since the onVisibleChanged signal is not fé ggered currently on the phone. Besides, switching to this onVisibleChanged signal makes sense and the code is cleaner.

I personaly don't mind if the cover update is broken by removing the update property while waiting for the Silica update.

qml/CoverPage.qml
```diff
@@ -22,9 +22,24 @@ import Sailfish.Silica 1.0
...
 
-    onUpdateChanged: previewImage.updatePreview()
+    // While peeking both main window and cover window are visible at the
+    // same time. Thus, we can capture from main window when cover
+    // becomes visible.
+    onVisibleChanged: {
+        if (visible) {
+            previewImage.updatePreview()
+        } else if (window.visible) {
+            previewImage.source = ""
```
pvuorela: What's the intention for this? Otherwise previewImage content is updated by the updatePreview() function which gets called from multiple places, so this looks a bit inconsistent. Also the fallback icon + file name is set based on image visibility, which is not changed with this.

pvuorela:
Silica changes seem source compatible, so no objections for removing temporary code.

qml/CoverPage.qml
rainemak: When cover is invisible and ApplicationWindow visible we can clear the preview. Thus, you see only the next preview and not glimpses of the old preview.

rainemak: This also fixes grabbing of large pdf (my test pdf is >1.5MB) as the capture is taken when the cover is actually visible. You'll see on the cover what's rendered on the app window side. AFAIS the error is that we're trying to grab before the actual content of the document is rendered (documentItem changed signal is not enough).

rainemak:
Rebased and removed the temporary code.

jpetrell:
LGTM

PR #66, opened 2016-01-04
[office] Add a dialog to display URL link before opening browser.

dcaliste:
There is currently a lack in security, in my opinion, when clicking on a link in a PDF document. It opens immediately the browser, without the the user to check that the URL of the link correspond to something he wants to do. I've received recently several fishing attempts by email in attached PDF documents with false links to banks…

This PR is still not really user friendly, I think, but I submit it anyway to start a discussion on the better way to inform the user of URL in links and avoid fishing (but not only). It currently opens a dialog with the URL and open the brozser on dialog acceptance.

What's your opinion on this matter ? I can try to implement a better solution if you think to any or has mockup about this from design, whatever.

### DIFF
--- a/plugin/PDFContextMenu.qml
+++ b/plugin/PDFContextMenu.qml
@@ -1,0 +1,136 @@
+/****************************************************************************************
+  **
+  ** Copyright (C) 2013-2016 Jolla Ltd., Damien Caliste
+  ** Contact: Martin Jones <martin.jones@jollamobile.com>
+  ** Contact: Damien Caliste <dcaliste@free.fr>
+  ** All rights reserved.
+  ** 
+  ** You may use this file under the terms of BSD license as follows:
+  **
+  ** Redistribution and use in source and binary forms, with or without
+  ** modification, are permitted provided that the following conditions are met:
+  **     * Redistributions of source code must retain the above copyright
+  **       notice, this list of conditions and the following disclaimer.
+  **     * Redistributions in binary form must reproduce the above copyright
+  **       notice, this list of conditions and the following disclaimer in the
+  **       documentation and/or other materials provided with the distribution.
+  **     * Neither the name of the Jolla Ltd nor the
+  **       names of its contributors may be used to endorse or promote products
+  **       derived from this software without specific prior written permission.
+  ** 
+  ** THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+  ** ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+  ** WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+  ** DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE LIABLE FOR
+  ** ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+  ** (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+  ** LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+  ** ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+  ** (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+  ** SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+  **
+  ****************************************************************************************/
+
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+
+Rectangle {
+    id: root
+
+    property alias title: labelTitle.text
+    property alias message: labelMessage.text
+    default property alias children: contentColumn.data
+
+    readonly property bool landscape: width > height
+    property Item _highlightedItem
+
+    gradient: Gradient {
+        GradientStop { position: 0.0; color: Theme.highlightDimmerColor }
+        GradientStop { position: 1.0; color: Theme.rgba(Theme.highlightDimmerColor, .91) }
+    }
+    opacity: 0.
+    Behavior on opacity { FadeAnimation { duration: 300 } }
+    visible: opacity > 0.
+    anchors.fill: parent
+        
+    MouseArea {
+        anchors.fill: parent
+        preventStealing: true
+
+        onPressed: _highlightMenuItem(mouse.y - contentColumn.y)
+        onPositionChanged: _highlightMenuItem(mouse.y - contentColumn.y)
+        onCanceled: _highlightMenuItem(null)
+        onReleased: {
+            if (_highlightedItem !== null) {
+                _highlightedItem.down = false
+                _highlightedItem.clicked()
+            }
+            root.opacity = 0.
+        }
+    }
+
+    Column {
+        spacing: Theme.paddingMedium
+        anchors.top: parent.top
+        anchors.topMargin: Theme.paddingLarge * 2
+        anchors.horizontalCenter: parent.horizontalCenter
+        width: parent.width - Theme.paddingLarge * 2
+        InfoLabel {
+            id: labelTitle
+            elide: Text.ElideRight
+            wrapMode: Text.Wrap
+            maximumLineCount: 2
+            color: Theme.highlightColor
+            opacity: .6
+        }
+        InfoLabel {
+            id: labelMessage
+            font.pixelSize: Theme.fontSizeMedium
+            wrapMode: Text.Wrap
+            elide: Text.ElideRight
+            maximumLineCount: landscape ? 1 : 4
+            color: Theme.highlightColor
+            opacity: .6
+        }
+    }
+
+    Column {
+        id: contentColumn
+
+        anchors.bottom: parent.bottom
+        anchors.bottomMargin: landscape ? Theme.paddingLarge : Theme.itemSizeSmall
+        width: parent.width
+    }
+
+    function _highlightMenuItem(yPos) {
+        var xPos = width / 2
+        var child = contentColumn.childAt(xPos, yPos)
+        if (!child) {
+            _setHighlightedItem(null)
+            return
+        }
+        var parentItem
+        while (child) {
+            if (child && child.hasOwnProperty("__silica_menuitem") && child.enabled) {
+                _setHighlightedItem(child)
+                break
+            }
+            parentItem = child
+            yPos = parentItem.mapToItem(child, xPos, yPos).y
+            child = parentItem.childAt(xPos, yPos)
+        }
+    }
+
+    function _setHighlightedItem(item) {
+        if (item === _highlightedItem) {
+            return
+        }
+        if (_highlightedItem) {
+            _highlightedItem.down = false
+        }
+        _highlightedItem = item
+        if (_highlightedItem) {
+            _highlightedItem.down = true
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


pvuorela:
For UI my first idea could be something like the browser has for long press of links: an overlay showing link and options. On this version it seems to now miss qml on CMakeLists. Also while transitioning to new page, the pdf misses opacity transition. I think scenegraph item needs a bit adjustment for that.

But then again, is there really that big of a security issue. On browser it needs special action to show the url and on email I don't think it's even possible.

dcaliste:
Oups, sorry for the missing entry in the CMakeLists file, my bad…

Ok, for the overlay like in the browser, I will give it a try. Especially, it gives opportunity to display more options like "copy to the clipboard", which can be useful.

Currently in the email reader, clicking on a link directly opens the browser, but in my opinion, it's the same issue. I would prefer to see the url before opening it, but maybe it's only a geek habit ? Besides, it provides an immediate feedback that something happens (the overlay appears) when clicking on a link, while currently (even in the mail reader) the browser window appears after some short seconds, but without any feedback.

As an after-thought on this, I would like to have a display like :
 External link
 - host: www.example.com
 - url : http://www.example.com/my/page.html
       Open in the browser
       Copy to clipboard

This would display clearly the destination server (in the idea to limit fishing for common users). But the QUrl C++ machinery is not available from the QML side to extract easily the host from an url…

dcaliste:
I've modified the commit to use an overlay like in the browser (and correct the CMakeList mistake). What do you think ? I agree that the UI is much more consistent like that through out the OS apps, nice.

To be more consistent, I may suggest to open this overlay only on long press. Like that, single press still opens the link immediatly like now, while a long press allows to review the link target (or copy it to clipboard) before taking action. Would it be better ?

pvuorela:
Like it lot better now. Overlay is quite less abrupt change. Haven't yet been able to ask design opinion, but I think it should copy Browser UI even more. Now there is less padding on top and bottom, the gradient is different and the buttons work diffently. Also it's possible to press on top of a button and move finger to scroll pdf underneath.

For long press, it could be consistent but then again Browser opens links internally whereas on PDF it can be a surprising application switch. Some of the links are handled internally, so for consistency they should have an overlay too. Something for design too, but might not be bad option as it is.

plugin/PDFContextMenu.qml
```diff
@@ -0,0 +1,92 @@
...
+
+        anchors.bottom: parent.bottom
+        anchors.bottomMargin: Theme.paddingLarge
+        width: parent.width
+
+        BackgroundItem {
+            MenuItem {
+                anchors.horizontalCenter: parent.horizontalCenter
+                //% "Open link in browser"
+                text: qsTrId("sailfish-office-me-pdf-open-link")
```
pvuorela: Can be as well mailto: url or something like that?

plugin/PDFContextMenu.qml
```diff
@@ -0,0 +1,92 @@
...
+            }
+            onClicked: {
+                root.opacity = 0.
+                Qt.openUrlExternally(root.url)
+            }
+        }
+        BackgroundItem {
+            MenuItem {
```
pvuorela: As noted, MenuItems don't seem to really work here.

plugin/PDFContextMenu.qml
dcaliste: You're right, I guess so. I'll generate a PDF like that to test and adapt the UI accordingly.

dcaliste:
Ok, thank for the positive feedback. I didn't fully copy the browser overlay at once, because it seems to me very complicated for some labels and menu entries. I understand what it's doing and more or less why it's doing it like that, but I wanted to experiment something simpler. For the gradient, sorry, forgot to revert experiment also ! Indeed, during usage I've never notice the gradient on the context menu of the browser. I've only see it when reading the source, so I tried a bigger transparency to see how it renders, but forgot to revert.

Well, I'm going to import more of the UI layout of the browser context menu. Since I won't rewrite some parts of the menuitem handling, I guess, I'll have to use the browser licence for this specific file, am I doing it right like that ?

Finally, I agree with your point for the internal links also, especially since we don't have a « back » button, pressing accidentally an internal link can be currently annoying, so having the overlay to confirm the jump would be nice.

dcaliste:
Here is an updated implementation that follows the same layouting than the one for the browser (except if I missed something). I've copied the highlighting source code for the MenuItems from the ```ContextMenu.qml``` code in the Silica source tree, thus I've changed the licence header for ```PDFContextMenu.qml``` to reflect this.

I've created a context menu also for the internal goto links in ```PDFView.qml```, and used a variation in the label when the external link starts with « mailto: ».

I quite like this context menu choice as it is now. Especially, it can be reused later for a selected text context menu for PDF annotations… I've in a branch named « selection » in my github some code to select portions of text in a PDF file. Opening this context menu with the text as message and several menu items like « copy to clipboard », « annotate », « highlight »… seems a good UI, better IMHO than a floating toolbar with tiny icons.

pvuorela:
Talked with design. Overlay UI approved, though needed only for external links, felt a bit too much for internal gotos.

plugin/PDFView.qml
```diff
@@ -190,7 +198,41 @@ SilicaFlickable {
...
         HorizontalScrollDecorator { color: Theme.highlightDimmerColor },
-        VerticalScrollDecorator { color: Theme.highlightDimmerColor }
+        VerticalScrollDecorator { color: Theme.highlightDimmerColor },
+        PDFContextMenu {
+            id: contextMenuLinks
+            //% "External link"
+            title: qsTrId("sailfish-office-tl-pdf-link")
+            MenuItem {
+                text: contextMenuLinks.message.indexOf("mailto:") === 0 ?
+                //% "Open mail composer"
```
pvuorela: Problem now is that email application might not be installed. And all the other urls are now considered Browser's responsibility, e.g. tel:12345.

How if we have special case for web urls http/https with the current text, and all the others would be "Open externally", being then handled with whatever is used. Checking current handler for mime type and mapping them to application names can be done later if really really needed.

plugin/PDFView.qml
```diff
@@ -190,7 +198,41 @@ SilicaFlickable {
 
     children: [
         HorizontalScrollDecorator { color: Theme.highlightDimmerColor },
-        VerticalScrollDecorator { color: Theme.highlightDimmerColor }
+        VerticalScrollDecorator { color: Theme.highlightDimmerColor },
+        PDFContextMenu {
```
pvuorela: Parent PDFView is not necessarily fullscreen. Scroll page down a bit to show toolbar, click on link -> overlay is anchored to top of toolbar. 

Rotating the device while showing overlay flashes the document briefly. Though might be the same issue as application main page not showing while backstepping, likely resulting from scenegraph setup.

plugin/PDFView.qml
```diff
@@ -144,6 +144,7 @@ SilicaFlickable {
 
         PinchArea {
             anchors.fill: parent
+            enabled: contextMenuLinks.opacity == 0. && contextMenuGoto.opacity == 0.
```
pvuorela: Would be cleaner if stacking order handled this. Maybe fixed if context menu is on PDFDocumentPage.

plugin/PDFContextMenu.qml
```diff
@@ -0,0 +1,135 @@
...
+            root.opacity = 0.
+        }
+    }
+
+    Column {
+        spacing: Theme.paddingMedium
+        anchors.top: parent.top
+        anchors.topMargin: Theme.paddingLarge * 2
+        width: parent.width - Theme.paddingLarge * 2
+        InfoLabel {
```
pvuorela: These are not properly centered. Result a bit on left side.

plugin/PDFContextMenu.qml
```diff
@@ -0,0 +1,135 @@
+/****************************************************************************************
+  **
+  ** Copyright (C) 2013-2016 Jolla Ltd., Damien Caliste
+  ** Contact: Martin Jones <martin.jones@jollamobile.com>
+  ** Contact: Damien Caliste <dcaliste@free.fr>
+  ** All rights reserved.
+  ** 
+  ** You may use this file under the terms of BSD license as follows:
```
pvuorela: Unfortunately relicensing seems problematic. Would Browser have only Jolla modifications on corresponding component, we could use whatever license, but there seems at least one commit coming outside. Cleanest option would be to rewrite the menu from scratch.

plugin/PDFView.qml
dcaliste: You're right, I'll change it for this (specific case for http(s), and generic for others).

plugin/PDFView.qml
dcaliste: You're right, I've notice the PDFView (not being fullscreen) issue. But I've no clue how to deal with this. PDFDocumentPage is the element being fullscreen, but PDFContextMenu should be logically linked with PDFView (at least because callback may be methods of PDFView like ```jumpTo()```).

Do you think it's possible to access the ApplicationWindow (or similar), so the Rectangle of the PDFContextMenu can be set at x:0 and y:0 and width: window.width… ?

plugin/PDFContextMenu.qml
dcaliste: Exact, forgot an horizontal shift by ```paddingLarge```…

plugin/PDFView.qml
dcaliste: Ok, I'll try to move PDFContextMenu to PDFDocumentPage and see which signals should be exchanged for the menu to interact with the PDFView content (scroll to position…).

plugin/PDFContextMenu.qml
dcaliste: Well, in fact, I copied the code from Silica/ContextMenu.qml (and keep the license from that file) not from the browser, so there may not be any relicensing issue because Jolla is only copyright owner ? Right ? You can see it because function names are those from the Silica ContextMenu item and not exactly the one of the BrowserContextMenu file.

Beside, I can rewrite the two copied functions, but it seems to me that it was more in the spirit of the license to copy the code and keep the license than to rewrite the same idea with different variable names and not mention the original copyright owner of the implementation.

plugin/PDFView.qml
pvuorela: Well, it should be possible to set different parent property value for the menu, but still I think it could be cleaner having the element where it visually should be. View already provides gotoPage() and similar functionality which is used from document page. Only additional requirement would be passing click signals from view.

dcaliste:
Ok, thanks for the (very quick) feedback and internal work at Jolla about this. I'll do the changes as discussed in the comments (and remove the overlay for the internal gotos) and resubmit when ready.

plugin/PDFContextMenu.qml
pvuorela: Yes, Silica is ok for source and even has the same license already. And you had even mentioned that earlier, sorry forgot that part.

dcaliste:
I've corrected the following:
* centred the content of the context menu;
* remove the context menu for the internal goto links;
* used a "open in browser" action only for ```http:``` and ```https:``` links, and a generic "open in external app" for others;
* moved the PDFContextMenu item to PDFDocumentPage.qml, which makes unnecessary to disable the pinch area in PDFView anymore and also renders now the context menu rectangle on top of the toolbar.

The last issue that remains is the flashing of the PDFView content when rotating the device while the context menu is open. I have no clue where does this come from, but I didn't really investigate neither…

Thanks for all suggestions, as for previous PR, this one looks much better now than when initially submitted.

plugin/PDFDocumentPage.qml
```diff
@@ -206,6 +211,26 @@ DocumentPage {
...
+        id: contextMenuLinks
+        //% "External link"
+        title: qsTrId("sailfish-office-tl-pdf-link")
+        MenuItem {
+            text: (contextMenuLinks.message.indexOf("http:") === 0 ||
+            contextMenuLinks.message.indexOf("https:") === 0) ?
+            //% "Open in browser"
+            qsTrId("sailfish-office-me-pdf-open-link") :
+            //% "Open in external application"
+            qsTrId("sailfish-office-me-pdf-open-external")
```
pvuorela: Please have some indentation on these. Starting all lines at same position makes it harder to notice where property value starts. Operators preferably at start of lines: https://wiki.qt.io/Qt_Coding_Style#Line_breaks

dcaliste: My bad, sorry for having to mention this again. Corrected.

pvuorela:
Looking good! Orientation is minor enough detail so it can be checked later. Thanks!

PR #67, opened 2016-01-18
[office] modify link selection on tap for a better sensivity.

dcaliste:
As discussed when dealing with PR #43, link can be quite difficult to tap on, especially for internal goto links that can be used for references with very tiny link box… We agreed then that it could be a good idea to try to find links on closest distance instead of the current implementation of augmented link boxes.

This PR introduces a square distance calculation that equals to zero inside a link box, and equals to the distance to the closest point on the link box border when outside. Then the link is chosen as the one with the closest value on all links within a wiggle room.

To be dpi independant, the wiggle room (which is in pixel in C++) is chosen in QML as the size of an itemSizeMedium / 2, making links easily reachable (they are more or less the size of a thumb with this choice).

### DIFF
--- a/pdf/pdfcanvas.cpp
+++ b/pdf/pdfcanvas.cpp
@@ -59,6 +59,7 @@ public:
         , flickable(0)
         , resizeTimer(nullptr)
         , spacing(10.f)
+        , linkWiggle(4.f)
     { }
 
     PDFCanvas *q;
@@ -76,6 +77,7 @@ public:
     QTimer *resizeTimer;
 
     float spacing;
+    float linkWiggle;
 
     QRectF visibleArea;
 
@@ -93,12 +95,8 @@ public:
             delete texture;
         texturesToClean.clear();
     }
-
-    static const float wiggleFactor;
 };
 
-
-const float PDFCanvas::Private::wiggleFactor(4.f);
 
 PDFCanvas::PDFCanvas(QQuickItem *parent)
     : QQuickItem(parent), d(new Private(this))
@@ -197,6 +195,19 @@ void PDFCanvas::setSpacing(float newValue)
     }
 }
 
+float PDFCanvas::linkWiggle() const
+{
+    return d->linkWiggle;
+}
+
+void PDFCanvas::setLinkWiggle(float newValue)
+{
+    if (newValue != d->linkWiggle) {
+        d->linkWiggle = newValue;
+        emit linkWiggleChanged();
+    }
+}
+
 QColor PDFCanvas::linkColor() const
 {
     return d->linkColor;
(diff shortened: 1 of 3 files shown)


pdf/pdfcanvas.cpp
```diff
@@ -267,22 +278,48 @@ void PDFCanvas::layout()
...
+        dist += (rect.x() + rect.width() - (qreal)point.x())
+                * (rect.x() + rect.width() - (qreal)point.x());
+    }
+    if ((qreal)point.y() < rect.y()) {
+        dist += (rect.y() - (qreal)point.y()) * (rect.y() - (qreal)point.y());
+    } else if ((qreal)point.y() > rect.y() + rect.height()) {
+        dist += (rect.y() + rect.height() - (qreal)point.y())
+                * (rect.y() + rect.height() - (qreal)point.y());
+    }
+
```
pvuorela: I think manhattan distance would be both simpler as code and as specification for function.

pdf/pdfcanvas.cpp
```diff
@@ -267,22 +278,48 @@ void PDFCanvas::layout()
     update();
 }
 
+static qreal dist2FromRect(const QRectF &rect, const QPoint &point)
+{
+    qreal dist = 0.;
+
+    if ((qreal)point.x() < rect.x()) {
+        dist += (rect.x() - (qreal)point.x()) * (rect.x() - (qreal)point.x());
+    } else if ((qreal)point.x() > rect.x() + rect.width()) {
```
pvuorela: Maybe use left(), right(), top(), bottom() instead of calculating such values yourself?

pdf/pdfcanvas.cpp
```diff
@@ -267,22 +278,48 @@ void PDFCanvas::layout()
...
+    return dist;
+}
+
 QUrl PDFCanvas::urlAtPoint(const QPoint &point)
 {
     for (int i = 0; i < d->pageCount; ++i) {
         const PDFPage &page = d->pages.value(i);
         if (page.rect.contains(point)) {
+            qreal dist2Min = d->linkWiggle * d->linkWiggle;
+            QUrl url = QUrl();
```
pvuorela: Unnecessary assignment.

pdf/pdfcanvas.cpp
dcaliste: Right, didn't notice existence of these functions…

pdf/pdfcanvas.cpp
dcaliste: No, as far as I understand, one cannot use the Manhattan distance here. I mean, computing the Manhattan distance between the tap position and the link centre won't give good results. 

Indeed, the iso-lines for this distance are point-centred diamond-shapes (*i.e.* rotated square of 45°, for the usual distance, the iso-lines are point-centred spheres). Links can be (are often in fact) elongated boxes in x direction. Diamond-shapes are not matching these elongated shapes (very bad ascii art):
```
Manhattan
     /\        
    / . \   (should be square-shape)
    \  /
     \/
dist2FromBox
  /----------------------------\ (should be rounded corners in ASCII art also)
  |  link with very long text  |
  \----------------------------/
```
With the given function, the distance iso-lines correspond to something like a margin around each link box.

dcaliste:
I've modified the commit to use ```right()```, ```left()```… which makes the distance function more readable. I've also removed the unnecessary assignation on creation of variable ```url```.

What do you think ?

pvuorela:
Earlier manhattan distance comment on an outdated diff, so continuing here.

Diamond shape link area would only result from links of single point, which dont' really exist. For a text rectangle that would be similar to what you drew except no rounded corners. Then again doesn't matter that much, both should be fine. If this way, I'd propose to name the function distanceSquared or similar. 2 without superscript is not that evident.

pdf/pdfcanvas.h
```diff
@@ -56,6 +57,15 @@ class PDFCanvas : public QQuickItem
     void setSpacing(float newValue);
 
     /**
+     * Getter for property #linkWiggle.
+     */
+    float linkWiggle() const;
+    /**
+     * Getter for property #linkWiggle.
+     */
```
pvuorela: Similar already here elsewhere, so ok to have, but can also omit comments as noise.

dcaliste:
Ok, I understand your point about the Manhattan distance. I thought you were implying to use QPoint::manhattanLength(), thus my answer to your comment. If I redefine the Manhattan distance to work on a QRect and not on a QPoint, I think I'll end up with more or less the same implementation, except having some absolute value function instead of products.

Well, I've changed the function and variable names to be squaredFunctionSomething() instead of dist2Something(). I've also removed the comments in the header to avoid adding noise.

Thanks for the review.

pvuorela:
Looking good to me and works nicely. Thanks!

PR #68, opened 2016-01-21
[office] Use new PasswordField item to unlock PDFs.

dcaliste:
Minor changes to be in coherence with the rest of password entries in Sailfish.

I've raised the silica version requirement, but without knowing when PasswordField was exactly introduced, I put it to the version currently in SF2.0.1.

I'm wondering if these kinds of minor PR can save your time by just looking at it and accepting (if allright of course), or if you prefer to track these small changes internally ?


pvuorela:
Looking good to me. Thanks!

PR #69, opened 2016-01-27
[office] Correct an issue with large links not working.

dcaliste:
Quite embarassing, I've mixed bottom() and top() for QRectF in last commit for links.

In fact, I've looked at the Qt documentation (5.5 version on the web) and it mentions that top() is equivalent to y(), thus I wrote the previous commit according to this. But I notice an issue (links not working) when the link height is bigger than the wiggle. Debugging on the phone, I noticed that top() is in fact y() + height(). So I don't know if it's a documentation issue on the web or if top() and bottom() were exchanged between Qt5.2 and Qt5.5.

What your opinion ?

[Example of PDF with large links not working currently](https://github.com/sailfishos/sailfish-office/files/106478/CALISTE_DAMIEN-YA6FPQ.pdf)

### DIFF
--- a/pdf/pdfrenderthread.cpp
+++ b/pdf/pdfrenderthread.cpp
@@ -90,10 +90,15 @@ public:
             Poppler::Page *page = document->page(i);
             QList<Poppler::Link*> links = page->links();
             for (Poppler::Link* link : links) {
+                // link->linkArea() may return negative heights,
+                // as mentioned in Freedesktop bug:
+                // https://bugs.freedesktop.org/show_bug.cgi?id=93900
+                // To avoid later unexpected asumption on height,
+                // link->linkArea() is normalized.
                 switch (link->linkType()) {
                 case (Poppler::Link::Browse): {
                     Poppler::LinkBrowse *realLink = static_cast<Poppler::LinkBrowse*>(link);
-                    QRectF linkArea = link->linkArea();
+                    QRectF linkArea = link->linkArea().normalized();
                     linkTargets.insert(i, QPair<QRectF, QUrl>(linkArea, realLink->url()));
                     break;
                 }
@@ -102,7 +107,7 @@ public:
                     // Not handling goto link to external file currently.
                     if (gotoLink->isExternal())
                         break;
-                    QRectF linkArea = link->linkArea();
+                    QRectF linkArea = link->linkArea().normalized();
                     QUrl linkURL = QUrl("");
                     QUrlQuery query = QUrlQuery();
                     query.addQueryItem("page", QString::number(gotoLink->destination().pageNumber()));


pvuorela:
Actually the code is fine and so should be Qt. Adding some debug output, I see that PDFRenderThreadPrivate::rescanDocumentLinks() creates link areas like QRectF(0.137753,0.3409 0.109194x-0.0130658), height is negative(!). Created https://bugs.freedesktop.org/show_bug.cgi?id=93900

dcaliste:
You're right, I may have guess the negative height issue. Especially that I read again the documentation yesterday and read the notice about negative height and width.

I propose, in the meantime waiting for Poppler to clarify the situation, to call ```normalized()``` on the given rectangles from Poppler. This solve the issue and keep the code computing the distance correct.

Is it better ?

pvuorela:
Normalized should be a simple enough solution, and not result bad behavior whatever Poppler returns. Could, though, wait a moment to see how it goes with the bug report.

Did wonder when reviewing earlier pull request why the wiggle distance of medium/2 is not bigger, but guess this can explain. Didn't test, but some of the paddings might be better afterwards?

dcaliste:
The problem with padding is for links that are too close together. Padding may expand links over other links and as the algorithm is choosing the first, the other links will not be clickable.

Example: "some text that deals with something, see ref [1], [2] and [3]" with links on the numbers, it won't be possible to tap on [2] or [3] because the padding of [1] will expand  over.

Moreover, I like the fact that wiggle can be chosen from QML so one can put a dpi independant value.

pvuorela:
By padding I meant replacing wiggle Theme.itemSizeMedium/2 with Theme.paddingSmall/Medium/Large. Currently because of negative heights, the wiggle is really Medium/2-link_height, right?

dcaliste:
Ah, ok sorry didn't understand your remark first.

With the negative height, currently, what happen is that the distance (that is compared after to the wiggle) when tapping outside the link is distance to the top part when tapped under and the distance to the bottom when tapped over the link. So for a thin link, top and bottom are closed together so the calculated distance (even with the negative height) results in more or less the distance to the link, so everything seems to work. But for fat links when tapping inside, the distance is more or less always equals to the size of the link, so if link is bigger than wiggle, tapping has no effect.

I agree that for small links, currently, we're comparing wiggle (Theme.itemSizeMedium / 2) with (distance + height), so wiggle in that cases is indeed as you said Medium/2-link_height.

I choose first wiggle to be Theme.itemSizeMedium/2, for links to be at least the size of an itemSizeMedium (half on top plus half on the bottom plus link height), which means that a thumb can always tap on them without much precision since it is the size of a tapping Item in ListViews. Now, we can use Theme.padding values instead for wiggle if it's more meaningfull.

pvuorela:
No immediate Poppler activity, so I think we could merge this. By testing the wiggle doesn't seem too large, so can just continue using what's there currently. Want to still add reference to freedesktop bug where normalized() is used, so it can be removed when not needed anymore?

Btw, testing your example pdf made me think whether links should be highlighted or not. The image there has quite a bit extra highlighting. Adobe Acrobat for Android doesn't seem to add extra elements.

PR #70, opened 2016-02-01
Implement text selection in PDF documents

dcaliste:
This PR introduces text selection in PDF documents. It is made of four commits:
* first commit makes available in PDFDocument class the list of text boxes for each pages as returned by Poppler.
* second commit implements a QAbstractModelList to expose a portion of the previous list of text boxes for each page. It introduces a method to retrieve a text box at a given point and methods to change the start and stop boundaries for text selection.
* third commits adds a longPress signal to the PDFLinkArea object.
* fourth commit uses all the previous ones to implement the user interface for text selection.

The way text is selected is as followed:
* long press on a word, the word is highlighted and two handles appears to both ends;
* tap and drag a handle (start or stop) to adjust the selection; when a handle is tapped, the document is zoomed and a bit shifted to allow to see where the handle is moved.
* changing selection automatically copy the text to clipboard.
* unselect by long press on an area without text.

It took me a lot of time to create this PR (first commit was done on October 2015), particularly for the UI part (dragging, not interfering with flick…). But this is essential for a more ambicious PR implementing PDF annotations, so one can select a portion of text to annotate (comment or highlight). Having the annotation working (using Poppler of course) would be a great addition.

But first, let's discuss the selection PR, trying to improve this first (not really) proposition.

### DIFF
--- a/pdf/pdfcanvas.cpp
+++ b/pdf/pdfcanvas.cpp
@@ -220,6 +220,7 @@ void PDFCanvas::setLinkColor(const QColor &color)
         emit linkColorChanged();
     }
 }
+
 QColor PDFCanvas::pagePlaceholderColor() const
 {
     return d->pagePlaceholderColor;
@@ -276,9 +277,17 @@ void PDFCanvas::layout()
     update();
 }
 
-static qreal squaredDistanceFromRect(const QRectF &rect, const QPoint &point)
+qreal PDFCanvas::squaredDistanceFromRect(const QRectF &pageRect,
+                                         const QRectF &reducedCoordRect,
+                                         const QPointF &point)
 {
     qreal dist = 0.;
+    QRectF rect {
+        reducedCoordRect.x() * pageRect.width(),
+        reducedCoordRect.y() * pageRect.height() + pageRect.y(),
+        reducedCoordRect.width() * pageRect.width(),
+        reducedCoordRect.height() * pageRect.height()
+    };
 
     if ((qreal)point.x() < rect.left()) {
         dist += (rect.left() - (qreal)point.x()) * (rect.left() - (qreal)point.x());
@@ -294,7 +303,7 @@ static qreal squaredDistanceFromRect(const QRectF &rect, const QPoint &point)
     return dist;
 }
 
-QUrl PDFCanvas::urlAtPoint(const QPoint &point)
+QUrl PDFCanvas::urlAtPoint(const QPointF &point)
 {
     for (int i = 0; i < d->pageCount; ++i) {
         const PDFPage &page = d->pages.value(i);
@@ -302,13 +311,8 @@ QUrl PDFCanvas::urlAtPoint(const QPoint &point)
             qreal squaredDistanceMin = d->linkWiggle * d->linkWiggle;
             QUrl url;
             for (const QPair<QRectF, QUrl> &link : page.links) {
-                QRectF hitTarget{
-                    link.first.x() * page.rect.width(),
-                    link.first.y() * page.rect.height() + page.rect.y(),
-                    link.first.width() * page.rect.width(),
-                    link.first.height() * page.rect.height()
-                };
-                qreal squaredDistance = squaredDistanceFromRect(hitTarget, point);
+                qreal squaredDistance =
+                    squaredDistanceFromRect(page.rect, link.first, point);
                 
                 if (squaredDistance < squaredDistanceMin) {
                     url = link.second;
@@ -322,7 +326,7 @@ QUrl PDFCanvas::urlAtPoint(const QPoint &point)
     return QUrl();
 }
 
-QRectF PDFCanvas::fromPageToItem(int index, const QRectF &rect)
+QRectF PDFCanvas::fromPageToItem(int index, const QRectF &rect) const
 {
     if (index < 0 || index >= d->pageCount)
         return QRectF();
@@ -332,6 +336,16 @@ QRectF PDFCanvas::fromPageToItem(int index, const QRectF &rect)
                   rect.y() * page.rect.height() + page.rect.y(),
                   rect.width() * page.rect.width(),
                   rect.height() * page.rect.height());
+}
+
+QPointF PDFCanvas::fromPageToItem(int index, const QPointF &point) const
+{
+    if (index < 0 || index >= d->pageCount)
+        return QPointF();
+
+    const PDFPage &page = d->pages.value(index);
+    return QPointF(point.x() * page.rect.width() + page.rect.x(),
+                   point.y() * page.rect.height() + page.rect.y());
 }
 
 void PDFCanvas::pageFinished(int id, QSGTexture *texture)
@@ -542,4 +556,15 @@ void PDFCanvas::pageSizesFinished(const QList<QSizeF> &sizes)
 {
     d->pageSizes = sizes;
     layout();
+}
+
+QPair<int, QRectF> PDFCanvas::pageAtPoint(const QPointF &point) const
+{
+    for (int i = 0; i < d->pageCount; ++i) {
+        const PDFPage& page = d->pages.value(i);
+        if (page.rect.contains(point)) {
+            return QPair<int, QRectF>{i, page.rect};
+        }
+    }
+    return QPair<int, QRectF>{-1, QRectF()};
 }
--- a/pdf/pdfdocument.cpp
+++ b/pdf/pdfdocument.cpp
@@ -108,6 +108,11 @@ PDFDocument::LinkMap PDFDocument::linkTargets() const
     return d->thread->linkTargets();
 }
 
+PDFDocument::TextList PDFDocument::textBoxesAtPage(int page)
+{
+    return d->thread->textBoxesAtPage(page);
+}
+
 void PDFDocument::classBegin()
 {
 }
--- a/pdf/pdflinkarea.cpp
+++ b/pdf/pdflinkarea.cpp
@@ -19,21 +19,22 @@
 #include "pdflinkarea.h"
 #include "pdfcanvas.h"
 #include <QUrlQuery>
+#include <QTimer>
 
 class PDFLinkArea::Private
 {
 public:
     Private()
         : canvas(nullptr)
-        , clickInProgress(false)
         , wiggleFactor(4)
     { }
 
     PDFCanvas *canvas;
 
-    bool clickInProgress;
     QPointF clickLocation;
     int wiggleFactor;
+
+    QTimer pressTimer;
 };
 
 PDFLinkArea::PDFLinkArea(QQuickItem *parent)
@@ -41,6 +42,10 @@ PDFLinkArea::PDFLinkArea(QQuickItem *parent)
     , d(new Private)
 {
     setAcceptedMouseButtons(Qt::LeftButton | Qt::RightButton | Qt::MidButton);
+
+    d->pressTimer.setInterval(750);
+    d->pressTimer.setSingleShot(true);
+    connect(&d->pressTimer, &QTimer::timeout, this, &PDFLinkArea::pressTimeout);
 }
 
 PDFLinkArea::~PDFLinkArea()
@@ -64,15 +69,31 @@ void PDFLinkArea::setCanvas(PDFCanvas *newCanvas)
 
 void PDFLinkArea::mousePressEvent(QMouseEvent *event)
 {
-    d->clickInProgress = true;
     d->clickLocation = event->pos();
+    d->pressTimer.start();
+}
+
+void PDFLinkArea::mouseMoveEvent(QMouseEvent *event)
+{
+    // Don't activate anything if the finger has moved too far
+    QRect rect((d->clickLocation - QPointF(d->wiggleFactor, d->wiggleFactor)).toPoint(),
+               QSize(d->wiggleFactor * 2, d->wiggleFactor * 2));
+    if (!rect.contains(event->pos())) {
+        d->pressTimer.stop();
+        return;
+    }
 }
 
 void PDFLinkArea::mouseReleaseEvent(QMouseEvent *event)
 {
-    d->clickInProgress = false;
+    // Don't activate click if the longPress already fired.
+    if (!d->pressTimer.isActive())
+        return;
+    d->pressTimer.stop();
+
     // Don't activate anything if the finger has moved too far
-    QRect rect((d->clickLocation - QPointF(d->wiggleFactor, d->wiggleFactor)).toPoint(), QSize(d->wiggleFactor * 2, d->wiggleFactor * 2));
+    QRect rect((d->clickLocation - QPointF(d->wiggleFactor, d->wiggleFactor)).toPoint(),
+               QSize(d->wiggleFactor * 2, d->wiggleFactor * 2));
     if (!rect.contains(event->pos())) {
         return;
     }
@@ -98,10 +119,19 @@ void PDFLinkArea::mouseReleaseEvent(QMouseEvent *event)
     } else {
         emit linkClicked(url);
     }
-    event->accept();
 }
 
 void PDFLinkArea::mouseDoubleClickEvent(QMouseEvent* event)
 {
     emit doubleClicked();
+}
+
+void PDFLinkArea::mouseUngrabEvent()
+{
+    d->pressTimer.stop();
+}
+
+void PDFLinkArea::pressTimeout()
+{
+    emit longPress(d->clickLocation);
 }
--- a/pdf/pdfrenderthread.cpp
+++ b/pdf/pdfrenderthread.cpp
@@ -26,8 +26,6 @@
 #include <QCoreApplication>
 #include <QUrlQuery>
 
-#include <poppler-qt5.h>
-
 #include "pdfjob.h"
 #include "pdftocmodel.h"
 
@@ -71,6 +69,18 @@ class PDFRenderThreadPrivate
 {
 public:
     PDFRenderThreadPrivate() : document(nullptr), tocModel(nullptr) { }
+    ~PDFRenderThreadPrivate()
+    {
+        for (QMap<int, QList<QPair<QRectF, Poppler::TextBox*> > >::iterator i =
+                 textBoxes.begin();
+             i != textBoxes.end(); i++) {
+            for (QList<QPair<QRectF, Poppler::TextBox*> >::iterator j = i.value().begin();
+                 j != i.value().end(); j++) {
+                delete j->second;
+            }
+        }
+    }
+
 
     PDFRenderThread *q;
 
@@ -81,6 +91,7 @@ public:
     PDFTocModel *tocModel;
 
     QMultiMap<int, QPair<QRectF, QUrl> > linkTargets;
+    QMap<int, QList<QPair <QRectF, Poppler::TextBox*> > > textBoxes;
 
     void rescanDocumentLinks()
     {
@@ -131,7 +142,27 @@ public:
             delete page;
         }
     }
-
+    void retrieveTextBoxes(int i)
+    {
+        if (i < 0 || i >= document->numPages()) {
+            return;
+        }
+        Poppler::Page* page = document->page(i);
+        QList<Poppler::TextBox*> plst = page->textList();
+        QList<QPair<QRectF, Poppler::TextBox*> > lst;
+        for (QList<Poppler::TextBox*>::const_iterator box = plst.begin();
+             box != plst.end(); box++) {
+            QSizeF psize = page->pageSize();
+            QRectF bbox = (*box)->boundingBox();
+            QRectF rect{ bbox.x() / psize.width(),
+                    bbox.y() / psize.height(),
+                    bbox.width() / psize.width(),
+                    bbox.height() / psize.height()};
+            lst.append(QPair<QRectF, Poppler::TextBox*>{rect, *box});
+        }
+        textBoxes.insert(i, lst);
+        delete page;
+    }
 };
 
 class PDFRenderThreadQueue : public QObject, public QQueue<PDFJob*>
@@ -212,6 +243,14 @@ QMultiMap<int, QPair<QRectF, QUrl> > PDFRenderThread::linkTargets() const
 {
     QMutexLocker(&d->thread->mutex);
     return d->linkTargets;
+}
+
+QList<QPair<QRectF, Poppler::TextBox*> > PDFRenderThread::textBoxesAtPage(int page)
+{
+    QMutexLocker(&d->thread->mutex);
+    if (!d->textBoxes.contains(page))
+        d->retrieveTextBoxes(page);
+    return d->textBoxes[page];
 }
 
 void PDFRenderThread::queueJob(PDFJob *job)
--- a/pdf/pdfselection.cpp
+++ b/pdf/pdfselection.cpp
@@ -1,0 +1,566 @@
+/*
+ * Copyright (C) 2015-2016 Caliste Damien.
+ * Contact: Damien Caliste <dcaliste@free.fr>
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; version 2 only.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+#include "pdfselection.h"
+#include "pdfdocument.h"
+
+class PDFSelection::Private
+{
+public:
+    Private()
+        : canvas(nullptr)
+        , pageIndexStart(-1)
+        , boxIndexStart(-1)
+        , pageIndexStop(-1)
+        , boxIndexStop(-1)
+        , wiggle(4.)
+    {
+    }
+
+    PDFCanvas *canvas;
+
+    QPointF start;
+    QPointF stop;
+
+    int pageIndexStart;
+    int boxIndexStart;
+
+    int pageIndexStop;
+    int boxIndexStop;
+
+    bool handleReversed;
+
+    qreal wiggle;
+
+    enum Position {
+        At,
+        Before,
+        After
+    };
+    void textBoxAtIndex(int index, int *pageIndex, int *boxIndex);
+    void textBoxAtPoint(const QPointF &point, Position position, int *pageIndex, int *boxIndex);
+    int sliceCount(int pageIndex1, int boxIndex1, int pageIndex2, int boxIndex2) const;
+    int count() const;
+};
+
+PDFSelection::PDFSelection(QObject *parent)
+  : QAbstractListModel(parent), d(new Private)
+{
+}
+
+PDFSelection::~PDFSelection()
+{
+    delete d;
+}
+
+QHash<int, QByteArray> PDFSelection::roleNames() const
+{
+    QHash<int, QByteArray> roles;
+
+    roles.insert(Rect, "rect");
+    roles.insert(Text, "text");
+
+    return roles;
+}
+
+void PDFSelection::Private::textBoxAtIndex(int index, int *pageIndex, int *boxIndex)
+{
+    *pageIndex = -1;
+    *boxIndex = -1;
+    PDFDocument *doc = canvas->document();
+    if (doc != nullptr) {
+        int count_ = count();
+        while (index < 0)
+            index += count_;
+        while (index >= count_)
+            index -= count_;
+        *boxIndex = index + boxIndexStart;
+        for (*pageIndex = pageIndexStart;
+             *pageIndex <= pageIndexStop; *pageIndex += 1) {
+            int nBoxes = doc->textBoxesAtPage(*pageIndex).count();
+            if (*boxIndex < nBoxes)
+                return;
+            else
+              *boxIndex -= nBoxes;
+        }
+    }
+}
+
+QVariant PDFSelection::data(const QModelIndex& index, int role) const
+{
+    if (!d->canvas)
+        return QVariant();
+    PDFDocument *doc = d->canvas->document();
+    if (!doc)
+        return QVariant();
+
+    int pageIndex = -1;
+    int boxIndex;
+    QVariant result;
+
+    if (index.isValid())
+        d->textBoxAtIndex(index.row(), &pageIndex, &boxIndex);
+    if (pageIndex >= 0) {
+        const PDFDocument::TextList &boxes = doc->textBoxesAtPage(pageIndex);
+        const QPair<QRectF, Poppler::TextBox*> &box = boxes.at(boxIndex);
+        switch(role) {
+        case Rect:
+            result.setValue<QRectF>(d->canvas->fromPageToItem(pageIndex, box.first));
+            break;
+        case Text:
+            result.setValue<QString>(box.second->text());
+            break;
+        default:
+            result.setValue<QString>(QString("Unknown role: %1").arg(role));
+            break;
+        }
+    }
+    return result;
+}
+
+QPair<int, QRectF> PDFSelection::rectAt(int index) const
+{
+    if (!d->canvas)
+        return QPair<int, QRectF>();
+    PDFDocument *doc = d->canvas->document();
+    if (!doc)
+        return QPair<int, QRectF>();
+
+    int pageIndex, boxIndex;
+    d->textBoxAtIndex(index, &pageIndex, &boxIndex);
+    if (pageIndex < 0)
+        return QPair<int, QRectF>();
+
+    const PDFDocument::TextList &boxes = doc->textBoxesAtPage(pageIndex);
+    const QPair<QRectF, Poppler::TextBox*> &box = boxes.at(boxIndex);
+    return QPair<int, QRectF> {pageIndex, box.first};
+}
+
+int PDFSelection::rowCount(const QModelIndex& parent) const
+{
+    if (parent.isValid())
+        return 0;
+    return count();
+}
+
+int PDFSelection::Private::sliceCount(int pageIndex1, int boxIndex1,
+                                      int pageIndex2, int boxIndex2) const
+{
+    if (!canvas)
+        return 0;
+    PDFDocument *doc = canvas->document();
+    if (!doc)
+        return 0;
+
+    if (pageIndex1 < pageIndex2 ||
+        (pageIndex1 == pageIndex2 && boxIndex1 < boxIndex2)) {
+        int n = -boxIndex1;
+        for (int i = pageIndex1; i < pageIndex2; i++)
+            n += doc->textBoxesAtPage(i).count();
+        return n + boxIndex2;
+    } else {
+        int n = -boxIndex2;
+        for (int i = pageIndex2; i < pageIndex1; i++)
+            n += doc->textBoxesAtPage(i).count();
+        return -n - boxIndex1;
+    }
+}
+
+int PDFSelection::Private::count() const
+{
+    if (pageIndexStart < 0 || pageIndexStop < 0 ||
+        boxIndexStart < 0 || boxIndexStop < 0) {
+        return 0;
+    }
+
+    if (!canvas || !canvas->document())
+        return 0;
+
+    return sliceCount(pageIndexStart, boxIndexStart,
+                      pageIndexStop, boxIndexStop) + 1;
+}
+
+int PDFSelection::count() const
+{
+    return d->count();
+}
+
+PDFCanvas* PDFSelection::canvas() const
+{
+    return d->canvas;
+}
+
+void PDFSelection::setCanvas(PDFCanvas *newCanvas)
+{
+    if (newCanvas != d->canvas) {
+        if (d->canvas)
+            d->canvas->disconnect(this);
+        d->canvas = newCanvas;
+
+        connect(d->canvas, &PDFCanvas::pageLayoutChanged,
+                this, &PDFSelection::onLayoutChanged);
+
+        emit canvasChanged();
+
+        onLayoutChanged();
+    }
+}
+
+float PDFSelection::wiggle() const
+{
+    return d->wiggle;
+}
+
+void PDFSelection::setWiggle(float newValue)
+{
+    if (newValue != d->wiggle) {
+        d->wiggle = newValue;
+        emit wiggleChanged();
+    }
+}
+
+// TODO: update the Before and After cases to handle RTL languages.
+void PDFSelection::Private::textBoxAtPoint(const QPointF &point, Position position,
+                                           int *pageIndex, int *boxIndex)
+{
+    *pageIndex = -1;
+    *boxIndex  = -1;
+    if (!canvas || !canvas->document())
+        return;
+    
+    // point is given in canvas coordinates.
+    QPair<int, QRectF> at = canvas->pageAtPoint(point);
+    if (at.first < 0)
+        return;
+    *pageIndex = at.first;
+
+    const PDFDocument::TextList &boxes = canvas->document()->textBoxesAtPage(*pageIndex);
+    switch (position) {
+    case PDFSelection::Private::At: {
+        qreal squaredDistanceMin = wiggle * wiggle;
+        int i = 0;
+        for (PDFDocument::TextList::const_iterator box = boxes.begin();
+             box != boxes.end(); box++) {
+            qreal squaredDistance =
+                canvas->squaredDistanceFromRect(at.second, box->first, point);
+                
+            if (squaredDistance < squaredDistanceMin) {
+                *boxIndex = i;
+                squaredDistanceMin = squaredDistance;
+            }
+            i += 1;
+        }
+        break;
+    }
+    case PDFSelection::Private::Before: {
+        // Find the last box index in pageIndex that is before @point,
+        // including at @point. If none is found in this page, returns
+        // the last box of the previous page.
+        QPointF reducedCoordPoint {point.x() / at.second.width(),
+                (point.y() - at.second.y()) / at.second.height()};
+        for (PDFDocument::TextList::const_iterator box = boxes.begin();
+             box != boxes.end(); box++) {
+            /* Stop counting boxes up as soon as a box after reducedCoordPoint is found. */
+            if (box->first.y() > reducedCoordPoint.y() ||
+                (box->first.y() + box->first.height() > reducedCoordPoint.y() &&
+                 box->first.x() > reducedCoordPoint.x())) {
+                if (*boxIndex == -1) {
+                    *pageIndex -= 1;
+                    if (*pageIndex >= 0)
+                        *boxIndex = canvas->document()->textBoxesAtPage(*pageIndex).length() - 1;
+                }
+                return;
+            }
+            *boxIndex += 1;
+        }
+        *boxIndex = boxes.length() - 1;
+        break;
+    }
+    case PDFSelection::Private::After: {
+        // Find the first box index in pageIndex that is after @point,
+        // including at @point. If none is found in this page, returns
+        // the first box of the next page.
+        QPointF reducedCoordPoint {point.x() / at.second.width(),
+                (point.y() - at.second.y()) / at.second.height()};
+        *boxIndex = 0;
+        for (PDFDocument::TextList::const_iterator box = boxes.begin();
+             box != boxes.end(); box++) {
+            /* Stop counting boxes down as soon as a box after reducedCoordPoint is found. */
+            if (box->first.y() > reducedCoordPoint.y() ||
+                (box->first.y() + box->first.height() > reducedCoordPoint.y() &&
+                 box->first.x() + box->first.width() > reducedCoordPoint.x())) {
+                return;
+            }
+            *boxIndex += 1;
+        }
+        *pageIndex += 1;
+        if (*pageIndex >= canvas->document()->pageCount())
+            *pageIndex = -1;
+        *boxIndex = 0;
+        break;
+    }
+    default:
+        break;
+    }
+}
+
+void PDFSelection::setStart(const QPointF &point)
+{
+    int pageIndex, boxIndex;
+    bool swap;
+
+    d->textBoxAtPoint(point, PDFSelection::Private::After, &pageIndex, &boxIndex);
+    if (pageIndex < 0 || boxIndex < 0)
+      return;
+
+    swap = (pageIndex > d->pageIndexStop
+            || (pageIndex == d->pageIndexStop && boxIndex > d->boxIndexStop));
+    if (swap) {
+        pageIndex = d->pageIndexStop;
+        boxIndex = d->boxIndexStop;
+        d->handleReversed = !d->handleReversed;
+        setStop(point);
+        if (pageIndex == d->pageIndexStop && boxIndex == d->boxIndexStop)
+            emit handle1Changed();
+    }
+    if (pageIndex == d->pageIndexStart && boxIndex == d->boxIndexStart) {
+        if (swap)
+            emit handle2Changed();
+        return;
+    }
+
+    const PDFDocument::TextList &boxes = d->canvas->document()->textBoxesAtPage(pageIndex);
+    QRectF box = boxes[boxIndex].first;
+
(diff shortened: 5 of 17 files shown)


pvuorela:
Tested a bit and showed to design. Generally nice, but noted following:
- No need for freely moving selection circle.
- Cursor should move a bit offset upwards from touch position so it's not hidden under the finger. Now seems to depend on how/where cursor was grabbed.
- Exiting selection mode with long press was a bit hard to figure out. Should exit with click, split screen change not applied when doing so.
- Preferably allowing to move cursor between characters.
- Could perhaps omit document zoom when moving cursor. Now the text not being re-rendered doesn't look good and this doesn't either scroll the document when touch point is near screen edges.
- Doesn't allow moving selection end before selection start or vice versa.
- Nice addition could be same vibra feedback as text editors have when entering selection mode.

pdf/pdfcanvas.cpp
```diff
@@ -278,9 +278,16 @@ void PDFCanvas::layout()
     update();
 }
 
-static qreal squaredDistanceFromRect(const QRectF &rect, const QPoint &point)
+qreal PDFCanvas::squaredDistanceFromRect(const QRectF &pageRect, const QRectF &redRect,
```
pvuorela: What's "red"?

pdf/pdfcanvas.cpp
```diff
@@ -335,6 +337,15 @@ QRectF PDFCanvas::fromPageToItem(int index, const QRectF &rect)
                   rect.width() * page.rect.width(),
                   rect.height() * page.rect.height());
 }
+QPointF PDFCanvas::fromPageToItem(int index, const QPointF &pt) const
```
pvuorela: Unnecessary abbreviation.

pdf/pdfcanvas.cpp
```diff
@@ -545,3 +556,14 @@ void PDFCanvas::pageSizesFinished(const QList<QSizeF> &sizes)
     d->pageSizes = sizes;
     layout();
 }
+
+QPair<int, QRectF> PDFCanvas::pageAtPoint(const QPointF &point) const
+{
+    for (int i = 0; i < d->pageCount; ++i) {
+        const PDFPage& page = d->pages.value( i );
```
pvuorela: Space around parameters style was gotten rid of.

pdf/pdflinkarea.cpp
```diff
@@ -33,7 +34,10 @@ class PDFLinkArea::Private
 
     bool clickInProgress;
     QPointF clickLocation;
+    ulong clickTimestamp;
     int wiggleFactor;
+
+    QTimer *pressTimer;
```
pvuorela: Unnecessary dynamic allocation. Could be just a direct member variable.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,439 @@
...
+}
+
+QPointF PDFSelection::start() const
+{
+    if (!d->canvas || d->iPageStart < 0)
+        return QPointF();
+    
+    return d->canvas->fromPageToItem(d->iPageStart, d->start);
+}
+void PDFSelection::setStart(const QPointF &pt)
```
pvuorela: Space between methods.

pvuorela: Also unnecessary abbreviations here.

plugin/PDFView.qml
```diff
@@ -124,12 +125,36 @@ SilicaFlickable {
     // Ensure proper zooming level when device is rotated.
     onWidthChanged: adjust()
 
+    PDF.Selection {
+        id: selection
+        canvas: pdfCanvas
+        wiggle: Theme.itemSizeSmall / 2
+        onTextChanged: if (text.length > 0) Clipboard.text = text
```
pvuorela: Would be more efficient if clipboard is updated only when cursor grab gets released.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,439 @@
...
+{
+public:
+    Private(): canvas(nullptr) {
+        roles.insert( Rect, "rect" );
+        roles.insert( Text, "text" );
+
+        iPageStart = iPageStop = -1;
+        iBoxStart = iBoxStop = -1;
+
+        wiggle = 4.;
```
pvuorela: Could include all in initialization list. AFAIK roles is requested only once so can as well just create a new instance on roles(). What does the "i" stand for?

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,439 @@
...
+
+    const PDFDocument::TextList &boxes = d->canvas->document()->textBoxesAtPage(iPage);
+    QRectF box = boxes[iBox].first;
+    
+    beginResetModel();
+    d->iPageStart = -1;
+    d->iBoxStart = -1;
+    d->iPageStop = -1;
+    d->iBoxStop = -1;
+    endResetModel();
```
pvuorela: No need to do both reset and insert.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,439 @@
...
+    
+    const PDFDocument::TextList &boxes = doc->textBoxesAtPage(d->iPageStart);
+    out = boxes.value(d->iBoxStart).second;
+    int i, j;
+    for (i = d->iPageStart; i <= d->iPageStop; i++) {
+        const PDFDocument::TextList &boxes = doc->textBoxesAtPage(i);
+        for (j = ((i == d->iPageStart) ? d->iBoxStart + 1 : 0);
+             j < ((i == d->iPageStop) ? d->iBoxStop + 1 : boxes.length());
+             j++) {
+            out += " ";
```
pvuorela: TextBox doesn't necessarily have space after, e.g. my test pdf has LaTEX with three boxes. There seems to be hasSpaceAfter() provided in Poppler.

pvuorela:
Not sure what's happening but had some difficulty getting some text to clipboard. Selected ok, but just didn't get available.

dcaliste:
Thanks for the quick answer (as usual, thanks again) and deep review (you find out the corner cases I didn't speak about in the introduction) ;)

Let's answer:
* don't understand first point, sorry :/ Do you mean that it's not necessary to move the black circles around when wanting to change the selection ?
* It allready moves a bit upward (Theme.itemSizeSmall if I remember), but there is a bug that I was not able to correct. If you press one selection handle and keep pressing waiting for the zoom in to happen, you'll see that the view is indeed shifted upward. The trouble is that dragging while the view is scaled up confuse the dragging. You can notice it by putting a very long zoom in transition (3 seconds), tap the handle and immediately drag a bit, then stop dragging while the zoom transition is still happening. You'll notice that the handle is displaced away down right from your finger, whereas you're not moving. I guess QML MouseArea drag attribute is not made to be used while a scaling transition occurs.
* long press for selection seems natural to me, it's working the same in text entries. Unselecting by click should be implemented, I agree.
* moving cursor between characters is more difficult! The Poppler method I used give boxes for words, not for characters. Besides, on a small device (phone, don't know for tablet), selection by word is quicker IMHO.
* no zoom when changing selection, yeh if you prefer (it took me days to figure out how to do this, but it was funny anyway).
* the two handles have an independent MouseArea (bigger than the rendered handle) so if the selection is small the two MouseArea overlap and the stop handle being on top, is prefered. No idea how to solve this easily.
* need to figure out to do vribation, but yeh it a very good idea.

So to summarise (before dealing with code comments):
* implement unselection on click ;
* remove zoom (will solve shift issue on dragging too quickly) ;
* add vibration ;
* discuss further character selection and handle overlapping.

pdf/pdfcanvas.cpp
dcaliste: Ah sorry always use abreviations that we use at work without thinking about it. Red is for reduced (*i.e.* coordinates in [0;1]). Could change redRect to reducedCoordRect. Still abbreviated but maybe clearer ?

pdf/pdflinkarea.cpp
dcaliste: I agree, would have prefered it like that also myself, but I copied it from pdfcanvas.cpp. Will change for member variable.

plugin/PDFView.qml
dcaliste: I agree, then the copy to clipboard should be done in the PDFSeelectionView object I guess.

pvuorela:
> don't understand first point, sorry :/ Do you mean that it's not necessary to move the black circles around when wanting to change the selection ?

Yeah, normal text selection in editors only has start and end cursors. Should be sufficient also here.

> moving cursor between characters is more difficult! The Poppler method I used give boxes for words, not for characters. Besides, on a small device (phone, don't know for tablet), selection by word is quicker IMHO

Acknowledged, though Poppler::Page does have charBoundingBox() too. Elsewhere we allow selection by character, so that would be more consistent, though by words could be good enough.

> no zoom when changing selection, yeh if you prefer

Should be easier as baseline too. Zooming can be implemented as polished feature separately if it feels worth it.

> need to figure out to do vribation, but yeh it a very good idea.

Silica -> TextBase, not too many lines needed.

Didn't yet go into details of this selection model while reviewing. Seemed complicated enough :)

pdf/pdfcanvas.cpp
pvuorela: Would be lot clearer already.

pdf/pdfselection.cpp
dcaliste: For the roles being initialised here, I did it after a comment in a previous PR, see https://github.com/sailfishos/sailfish-office/pull/36#discussion_r35949539 but reading again today, I notice that you already commented at that time that it can be put in the accessor of roles.

The i stand for index, will change this.

pdf/pdfselection.cpp
dcaliste: I'll give a look to this method. I have also some PDF (french layout) were character with accents are not in the same text box than the word itself.

I remember that Evince or Xpdf have this issue a long time ago. I should give a look on how Evince is doing text selection and implement the same.

dcaliste:
> Yeah, normal text selection in editors only has start and end cursors. Should be sufficient also here.
Yes, but the mouse cursor is freely moving and text selection is not necessary always strictly under the cursor, especially on complex layouts (not rectangular paragraphs) and the visual feedback of were is the mouse cursor is welcome then. I was thinking the same for the text selection here. I'll try anyway with handles being always on start / stop of the selection.

> Elsewhere we allow selection by character, so that would be more consistent, though by words could be good enough.
I agree with you, it would be more consistent. But I didn't want to go that direction first because it seems to me to be more complex to implement. You can have lot of characters in a selection and a QML model is not anymore a reasonable choice, will be too slow. Otherwise, one has to agregate all boxes to create a continuous selection (like in modern pdf reader in fact), but this job seems huge to me also. I would prefer to keep it for later improvements !

> Didn't yet go into details of this selection model while reviewing. Seemed complicated enough :)
Well, I can add documentation if you think it's necessary. The idea is the following:
* From Poppler we store (on demand) in PDFDocument a QList<QPair<QRectF, QString>> of all text boxes of a page. These text boxes are provided by Poppler in order from top page to bottom.
* The ListModel just stores two values for start position and two values for stop position. The first value *iPage* is the index of the page the handle is on and *iBox* is the index of the text box in the list.
* accessing a particular data *i* from the model is done by counting boxes from (iPageStart, iBoxStart) and stopping when *i* is reached.
* when a handle is moved, we change these two values iPage, iBox depending on drag position.

pdf/pdfselection.cpp
dcaliste: Well, here, I want to set a new coontent for the model, so do a clear and a new insert.

If I don't reset the model before inserting, doing a selection works fine, but selecting another word after make the previous selection still existing.

pdf/pdfselection.cpp
dcaliste: I use, as you suggest, the routine hasSpaceAfter() from Poppler, and use a little trick here. Poppler objects are only accessible in pdfrenderthread.cpp. So the hasSpaceAfter() is lost when reaching pdfselection.cpp, only rectangle and text are stored. The trick consists in appending a space to the stored string when hasSpaceAfter() is true. The text is then reconstructed by concatenating all the string without adding space inbetween each. This trick may leave a trailing space on the selected text.

dcaliste:
I've updated the PR. If I didn't forget any code comments, they all should be addressed (no abbreviations, space between functions, no space between words when Poppler don't say hasSpaceAfter, initialisation of roles in role acccessor, member attribute for the QTimer instead of dynamic allocation). I've also modified:
* copy text to clipboard on mouse ungrab and first selection only;
* remove the zoom effect on handle drag;
* add a feedback effect on first selection;
* unselect on clic on background.

Besides, I've added an animation when first selection is done, like in text fields, the handlers come from both sides of the screen.

What do you think (and design team) ?

pvuorela:
Feels lot nicer!

So, hows with text selection cursors staying at the start and end of text instead of moving finger?

Also one detail missing is that exiting text selection with click shouldn't change the split screen mode.

pdf/pdfdocument.cpp
```diff
@@ -108,6 +108,10 @@ PDFDocument::LinkMap PDFDocument::linkTargets() const
 {
     return d->thread->linkTargets();
 }
+const PDFDocument::TextList PDFDocument::textBoxesAtPage(int page)
```
pvuorela: Empty line as method separator.

A bit personal preference, but I think non-pointer/reference as const return type are pointless. Caller can assign to const variable if that seems fit.

pdf/pdfrenderthread.cpp
```diff
@@ -208,6 +230,13 @@ QMultiMap<int, QPair<QRectF, QUrl> > PDFRenderThread::linkTargets() const
     QMutexLocker(&d->thread->mutex);
     return d->linkTargets;
 }
+const QList< QPair< QRectF, QString > > PDFRenderThread::textBoxesAtPage(int page)
```
pvuorela: Empty line here as well. Inside method there are extra spaces on if (), and QMutexLocker has curly brace initialization not used elsewhere.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,447 @@
...
+#include "pdfdocument.h"
+
+class PDFSelection::Private
+{
+public:
+    Private(): canvas(nullptr) {
+        pageIndexStart = pageIndexStop = -1;
+        boxIndexStart = boxIndexStop = -1;
+
+        wiggle = 4.;
```
pvuorela: Had a comment if all should be in initialization list.

pdf/pdfrenderthread.cpp
```diff
@@ -126,7 +127,28 @@ class PDFRenderThreadPrivate
...
+        QList< QPair< QRectF, QString > > lst;
+        for (QList<Poppler::TextBox*>::const_iterator box = plst.begin();
+             box != plst.end(); box++) {
+            QSizeF psize = page->pageSize();
+            QRectF bbox = (*box)->boundingBox();
+            QRectF rect{ bbox.x() / psize.width(),
+                    bbox.y() / psize.height(),
+                    bbox.width() / psize.width(),
+                    bbox.height() / psize.height()};
+            lst.append(QPair<QRectF, QString>{rect, (*box)->text() + ((*box)->hasSpaceAfter() ? " " : "")});
```
pvuorela: As you said, this will almost always append a space on copied text, which feels unfortunate. How about class for text box instead of QPair? Not fully sure what to think of using Poppler one. In practice it looks safe, though Poppler doesn't seem to specify thread safety nearly at all.

Btw, noticed that on Poppler operations, toc model calls some outside render thread.

pdf/pdflinkarea.cpp
```diff
@@ -66,23 +73,40 @@ void PDFLinkArea::mousePressEvent(QMouseEvent *event)
...
 
-void PDFLinkArea::mouseReleaseEvent(QMouseEvent *event)
+void PDFLinkArea::mouseMoveEvent(QMouseEvent *event)
 {
+    // Don't activate anything if the finger has moved too far
+    QRect rect((d->clickLocation - QPointF(d->wiggleFactor, d->wiggleFactor)).toPoint(),
+               QSize(d->wiggleFactor * 2, d->wiggleFactor * 2));
+    if (!rect.contains(event->pos()))
+        return;
     d->clickInProgress = false;
```
pvuorela: ClickInProgress seems to be in practice the same thing as timer running. Could perhaps simplify code by just using timer state?

As a detail here the timer is left running but the handler checks clickInProgress.

plugin/PDFView.qml
```diff
@@ -32,9 +32,11 @@ SilicaFlickable {
     property alias currentPage: pdfCanvas.currentPage
 
     property bool scaled: pdfCanvas.width != width
+    property QtObject _feedbackEffect
 
     signal clicked()
     signal linkClicked(string linkTarget)
+    signal selectionClicked(string text)
```
pvuorela: Don't see any use for this.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,447 @@
...
+        d->boxIndexStart < 0 || d->boxIndexStop < 0)
+        return out;
+
+    if (!d->canvas)
+        return out;
+    PDFDocument *doc = d->canvas->document();
+    if (!doc)
+        return out;
+    
+    out = QLatin1String("");
```
pvuorela: Don't think this can be assumed latin.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,447 @@
...
+    return false;
+}
+
+QString PDFSelection::text() const
+{
+    QString out;
+
+    if (d->pageIndexStart < 0 || d->pageIndexStop < 0 ||
+        d->boxIndexStart < 0 || d->boxIndexStop < 0)
+        return out;
```
pvuorela: More nitpicking on style: braces if conditions are multiple lines.

pdf/pdfselection.cpp
pvuorela: Oh right, it's already declared QString. Then again this would be then more or less useless line?

pdf/pdfdocument.cpp
dcaliste: Line added and const removed (in pdfrenderthread.h also).

pdf/pdfrenderthread.cpp
dcaliste: Line added, space removed and curly braces changed for parenthesis.

pdf/pdfselection.cpp
dcaliste: Oh sorry just understand your remark now… Corrected.

pdf/pdfrenderthread.cpp
dcaliste: I've switched to use the Poppler structure  itself. I agree there is no point in reimplementing all Poppler objects. The only issue that makes me kept the QPair is that the bounding box is given in 1/72 of inches which requires to have the page size at hand always and do conversion each time we use the box. So I store a QPair<QRectF, Poppler::TextBox*>, instead of the previous QPair<QRectF, QString>.

Then space is added when mentioned and not for last word in selection.

pdf/pdflinkarea.cpp
dcaliste: clickInProgress was not introduced by me. I noticed that it's in fact unused but didn't touch it, not to interfere. I've now removed it completly. If you think I should not, I can put it back.

dcaliste: Well, in fact, clickInProgress has a different meaning than the timer. Both the timer and clickInProgress are triggered on press event. Then a motion event out of the wiggle should cancel both. Then the timer trigger and later click event on released is cancelled by clickInProgress to false.

But anyway, you are right there was a bug in the PR that makes moving a bit inside the wiggle cancel later click on release because the clickInProgress = false was not in the if. It's corrected.

plugin/PDFView.qml
dcaliste: It's a remaining of later implementation to allow a context menu on selection tap (to ask for annotation creation on the selection for instance). It's removed from this PR.

pdf/pdfselection.cpp
dcaliste: You're right, line removed.

pdf/pdfselection.cpp
dcaliste: Ah my bad, forgot this rule !

dcaliste:
Ok, thanks again for the quick and deep review. I've run over the comments one by one not to forget any and put a acknowledgement on each to be sure and not waste your time.

I've changed the click action when we have a selection to unselect it without opening the drawer.

About the handles not following the finger, it's not as trivial as I hoped. I was thinking about commenting some lines in QML here and there, but it's not working. The finger detection can not be done by simple MouseArea over the handles anymore. I will have to mess up with the PDF.linkArea to detect finger maybe. Well, I'll try to do it in a separate branch. When ready, I'll give you a pointer.

dcaliste:
I notice this morning that zooming while a selection is set present some issues with the handle positions. This was introduced in the first modifications to this PR when adding an animation to move the handle from screen border. I'll try to give a look today.

dcaliste:
Ok, I've modified a bit the PDFSelectionHandle.qml for the x position animation to be activated only when the selection appears and not whenver x is changed like before.

pvuorela:
Talked with design and selecting words and not able to move selection start beyond end were ok for start, but is was preferred to keep cursors at text instead of floating freely.

plugin/PDFSelectionView.qml
```diff
@@ -0,0 +1,58 @@
...
+        width: rect.width
+        height: rect.height
+    }
+
+    children: [
+        PDFSelectionHandle {
+            id: start
+            handle: root.model.start
+            onDragged: root.model.start = at
+            visible: root.visible
```
pvuorela: I think visible can be omitted on these, should follow visibility of parent by default.

pdf/pdfselection.h
```diff
@@ -0,0 +1,101 @@
...
+    void countChanged();
+    void canvasChanged();
+    void startChanged();
+    void stopChanged();
+    void textChanged();
+    void wiggleChanged();
+
+private:
+    class Private;
+    const QScopedPointer<Private> d;
```
pvuorela: Not a big deal, but plain pointers more commonly used.

dcaliste:
Thank you for having pushed to and discussed this with the design team. I'll implement the handle-stick-to-selection version and ping when it's ready for PR. This might take some time, because I'll be more occupied than I've been in the last ten days.

For the issue of start not being able to overpass end, I've thought about it recently, and maybe just swapping pageIndexStart (*resp.* boxIndexStart) with pageIndexStop (*resp.* boxIndexStop) in the start and stop setters should be fine. After all, it's the advantage of the selected text box list not being actually a list but just indices in a more global list. I will look at it after the handle stuff.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,433 @@
...
+                    break;
+                } else {
+                    boxIndex -= nBoxes;
+                }
+            }
+        }
+    }
+    if (pageIndex >= 0) {
+        const PDFDocument::TextList &boxes = doc->textBoxesAtPage(pageIndex);
+        const QPair<QRectF, Poppler::TextBox*> &box = boxes.at(boxIndex);
```
pvuorela: May crash if for some reason an invalid index is passed as parameter.

pvuorela: No sorry, pageIndex protects.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,433 @@
...
+    d->stop = QPointF(box.x() + box.width(), box.y() + box.height() / 2.);
+    emit stopChanged();
+
+    emit countChanged();
+    emit textChanged();
+}
+
+void PDFSelection::selectAt(const QPointF &point)
+{
+    unselect();
```
pvuorela: Fine as is, but optimally there wouldn't be unnecessary resets or inserts. Now with long press this does first reset even if there wasn't a selection. Protecting unselect() for that should be trivial. Rarer case of long press while selection exists could only do one reset to update state, but that complicates the code and not necessarily worth it.

pdf/pdflinkarea.cpp
```diff
@@ -66,17 +73,36 @@ void PDFLinkArea::mousePressEvent(QMouseEvent *event)
...
+
     // Don't activate anything if the finger has moved too far
-    QRect rect((d->clickLocation - QPointF(d->wiggleFactor, d->wiggleFactor)).toPoint(), QSize(d->wiggleFactor * 2, d->wiggleFactor * 2));
+    QRect rect((d->clickLocation - QPointF(d->wiggleFactor, d->wiggleFactor)).toPoint(),
+               QSize(d->wiggleFactor * 2, d->wiggleFactor * 2));
     if (!rect.contains(event->pos())) {
         return;
     }
 
+    event->accept();
```
pvuorela: Does this affect anything? No acceptance on press done and this is also after finger move check.

pvuorela:
(Moving comment hidden in earlier version here)
> Well, in fact, clickInProgress has a different meaning than the timer.

Would be if mouseReleaseEvent was allowed to emit click signals after long press. Don't think there's now a chance that clickInProgress and pressTimer.isActive() have different values. In such case clickInProgress could be removed and value checks replaced with isActive().

dcaliste:
For clickInProgress *vs.* pressTimer, you're absolutely right. I didn't thought about QTimer.isActive() (*i.e.* I didn't read enough the manual !). So I was thinking that I cannot detect that timer has run and stop.

pdf/pdfselection.cpp
dcaliste: Ok, I've protected the ```unselect()``` call when there is no selection already.

pdf/pdfselection.h
dcaliste: Ok corrected.

pdf/pdflinkarea.cpp
dcaliste: I've removed this line and cannot detect anything usual, so keep it without.

pvuorela:
Bug: crashes if moving selection cursor on top of the other.

dcaliste:
Thanks for pointing this out. I'm still working on the dragging code, not pleased with the result :(

Besides, handles are now start or stop and can be exchanged (bug appart) !

plugin/PDFView.qml
```diff
@@ -188,6 +217,23 @@ SilicaFlickable {
...
+            stopAttachX: base.width
+            onVisibleChanged: if (visible && _feedbackEffect) _feedbackEffect.play()
+            onDraggingChanged: {
+                if (dragging) {
+                    dragOffset.to = base.contentY + Theme.itemSizeSmall
+                    dragOffset.start()
+                } else {
+                    dragOffset.to = base.contentY - Theme.itemSizeSmall
+                    dragOffset.start()
+                }
```
pvuorela: Clicking on cursor handle keeps moving the view up little by little.

dcaliste: That's right, corrected.

dcaliste:
I've corrected the issue that makes the view moving up when taping on an handle while the shifting animation is not finished.

I've encountered the issue with the clip-board not registering changes anymore (selection is fine, but clipboard is not updated). It seems that the issue is in the clipboard code, since selecting text in any other application has the exact same behavior: not updating the clipboard content.

I cannot reproduce the crashing bug you reported when the two handles are put on top of each other. I've tried while dragging or by dropping on the other handle. Can you reproduce it easily and describe how to, please ?

Finally, I've a version where the selection handles and the drag source items are separated. Currently, I've kept a visible item for the shifted finger position in addition to the selection handle. You can remove it by commenting the rectangles at the end of PDFSelectionDrag.qml. I do prefer with them visible !

I think the difference between a text buffer where the cursor can follow the finger is that the cursor is bound to lines in text buffers. And space between paragraphs are represented by empty lines. Here in the PDF document, there is no line concept, so the cursor can be floating between lines or between paragraphs and it is not clear for the user where the cursor actually is in that case (typical case, the selection handle is in the middle of a line, one moves its finger a bit upward, the line gets completly selected with no clue where the cursor is. Then moving right or left doesn't change the selection since the cursor is actually floating between the two lines, but the user has no visual feedback).

pvuorela:
> I cannot reproduce the crashing bug you reported when the two handles are put on top of each other. I've tried while dragging or by dropping on the other handle. Can you reproduce it easily and describe how to, please ?

This is likely due to release version of Qt compiled with assertions off, while development build I use has them. I'm getting:
[F] qt_assert:2096 - ASSERT: "last >= first" in file itemmodels/qabstractitemmodel.cpp, line 2656

dcaliste:
> ASSERT: "last >= first" in file itemmodels/qabstractitemmodel.cpp, line 2656

Thanks a lot for the tips. With it, the assert was necessarily in ```beginRemoveRows()``` which is called only twice in ```pdfselection.cpp```. Then the solution is found: I didn't treat properly the case where ```nBoxes == 0``` in setStart() and setStop().

pdf/pdfcanvas.cpp
```diff
@@ -335,6 +338,15 @@ QRectF PDFCanvas::fromPageToItem(int index, const QRectF &rect)
                   rect.width() * page.rect.width(),
                   rect.height() * page.rect.height());
 }
+QPointF PDFCanvas::fromPageToItem(int index, const QPointF &point) const
```
pvuorela: Empty line before methods.

pdf/pdfdocument.cpp
```diff
@@ -108,6 +108,10 @@ PDFDocument::LinkMap PDFDocument::linkTargets() const
 {
     return d->thread->linkTargets();
 }
+PDFDocument::TextList PDFDocument::textBoxesAtPage(int page)
```
pvuorela: Empty line.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,486 @@
...
+
+    roles.insert(Rect, "rect");
+    roles.insert(Text, "text");
+
+    return roles;
+}
+
+QVariant PDFSelection::data(const QModelIndex& index, int role) const
+{
+    PDFDocument *doc = d->canvas->document();
```
pvuorela: For safety, all public methods could have a null check for canvas if accessed.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,486 @@
...
+    if (newCanvas != d->canvas) {
+        if (d->canvas)
+            d->canvas->disconnect(this);
+        d->canvas = newCanvas;
+
+        connect(d->canvas, &PDFCanvas::pageLayoutChanged, this, &PDFSelection::handle1Changed);
+        connect(d->canvas, &PDFCanvas::pageLayoutChanged, this, &PDFSelection::handle2Changed);
+        connect(d->canvas, &PDFCanvas::pageLayoutChanged, this, &PDFSelection::onLayoutChanged);
+
+        emit canvasChanged();
```
pvuorela: Should invalidate data and signal all other changes too?

plugin/PDFView.qml
```diff
@@ -188,6 +218,28 @@ SilicaFlickable {
                 height: match.height + Theme.paddingSmall / 2
             }
         }
+
+        PDFSelectionView {
+            id: selectionView
```
pvuorela: Cosmetic detail, but can omit ids when they are not referred anywhere.

plugin/PDFView.qml
```diff
@@ -188,6 +218,28 @@ SilicaFlickable {
...
+            model: selection
+            startAttachX: contentX
+            stopAttachX: contentX + base.width
```
pvuorela: contentXs are from base too? Could have explicit id in all.

pvuorela:
No more assert triggering, nice!

Couple details still: moving start handle beyond end keeps the earlier last word selected. Usually this would keep handle in place but select from its other side -> first word after earlier end should start selection. 

Also grabbing and immediately moving finger downwards confuses the offset, having selection done under one's finger. I can show both versions of floating drag to design later on, but this might affect the usability a bit too.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,486 @@
...
+    // point is given in canvas coordinates.
+    QPair<int, QRectF> at = canvas->pageAtPoint(point);
+    if (at.first < 0)
+        return;
+    *pageIndex = at.first;
+
+    const PDFDocument::TextList &boxes = canvas->document()->textBoxesAtPage(*pageIndex);
+    switch (position) {
+    case PDFSelection::Private::At: {
+        *boxIndex = -1;
```
pvuorela: Already set at method start.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,486 @@
...
+        }
+        break;
+    }
+    case PDFSelection::Private::Before: {
+        QPointF reducedCoordPoint {point.x() / at.second.width(),
+                (point.y() - at.second.y()) / at.second.height()};
+        *boxIndex = -1;
+        for (PDFDocument::TextList::const_iterator box = boxes.begin();
+             box != boxes.end(); box++) {
+            /* Stop counting boxes up as soon as a box after reducedCoordPoint is found. */
```
pvuorela: Got a first confused with Before and After. Though they would mean like "give indexes for text box before this point", but suppose it's more like "give indexes for text boxes where point is before it".

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,486 @@
...
+            if (box->first.y() > reducedCoordPoint.y() ||
+                (box->first.y() + box->first.height() > reducedCoordPoint.y() &&
```
pvuorela: Better not do too much for this PR, but the assumption doesn't necessarily hold for RTL languages.

Tested: ftp://ctan.tug.org/tex-archive/macros/latex/exptl/mem/arabic.pdf 
-> text gets selected as the flow would be LTR. Then again so does Firefox's internal pdf viewer and Evince. Not sure if pdf itself should have properly ordered boxes or should textDirection() be handled somehow. Maybe some TODO here?

pdf/pdfselection.cpp
pvuorela: No wait, so it's always after point but box start or end x coordinate?

pvuorela:
Selection at top of the page -> grab handle -> selection happens out of the screen.

pdf/pdfselection.cpp
```diff
@@ -0,0 +1,486 @@
...
+    if (pageIndex >= 0) {
+        const PDFDocument::TextList &boxes = doc->textBoxesAtPage(pageIndex);
+        const QPair<QRectF, Poppler::TextBox*> &box = boxes.at(boxIndex);
+        switch(role) {
+        case Rect:
+            result.setValue<QRectF>(d->canvas->fromPageToItem(pageIndex, box.first));
+            break;
+        case Text:
+            result.setValue<QString>(box.second->text());
+            break;
```
pvuorela: Doesn't matter much, but suppose Text role is left unused.

pdf/pdfcanvas.cpp
dcaliste: Corrected.

pdf/pdfdocument.cpp
dcaliste: Done.

pdf/pdfselection.cpp
dcaliste: Right, it's done elsewhere but not here, my mistake.

pdf/pdfselection.cpp
dcaliste: You're right. I've updated onLayoutChanged() and call here.

plugin/PDFView.qml
dcaliste: Removed.

pdf/pdfselection.cpp
dcaliste: Removed.

pdf/pdfselection.cpp
dcaliste: I've added lengthy comments to explain the before / after convention.

pdf/pdfselection.cpp
dcaliste: Yep, it's working only for LTR languages… I've added a TODO for RTL.

pdf/pdfselection.cpp
dcaliste: Yes, that's right. I think it doesn't cost anything to keep it here and provides data to QML in case someone wants to do something with the selection model. Do you prefer to have it removed as dead code ?

pvuorela: Nah, doesn't matter much being there or not, only few lines difference. However you like.

dcaliste:
Thank you for all the reviewing of yesterday. I've answered all inline comments not to miss any. Here remains some point that were not addressed in the inline comments:

> moving start handle beyond end keeps the earlier last word selected.

Yes, I need to change a bit more the internal code to handle the case when the selection becomes invalid. Currently there is no way to reduce the selection to zero by design of the code, so the transient state when moving the start handle after the end handle but without reaching the next word is not representable for the moment (there is a silent assumption in the code that boxIndexStart <= boxIndexStop I need to remove this assumption).

> grabbing and immediately moving finger downwards confuses the offset

Yes, I know, but I have no clue how to fix this or where it even comes from. The offset is done by changing contentY, I don't understand why moving the finger quickly confuse the translation. I thought that it comes from the flickable that steals the mouse before dragging becomes active, so I put drag.threshold to zero but with no effect :(

> Selection at top of the page -> grab handle -> selection happens out of the screen.

You're right, I agree it's bad. Besides, I don't really know how to handle this case. I mean, the idea, when dragging to offset the screen is to visualize what happen 'above' the finger, but beginning a dragging on top of screen means that we have no place to display 'above' the finger. So I don't know.

IMHO, I should correct the first point and leave the last one as it is. The second is really badly affecting the usability I agree and I should try to find a solution, but as I said I'm a bit clueless now.

plugin/PDFView.qml
dcaliste: Done.

dcaliste:
PS: pushed minor modifications in pdfselection.cpp to fulfill coding rule of no curly brace after if and for when content is single line and condition is also single line.

pvuorela:
> >    moving start handle beyond end keeps the earlier last word selected.

> Yes, I need to change a bit more the internal code to handle the case when the selection becomes invalid. Currently there is no way to reduce the selection to zero by design of the code, so the transient state when moving the start handle after the end handle but without reaching the next word is not representable for the moment (there is a silent assumption in the code that boxIndexStart <= boxIndexStop I need to remove this assumption).

Silica text editors allow selection state with start == end, but if it's easier to implement, guess it could be ok to keep word selected when handles overlap, changing selection to next word only when start moves beyond end, skipping the empty selection state. 

> >    Selection at top of the page -> grab handle -> selection happens out of the screen.

> You're right, I agree it's bad. Besides, I don't really know how to handle this case. I mean, the idea, when dragging to offset the screen is to visualize what happen 'above' the finger, but beginning a dragging on top of screen means that we have no place to display 'above' the finger. So I don't know.

Need to restrict animation's "to" with dragged handle's y position, or something like that, I guess. If problematic, could live with it for starters.

No immediate idea what to do with broken offset. Can try to investigate too, but  having lots of things to do at the moment.

dcaliste:
> selection to next word only when start moves beyond end, skipping the empty selection state. 

Sorry, I've just noticed that I didn't explained why this is problematic not to have the empty state. Take the case of the selection being the last word of a paragraph. Move the first selection past the end of the word horizontally, if I automatically select the next box, without having the transient state where nothing is selected during the time the user move the handle, I'm afraid it will be strange to see the first word of the next paragraph being selected while the handle is still being drag toward it.
Well, it's not a complicated job to introduce the empty state (I think), but it will take some days, depending on my available free time.

> If problematic, could live with it for starters.

I will give a look, as you said, at least to limit offset when there's no room on top of screen.

> Can try to investigate too, but having lots of things to do at the moment.

For sure, don't waste your time ! I will continue to look at this also, but I cannot promise anything since I have no clue where it comes from and not many ideas how to investigate it.

Anyway, code quality of this PR is improving in my opinion, so it's great.

dcaliste:
> grabbing and immediately moving finger downwards confuses the offset

I found the culprit this morning. I was not looking at the right place as usual in these circumpstances.

Well, the issue is in ```PDFSelectionDrag.qml```. There the drag centre is bound to handle1/2 of the model when *not* dragging. As soon as dragging starts, binding is released and its position is translated by the MouseArea drag effect. So when waiting for the offset transition to complete, the drag translations are applied on the offseted position, while when dragging immediatly, the binding is released too early and the offset is not applied by the transition when the drag translation takes place.

Ok, so now that I understand the issue, I should hopefully be able to find a better implementation to handle offset transition and drag.

dcaliste:
>  grabbing and immediately moving finger downwards confuses the offset

Finally corrected !

> Selection at top of the page -> grab handle -> selection happens out of the screen.

The offset is not applied if the selection is to close to the top, see line 162 in PDFView.qml.

> moving start handle beyond end keeps the earlier last word selected.

Not modified. I'm wondering: when in the transient state (no selection because the start moves beyond the stop but before the next box), the handles are not displayed anymore and will pop when the next box becomes selected. Is it something we want ? With the current situation, there is always a selection and the handles are clearly visible around the edges of the selection.

What do you think (the two corrections and the question) ?

pvuorela:
Checked with design again. Floating handles were approved, wouldn't necessarily be needed with character based selection but now with word based selection a change might need a lot of movement. 

As a small modification, now that there are four circles on the screen, it might help to make selection start and end cursors become bars when drag is active, similar how Silica text selection works. What do you think?

Seems also working lot better now. Still able to get some strange states if text doesn't flow conventionally, e.g. on your train ticket example I'm now having "|Nom du |Passager(s)" where last word is selected but cursors are around first two words. Basic case works, however, so I'm ok living with it for now.

> I'm wondering: when in the transient state (no selection because the start moves beyond the stop but before the next box), the handles are not displayed anymore and will pop when the next box becomes selected. Is it something we want ? With the current situation, there is always a selection and the handles are clearly visible around the edges of the selection.

If selection start and end position are the same, there should be a single handle, or the two on top of each other, on the same place.

plugin/PDFView.qml
```diff
@@ -120,9 +121,61 @@ SilicaFlickable {
...
+                    selectionOffset.start = base.contentY
+
+                // Limit offset when being at the bottom of the view.
+                selectionOffset.to = selectionOffset.start +
+                    Math.min(Theme.itemSizeSmall
+                             , Math.max(0, base.itemHeight - base.height - base.contentY))
+                // Limit offset when being at the top of screen
+                selectionOffset.to = Math.max(
+                    base.contentY
+                    , Math.min(selectionOffset.to
```
pvuorela: Cosmetics, but these commas at start of line are not that common style.

plugin/PDFSelectionDrag.qml
```diff
@@ -0,0 +1,101 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+
+MouseArea {
+    id: root
+
+    property bool press
```
pvuorela: MouseArea:pressed should be the same?

dcaliste:
> As a small modification, now that there are four circles on the screen, it might help to make selection start and end cursors become bars when drag is active, similar how Silica text selection works. What do you think?

Good suggestion, will try it today or tomorrow. Will also correct the two-inline comments. Thanks for the review.

plugin/PDFView.qml
dcaliste: Ok, got confused with the style for logical operators or inline test that should be written:
```
 cond1
 || cond2
```
and
```
test
? if-action
: else-action
```
when lines are too longs, if I'm not mistaken.

Anyway, I've corrected this comma style issue in PDFSelectionDrag.qml and PDFSelectionHandle.qml also.

plugin/PDFSelectionDrag.qml
dcaliste: True, press property removed and code updated accordingly.

dcaliste:
I've changed a bit PDFSelectionHandle.qml to have a dragged state that changes the cursor shape from circle to a vertical bar, as you suggested, it's quite nice and it helps to understand the selection when passing start over end and vice versa.

I've also changed the size of the selection handles and drag handles to use the same value as in Silica text item (namely Theme.iconSizeSmall / 2). It's nice and consistent. But currently PDF.Selection object lacks of a rowHeight property to be pixel perfect, so I've hard-coded the height of the bar to a value derived from Theme. It may work for the moment, but a proper calculation of rowHeight would be better. I'll try tomorrow as an average value of all the rect.height in the selection.

> Seems also working lot better now. Still able to get some strange states if text doesn't flow conventionally, e.g. on your train ticket example I'm now having "|Nom du |Passager(s)" where last word is selected but cursors are around first two words.

Yes, I've noticed this also. Last week, I've looked at the code of Evince to see how they are doing. Well, it seems that they are using functions of Poppler that are exported to GLib but not to Qt. It's the difference between list of words in the document and list of word *for selection*. These functions directly provide selection rectangles with proper height to span between lines (and ordered without to handle it ourself through coordinates of boxes)… I will try next month, I think, to call these functions of Poppler direcly to be able to have the same functionnalities than in poppler-glib. Maybe upstream them to poppler-qt if there is any interest. It's quite a pile of work, so you should not expect it soon, but there is no reason not to be able to have the same behaviour than in Evince.

pvuorela:
Looked good!

> I've also changed the size of the selection handles and drag handles to use the same value as in Silica text item (namely Theme.iconSizeSmall / 2). It's nice and consistent. But currently PDF.Selection object lacks of a rowHeight property to be pixel perfect, so I've hard-coded the height of the bar to a value derived from Theme. It may work for the moment, but a proper calculation of rowHeight would be better. I'll try tomorrow as an average value of all the rect.height in the selection.

Can be a bit funny when text is huge or small, but good enough already for merging. Just say if you want this version in first. If scaling the size, shouldn't height be the one used in the first/last selected word? 

> Yes, I've noticed this also. Last week, I've looked at the code of Evince to see how they are doing. Well, it seems that they are using functions of Poppler that are exported to GLib but not to Qt. It's the difference between list of words in the document and list of word for selection. These functions directly provide selection rectangles with proper height to span between lines (and ordered without to handle it ourself through coordinates of boxes)… I will try next month, I think, to call these functions of Poppler direcly to be able to have the same functionnalities than in poppler-glib. Maybe upstream them to poppler-qt if there is any interest. It's quite a pile of work, so you should not expect it soon, but there is no reason not to be able to have the same behaviour than in Evince.

Nice!

dcaliste:
> If scaling the size, shouldn't height be the one used in the first/last selected word?

I've added two properties that follow handle{1,2} changes, representing their respective heights and bind the bar height on these.

Beside undiscovered remaining bugs, I guess the PR looks fine now.

pvuorela:
Shiny! Was about to hit merge, but then encountered a crash. Select word or two from start of the page, rotate the device:

#0  PDFSelection::rectAt (this=0x2a4eecb0, index=-1) at /home/pvuorela/rpmbuild/BUILD/sailfish-office/pdf/pdfselection.cpp:148
#1  0xae97f06c in PDFSelection::handle1Height (this=<optimized out>) at /home/pvuorela/rpmbuild/BUILD/sailfish-office/pdf/pdfselection.cpp:386
#2  0xae981540 in PDFSelection::qt_metacall (this=0x2a4eecb0, _c=<optimized out>, _id=4, _a=0xbeffc810) at /home/pvuorela/rpmbuild/BUILD/sailfish-office/pdf/moc_pdfselection.cpp:227
#3  0xb63f613a in QV4::QObjectWrapper::getProperty(QObject*, QV4::ExecutionContext*, QQmlPropertyData*, bool) () from /usr/lib/libQt5Qml.so.5

dcaliste:
Oups, my fault when introducing the PDFSelection::rectAt() method. For convenience, to be able to use negative index *à la* Python, so rectAt(-1) gives the last rectangle in the selection, I use:
```index % count()```
which is *not* the mathematical modulus and doesn't provide result in ```[0;count()[```. I don't know how many time I should encounter this bug before remembering that operator ```%``` in not modulus…

My bad, corrected.

pvuorela:
Not crashing anymore, now hitting the merge button. Big thanks for this!

dcaliste:
Great ! Thank you also to you for all the long improvement process and review. If you're curious, you can give a look at the branch annotate in my github, it's the work that made this selection PR necessary.

It's quite functional now for creating, modifying and deleting text and highlight annotations. Remains a lot of polishing to do anyway:
- multipage annotations (for the user point of view since annotation are indeed saved per page)
- partial rendering (currently when modifying an annotation, the whole page is re-rendered). This may have a side effect that partial rendering for zoom purpose may be implemented also.
- support for caret annotation.
- clip of the PDFView that is not working when coming back from the annotation dialog
- finish UI for exportation, especially I've currently no clue how to nicely warn the user that he didn't export his modifications before returning to the document list.

PR #71, opened 2016-02-20
Correct two minor issues in the PDF toolbar.

dcaliste:
The two commits should correct two minor issues in the toolbar:
* the icon and the page current / count label are not highlighted when tapped on. The first commit correct this. I decided to make the first Item an IconButton to express the idea that the toolbar contains a button here with an icon, and also a label, thus  I added it as a children.
* the two prev / next buttons of the search result are quickly visible on the bottom left when clicking on the page icon and the page is transitioning to the index page. The commit should correct this by setting the visibility of the buttons only when they are needed.


pvuorela:
LGTM

PR #73, opened 2016-02-23
[office] Correct missing page label in detail page.

dcaliste:
The commit 83fd9a1398db747137f0e5caa0a63514a0aaa2e6 that removes the Calligra import in the details page breaks the switch in that page. As a result, the label for the page count is not displayed.

This PR corrects this, not by reintroducing the Calligra import, but by passing the mime type to the detail page and handling this accordingly.

### DIFF
--- a/plugin/DetailsPage.qml
+++ b/plugin/DetailsPage.qml
@@ -25,6 +25,7 @@ Page {
 
     property url source
     property int indexCount
+    property string mimeType
 
     FileInfo {
         id: info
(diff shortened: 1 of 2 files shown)


pvuorela:
Still avoiding import is very nice. Could have myself made the index count string property set by different types of document pages to avoid duplicating the switch case even more. For example PresentationPage should know that it contains only presentations.

Anyway, fixes a bug so good to go this way too.

dcaliste:
Yep, I don't like the duplication of the mimetype switch either. But I wanted something simple to correct the bug, as you said, but that didn't reintroduce the Calligra import for PDF documents…

One can keep in mind to factorise the switch code (it appears already twice in the Main.qml file itself) in a later PR.

pvuorela:
Alright. Merging.

PR #74, opened 2016-03-08
[sailfish-office] Restrict zoom level same amount in both orientations. 

pvuorela:
Too big zoom levels might hit texture size limits, leading to rendering failure.

### DIFF
--- a/plugin/PDFView.qml
+++ b/plugin/PDFView.qml
@@ -39,15 +39,8 @@ SilicaFlickable {
     signal updateSize(real newWidth, real newHeight)
 
     function clamp(value) {
-        if (value < width) {
-            return width
-        }
-
-        if (value > width * 2.5) {
-            return width * 2.5
-        }
-
-        return value
+        var maximumZoom = Math.min(Screen.height, Screen.width) * 2.5
+        return Math.max(width, Math.min(value, maximumZoom))
     }
 
     function zoom(amount, center) {


dcaliste:
Code looks fine for me and modification is sensible. Especially it means that zoom level is not changed when rotating, which is nice (currently being in landscape at max zoom level and rotate to portrait induce a change in zoom level that is not restored when rotating back to landscape).

Besides, for device in rotated orientation (landscape for phone or portrait for tablet), it means that the max zoom level compared to now will be limited. For the phone, the max zoom level in landscape will now be limited to 540 * 2.5 / 960 ~ 1.4 which is quite small considering that the previous 2.5 was sometimes too low to distinguish small details in the PDF. We really need to implement partial rendering. Maybe next job for me after annotations if not implemented by someone else before.

pvuorela:
Was thinking if we want to increase the restriction a bit as not making the worst case any worse, but then again 2.5 page size is quite a big texture anyway. Dunno.

jpetrell:
Would be nice to have some mechanism to detect the maximum texture size, this may break again later on. I guess then we should fix the texture size side. LGTM.

pvuorela:
Hopefully later on we'll get a bettering rendering setup, not a single huge texture for page but rather the needed visible parts.

dcaliste:
Well, in my opinion, there is no other way than partial rendering. I fully agree that for higher zoom levels, textures are way too big, especially on HD screens. As I said, your PR is good, introducing a consistent zoom level between landscape and portrait, and for this reason alone, I think it's a good think to let it in and correct the texture handling later.

PR #75, opened 2016-03-20
[office] Use shared mime-to-icon function. Fixes JB#32309

jpetrell:
Share icon mapping with the file manager. Use scalable icons from platform theme packages. Get rid of graphic-plugin-*, which are currently unused in Documents app. Fix vertical centering of labels in FileListPage list items.

This is a stopgap solution. If I have time I'll try to move icon declarations to desktop files to comply with freedesktop.org specifications.


qml/FileListPage.qml
```diff
@@ -154,55 +155,52 @@ Page {
...
+                    width: parent.width
+                    height: sizeLabel.height
+                    Label {
+                        id: sizeLabel
+                        text: Format.formatFileSize(model.fileSize)
+                        font.pixelSize: Theme.fontSizeExtraSmall
+                        color: listItem.highlighted ? Theme.secondaryHighlightColor : Theme.secondaryColor
+                    }
+                    Label {
+                        id: timestamp
```
dcaliste: This id is not refered to and may be removed.

dcaliste:
These are interesting modifications, bringing more code sharing when possible. I approuve them 100% but besides, waiting for Sailfish.FileManager to arrive into the public SDK, only Jolla will be capable of deploying sailfish-office to device and test modifications after this PR has been merged.

As a workaround, I can keep a branch of sailfish-office before the inclusion of this PR and rebase on it each time I want to test on device… A bit handy but working I guess.

I don't know what would be your general recommendation for case of migration like this one where package and SDK should ideally be kept synchronised.

pvuorela:
A bit mixed feeling about using Sailfish.FileManager. An extra dependency not available outside for somewhat simple mapping. Icons not available either atm, but those at least should start working.

Would suggest at least separating UI changes to its own commit so people outside can revert one commit and only affect the parts that handle icons.

jpetrell:
Separated layout and icon changes to own commits, but I am ok duplicating the mapping function as well.

Hopefully we get internal sub-modules public at some point to avoid such conflicts.

qml/FileListPage.qml
```diff
@@ -157,7 +157,7 @@ Page {
 
             Image {
                 id: icon
-                property string fileMimeType: window.mimeToIcon(model.fileMimeType)
+                property string fileMimeType: MimeFormatter.mimeTypeToIcon(model.fileMimeType)
```
dcaliste: Thanks for dividing the commit into two. Just here, this MimeFormatter should go to the other commit I guess (or I missed something).

jpetrell:
Updated: after discussing with pvuorela agreed to remove sailfish-components-filemanager dependency.

dcaliste:
Another possibility, I thought of this morning, would be to keep the QML with the MimeFormatter stuff and put temporarily the (part of the) code for this object into the sailfishoffice plugin code. Currently, in my opinion, this plugin already has some code in the file ```plugins/fileinfo.cpp``` that should be in Sailfish.FileManager to obtain file size, modification date… So like that, the QML is ready for the migration and when Sailfish.FileManager becomes public, the code in fileinfo.cpp and in mimeformatter.cpp (or equivalent) can be simply removed. What do you think ?

pvuorela:
LGTM, though could have still kept layout changes separate.

@dcaliste Sailfish.FileManager is more for UI parts. File sizes and modification times are UI-agnostic middleware details:
https://git.merproject.org/mer-core/nemo-qml-plugin-filemanager/merge_requests/2//diffs

jpetrell:
Updated icons are partially colored, removed "?" + Theme.primaryColor to avoid turning color parts white.

dcaliste:
@pvuorela you're right, didn't remember about the nemo-qml-plugin-filemanager. When this plugin will be accepted, there will be sense to remove ```plugins/fileinfo.cpp``` and use this plugin, wouldn't there ?

@jpetrell you may also want to update a Require entry in the spec file to point on something that has the new graphics inside. What do you think ?

pvuorela:
@dcaliste IMO somewhere borderline whether to move. Class is not too complicated either here and FileManager is mostly boosted FolderListModel. If there's more demand for similar functionality elsewhere, then yes.

jpetrell:
Now with rRequires to the icon package.

PR #76, opened 2016-03-23
[Office] Correct text selection in documents with columns.

dcaliste:
The previous code to change selection by moving handles does not work in a multi-column document (my bad). It resulted in always moving the selection edge into the first column.

This PR hopefully corrects it. This problem was difficult to tackle that I thought of initially. Currently, the code is scanning all text boxes, classifying them as "after" or "before" the "cursor" position. When a before to after transition is found, the before or after box is returned depending on the value of the position argument of the routine. This works well in a document where the layout go from top to bottom.

But in a multi-column document, the layout goes from top to bottom and returns to top again, several times. Thus it exists several before / after transitions for a given cursor position and deciding which one is the best can be tricky (think about when the cursor is in the margin between columns). Finally, I decided to sort these transitions by adopting the one that has the closest before / after boxes to the cursor position.

All this need to be tested. I've done testing for two days now, but maybe additional testing should be good…

### DIFF
--- a/pdf/pdfselection.cpp
+++ b/pdf/pdfselection.cpp
@@ -267,51 +267,85 @@ void PDFSelection::Private::textBoxAtPoint(const QPointF &point, Position positi
         }
         break;
     }
-    case PDFSelection::Private::Before: {
+    case PDFSelection::Private::Before:
         // Find the last box index in pageIndex that is before @point,
         // including at @point. If none is found in this page, returns
         // the last box of the previous page.
-        QPointF reducedCoordPoint {point.x() / at.second.width(),
-                (point.y() - at.second.y()) / at.second.height()};
-        for (PDFDocument::TextList::const_iterator box = boxes.begin();
-             box != boxes.end(); box++) {
-            /* Stop counting boxes up as soon as a box after reducedCoordPoint is found. */
-            if (box->first.y() > reducedCoordPoint.y() ||
-                (box->first.y() + box->first.height() > reducedCoordPoint.y() &&
-                 box->first.x() > reducedCoordPoint.x())) {
-                if (*boxIndex == -1) {
-                    *pageIndex -= 1;
-                    if (*pageIndex >= 0)
-                        *boxIndex = canvas->document()->textBoxesAtPage(*pageIndex).length() - 1;
-                }
-                return;
-            }
-            *boxIndex += 1;
-        }
-        *boxIndex = boxes.length() - 1;
-        break;
-    }
     case PDFSelection::Private::After: {
         // Find the first box index in pageIndex that is after @point,
         // including at @point. If none is found in this page, returns
         // the first box of the next page.
         QPointF reducedCoordPoint {point.x() / at.second.width(),
                 (point.y() - at.second.y()) / at.second.height()};
-        *boxIndex = 0;
+        
+        int index = 0;
+        PDFSelection::Private::Position previousBoxPosition = PDFSelection::Private::Before;
+        QRectF previousBox {0, -1, 1, 1};
+        qreal closestSquaredDistance = 2.;
+        int boxAfterIndex = boxes.length();
         for (PDFDocument::TextList::const_iterator box = boxes.begin();
              box != boxes.end(); box++) {
-            /* Stop counting boxes down as soon as a box after reducedCoordPoint is found. */
-            if (box->first.y() > reducedCoordPoint.y() ||
-                (box->first.y() + box->first.height() > reducedCoordPoint.y() &&
-                 box->first.x() + box->first.width() > reducedCoordPoint.x())) {
-                return;
+            PDFSelection::Private::Position currentBoxPosition =
+                (box->first.bottom() < reducedCoordPoint.y()
+                 || (box->first.top() < reducedCoordPoint.y()
+                     && box->first.left() < reducedCoordPoint.x()))
+                ? PDFSelection::Private::Before
+                : PDFSelection::Private::After;
+            if (previousBox.contains(reducedCoordPoint))
+                currentBoxPosition = PDFSelection::Private::After;
+            // Find a transition in the list of boxes. A transition is
+            // when the current box switch from being before the given point
+            // to being after.
+            if (previousBoxPosition == PDFSelection::Private::Before &&
+                currentBoxPosition == PDFSelection::Private::After) {
+                // The actual transition that will be kept is the closest
+                // to the given point.
+                QPointF boxCenter = previousBox.center();
+                qreal squaredDistance =
+                    (boxCenter.x() - reducedCoordPoint.x()) *
+                    (boxCenter.x() - reducedCoordPoint.x()) +
+                    (boxCenter.y() - reducedCoordPoint.y()) *
+                    (boxCenter.y() - reducedCoordPoint.y());
+                if (squaredDistance < closestSquaredDistance) {
+                    boxAfterIndex = index;
+                    closestSquaredDistance = squaredDistance;
+                }
+                boxCenter = box->first.center();
+                squaredDistance =
+                    (boxCenter.x() - reducedCoordPoint.x()) *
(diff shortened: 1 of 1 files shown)


dcaliste:
Oups, sorry, made a mistake when rebasing on master. I did it on an older version. I will correct this tomorrow morning. Sorry for the inconvenience.

dcaliste:
Ok, I've pushed the correct version again, rebase on master.

pvuorela:
Tested this and while selection worked a lot better, moving selection to last word in left side column of feature list, a big portion of right side column got also selected: 

http://www.pdf995.com/samples/pdf.pdf

pvuorela:
Btw, selecting the "approved" text was quite an interesting detail, also on desktop PDF readers.

pdf/pdfselection.cpp
```diff
@@ -267,51 +267,78 @@ void PDFSelection::Private::textBoxAtPoint(const QPointF &point, Position positi
...
         // Find the first box index in pageIndex that is after @point,
         // including at @point. If none is found in this page, returns
         // the first box of the next page.
         QPointF reducedCoordPoint {point.x() / at.second.width(),
                 (point.y() - at.second.y()) / at.second.height()};
-        *boxIndex = 0;
+        
+        int index = 0;
+        PDFSelection::Private::Position previousBoxPosition = PDFSelection::Private::Before;
+        PDFSelection::Private::Position currentBoxPosition;
```
pvuorela: Cosmetics, but since used only inside for loop, could limit scope by introducing variable there.

pdf/pdfselection.cpp
```diff
@@ -267,51 +267,78 @@ void PDFSelection::Private::textBoxAtPoint(const QPointF &point, Position positi
...
-            *boxIndex += 1;
+            previousBoxPosition = currentBoxPosition;
+            previousBox = box->first;
+            index += 1;
+        }
+        if (position == PDFSelection::Private::Before ||
+            (position == PDFSelection::Private::After &&
+             *boxIndex > 0 && *boxIndex <= boxes.length() &&
+             boxes[*boxIndex - 1].first.contains(reducedCoordPoint)))
+            *boxIndex -= 1;
```
pvuorela: Usually not that important, but here operators at line start and braces with multi-line if statement could help readability.

pdf/pdfselection.cpp
```diff
@@ -267,51 +267,78 @@ void PDFSelection::Private::textBoxAtPoint(const QPointF &point, Position positi
...
+        if (*boxIndex < 0) {
```
pvuorela: This can only be true if set in above line. Maybe move this block inside above if?

pdf/pdfselection.cpp
```diff
@@ -267,51 +267,78 @@ void PDFSelection::Private::textBoxAtPoint(const QPointF &point, Position positi
...
+                }
             }
-            *boxIndex += 1;
+            previousBoxPosition = currentBoxPosition;
+            previousBox = box->first;
+            index += 1;
+        }
+        if (position == PDFSelection::Private::Before ||
+            (position == PDFSelection::Private::After &&
+             *boxIndex > 0 && *boxIndex <= boxes.length() &&
```
pvuorela: boxes.length() is guaranteed here?

pdf/pdfselection.cpp
```diff
@@ -267,51 +267,78 @@ void PDFSelection::Private::textBoxAtPoint(const QPointF &point, Position positi
...
+                // The actual transition that will be kept is the closest
+                // to the given point.
+                QPointF boxCenter = previousBox.center();
+                qreal squaredDistance =
+                    (boxCenter.x() - reducedCoordPoint.x()) *
+                    (boxCenter.x() - reducedCoordPoint.x()) +
+                    (boxCenter.y() - reducedCoordPoint.y()) *
+                    (boxCenter.y() - reducedCoordPoint.y());
+                if (squaredDistance < closestSquaredDistance) {
+                    *boxIndex = index;
```
pvuorela: Maybe could help readability if boxIndex wouldn't be used for temporary state, but document the usage with e.g. int boxAfter?

pdf/pdfselection.cpp
```diff
@@ -267,51 +267,78 @@ void PDFSelection::Private::textBoxAtPoint(const QPointF &point, Position positi
...
+            (position == PDFSelection::Private::After &&
+             *boxIndex > 0 && *boxIndex <= boxes.length() &&
+             boxes[*boxIndex - 1].first.contains(reducedCoordPoint)))
+            *boxIndex -= 1;
+        if (*boxIndex < 0) {
+            *pageIndex -= 1;
+            if (*pageIndex >= 0)
+                *boxIndex = canvas->document()->textBoxesAtPage(*pageIndex).length() - 1;
+        }
+        if (*boxIndex == boxes.length()) {
```
pvuorela: What if previous page contains same amount of boxes and above code fetches index from it?

pdf/pdfselection.cpp
```diff
@@ -267,51 +267,78 @@ void PDFSelection::Private::textBoxAtPoint(const QPointF &point, Position positi
...
+                    closestSquaredDistance = squaredDistance;
+                }
             }
-            *boxIndex += 1;
+            previousBoxPosition = currentBoxPosition;
+            previousBox = box->first;
+            index += 1;
+        }
+        if (position == PDFSelection::Private::Before ||
+            (position == PDFSelection::Private::After &&
```
pvuorela: Doesn't matter much, but if position is not before, it's guaranteed to be after.

pdf/pdfselection.cpp
dcaliste: Right, moved inside.

pdf/pdfselection.cpp
dcaliste: That's true, I've moved the operators at the beginning of lines.

pdf/pdfselection.cpp
dcaliste: That's a very good suggestion. I'm using ```boxAfterIndex``` in the for loop and set ```*boxIndex``` to it or to it minus one after.

pdf/pdfselection.cpp
dcaliste: Good catch, I changed it for a ```else if``` statement.

pdf/pdfselection.cpp
dcaliste: I agree, but I like the remaining when reading.

pdf/pdfselection.cpp
dcaliste: That's true, I've removed this part of the test.

dcaliste:
I've addressed the various inline comments. Additionnaly, I've added a test to ensure that when the cursor is over a word, this is always considered as a transition between this word and the next. It corrects the issue of the last word of first column in your example document.

What do you think ?

pvuorela:
Tested the same document again. Moving selection to last word of left side column -> last word selected. Moving further to right -> big portion of next column selected even though cursor hadn't yet reached the column's left edge. But it's now possible to select only left side, so good enough to merge. And comparing with Evince, it changes selection to above lines which is also surprising.

LGTM

dcaliste:
> Moving further to right -> big portion of next column selected
Yes, it's because the left column is smaller than the right one. There will be various corner cases like this one because we're not dealing with columns in the algorithm, just with text flow that goes from top to bottom.

I thought about it much time last month when I discovered the issue but I was not avaible to find any consise solution. The current code is already quite messy in my opinion and it would not be a good idea to try to patch it for corner cases. The last trick to always put the transition at the cursor when it is on a word, should guarantee that it's always possible to move the selection on what we want, strange big selections appearing only when moving in space between paragraphs with unusal layouts.

So to be short, thanks for merging, knowing that there is (not too many) corner cases.

pvuorela:
> The current code is already quite messy in my opinion and it would not be a good idea to try to patch it for corner cases.

Agree.

PR #78, opened 2016-04-21
[office] Avoid to render partial page when full page can fit into limit.

dcaliste:
The additional request that you were reporting was due to how the shown page were rendered. With the last PR, any page with a visible surface was requesting a size of textureLimit centered on the middle of the visible part of the page. So when a page was rendered with just a small top portion, the requested part may not cover the full page even if a better centered textureLimit would have allow to render the full page anyway.

To be short, this PR adds the notion for the visible pages to request the full page if it fits into texture limit, whatever the visible part size. For areas that don't fit into texture limit, the request is the same than for the last PR, _i.e._ partial rendering centered on visible part.

### DIFF
--- a/pdf/pdfcanvas.cpp
+++ b/pdf/pdfcanvas.cpp
@@ -456,28 +456,34 @@ QSGNode* PDFCanvas::updatePaintNode(QSGNode *node, QQuickItem::UpdatePaintNodeDa
         QRect pageRect = {
             0, 0, d->renderWidth, int(page.rect.height() * renderingRatio)
         };
-        QRect showableArea = {
-            int(renderingRatio * (visibleArea.x() - float(window()->width() / 4.) - page.rect.x())),
-            int(renderingRatio * (visibleArea.y() - float(window()->height() / 4.) - page.rect.y())),
-            int(renderingRatio * (visibleArea.width() + float(window()->width() / 2.))),
-            int(renderingRatio * (visibleArea.height() + float(window()->height() / 2.)))
-        };
-        showableArea = showableArea.intersected(pageRect);
-        // Limit showableArea with textureLimit to avoid looping request
-        // for something too big.
-        textureLimit.moveCenter(showableArea.center());
-        showableArea = showableArea.intersected(textureLimit);
 
-        if (showPage && (page.texture == nullptr
-                         || page.renderWidth != d->renderWidth
-                         || (page.renderWidth == int(width()) &&
-                             !page.textureArea.contains(showableArea)))) {
-            if (!page.requested) {
-                d->document->requestPage(i, d->renderWidth, window(), textureLimit);
-                page.requested = true;
+        if (showPage) {
+            textureLimit.moveTo(0, 0);
+            bool fullPageFit = textureLimit.contains(pageRect);
+            QRect showableArea = {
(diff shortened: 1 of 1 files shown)


pvuorela:
LGTM. 

Btw, tested loading PDF Reference  1.7, some 1300 pages. Scrolling that around clearly has worse frame rate. Could suspect updatePaintNode() plays a role in it, iterating over all the pages etc. Loads for ages too.

Btw, part 2, was thinking if scroll velocity could/should be used for partial rendering requests, i.e. requesting a bit more into the direction where user is scrolling. Might not be worth it, just a thought.

dcaliste:
Thanks for merging.

> Could suspect updatePaintNode() plays a role in it, iterating over all the pages etc.

Yes, if profiling indeed points to ```updatePaintNode()``` that would be better to use a dichotomy algorithm to find the pages appearing in the visible part, since the page list is sorted in increasing y values. It would give a O(ln N) complexity instead of the current O(N). Added to the TODO list !

> Loads for ages too.

Well this I guess comes from Poppler. Evince should give long load time also. If not there's an unexpected bottle neck in sailfish-office.

> was thinking if scroll velocity could/should be used for partial rendering requests

I thought about this very quickly some time ago. There are several issues:
* we don't know the damping algorithm used by the flickable item to decelerate the visible position, so we cannot compute where the page will stop (if there is no additionnal user interaction).
* we don't know the current speed values which are required to compute the movement, we can only approximate them with finite differences and a clock, quite ugly (and possibly not enough accurate).

Because of these, I trashed the idea quickly but if you have some clues on these two points, that would be interesting to dig a bit further.

pvuorela:
> >    Could suspect updatePaintNode() plays a role in it, iterating over all the pages etc.

> Yes, if profiling indeed points to updatePaintNode() that would be better to use a dichotomy algorithm to find the pages appearing in the visible part, since the page list is sorted in increasing y values. It would give a O(ln N) complexity instead of the current O(N). Added to the TODO list !

Didn't go yet to profiling, but it is a high suspect since document size shouldn't affect that many places.

> > Loads for ages too.
> Well this I guess comes from Poppler. Evince should give long load time also. If not there's an unexpected bottle neck in sailfish-office.

My desktop machine is of course lot more powerful, but on it the document is opened almost instantly with Evince. Does make me think there's something fishy going around on sailfish-office side.

>     we don't know the damping algorithm used by the flickable item to decelerate the visible position, so we cannot compute where the page will stop (if there is no additionnal user interaction).
>    we don't know the current speed values which are required to compute the movement, we can only approximate them with finite differences and a clock, quite ugly (and possibly not enough accurate).

What I had in mind was keeping it lot simpler. If moving (fast enough?) to some direction, rendering request could ask for rectangle that includes the current view, but requested content outside it being more (not necessarily fully) on scroll direction. If user instantly stops scrolling, the current view texture will still arrive. Current speed should be available with horizontal/verticalVelocity properties.

But still, keeping it simple has its benefits too.

dcaliste:
> My desktop machine is of course lot more powerful, but on it the document is opened almost instantly with Evince. Does make me think there's something fishy going around on sailfish-office side.

Ah, good point, very interesting. I'll give a look, loading time should be of the same order of magnitude, not far behind.

PR #79, opened 2016-04-28
[office] Fix zero page PDF documents.

dcaliste:
When testing the 1310 pages PDF reference document to look for slowing factors, it happens that this PDF file cannot be opened with Poppler 0.24.5 (which is currently on my phone), it gives some errors and reports to have no page.

I notice so that zero-page document handling is broken (infinite loop requesting page size + no way to go back to document list). This PR should fix the two issues by setting the failure flag to doculment when the document has no page.


pvuorela:
LGTM.

Issue #80, opened 2016-05-10
Add initial support for annotations


dcaliste:
I open an issue here, so the discussion that has been started in the comments of another PR (see #58) may be continue here and easily found back.

The current status of the implementation is the following:
* annotations are created by selecting a text and tapping on the selection, via a context menu like for URL ;
* support for highlight annotation (hightlight, under-line…) and text annotation (only linked text) ;
* annotations can be modified by tapping on them (making a dialog appears in) or deleted by choosing the pull up menu in the annotation dialog ;
* annotations are automatically saved on-the-fly to a file with a tailing « -annotated » suffix, or the same if the suffix already exists.

There are still pending issues in the current implementation:
* page is fully rerendered each time an annotation is created or modified, blanking the view ;
* page is not cropped when the annotation dialog is moved away (accepted or rejected) ;
* saving to the same file (*i.e.* when modifying ```toto-annotated.pdf```) creates a broken PDF file that cannot be opened anymore ;
* there is no support for multi-page annotation (selecting text spanning over two pages and highlight it results in having text on the first page being annotated). This restriction exists in Evince also.
* the bounding box of a text annotation is wrong. It is set to the initial user selection, and not to the size of the icon used by Poppler to indicate a linked-annotation.

@pvuorela when you have time, you may ask the design team their opinion on this, as you suggested in some previous comments, to polish (or rework) the UI part. What's your opinion on this and specifically the mentionned issues ? This is currently not fast moving because of the previously mentionned issues that are tricky (broken saved file, or non-crop page) or not so easy to implement (partial page rendering), but hopefully I can address all in the coming weeks.

pvuorela:
I'll try to bug design more on these. There was some initial plan how annotations could be done UI-wise but still not finished due to tons of things proceeding.

> page is not cropped when the annotation dialog is moved away (accepted or rejected)

Same issue maybe as backstepping to front page doesn't show it until pdf page is fully away? I think something on scene graph level.

> saving to the same file (i.e. when modifying toto-annotated.pdf) creates a broken PDF file that cannot be opened anymore

Hmh, any idea what's wrong with resulting file? Poppler bug somehow?

> there is no support for multi-page annotation (selecting text spanning over two pages and highlight it results in having text on the first page being annotated). This restriction exists in Evince also.

IMO this should be ok.

dcaliste:
> Hmh, any idea what's wrong with resulting file? Poppler bug somehow?

[...]

PR #81, opened 2016-06-01
Add support for highlight and linked text annotations

dcaliste:
As discussed in #80 , here is the merge request to add support for highlight and text annotations. I'm using it for days now, and I didn't notice any issues. Maybe, one can start reviewing the C++ code when time is available and leave the UI parts for later, or even for a later PR when improved design will be ready.

What do you think ?

### DIFF
--- a/pdf/pdfannotation.cpp
+++ b/pdf/pdfannotation.cpp
@@ -306,7 +306,7 @@ void PDFHighlightAnnotation::setStyle(PDFHighlightAnnotation::HighlightType valu
     emit styleChanged();
     
     if (d->m_document != nullptr && d->m_page >= 0)
-        d->m_document->onPageModified(d->m_page);
+        d->m_document->onPageModified(d->m_page, d->m_annotation->boundary());
 }
 
 void PDFHighlightAnnotation::attach(PDFDocument *document, PDFSelection *selection)
--- a/pdf/pdfannotation.cpp
+++ b/pdf/pdfannotation.cpp
@@ -1,0 +1,346 @@
+/* -*- c-basic-offset: 4 -*- */
+/*
+ * Copyright (C) 2016 Caliste Damien.
+ * Contact: Damien Caliste <dcaliste@free.fr>
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; version 2 only.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+#include "pdfannotation.h"
+
+static QString currentAuthor;
+
+class PDFAnnotation::Private
+{
+public:
+    Private(Poppler::Annotation *annotation, PDFDocument *document = nullptr, int page = -1)
+        : owner(document == nullptr || page < 0)
+        , m_annotation(annotation)
+        , m_document(document)
+        , m_page(page)
+    {
+    }
+    ~Private()
+    {
+        if (owner)
+            delete m_annotation;
+    }
+    bool owner;
+    Poppler::Annotation *m_annotation;
+    PDFDocument *m_document;
+    int m_page;
+};
+
+PDFAnnotation::PDFAnnotation(Poppler::Annotation *annotation, QObject *parent)
+    : QObject(parent), d(new PDFAnnotation::Private(annotation))
+{
+    if (annotation && !currentAuthor.isEmpty())
+        annotation->setAuthor(currentAuthor);
+}
+
+PDFAnnotation::PDFAnnotation(Poppler::Annotation *annotation,
+                             PDFDocument *document, int ipage, QObject *parent)
+    : QObject(parent), d(new PDFAnnotation::Private(annotation, document, ipage))
+{
+}
+
+PDFAnnotation::~PDFAnnotation()
+{
+    delete d;
+}
+
+PDFDocument* PDFAnnotation::document() const
+{
+    return d->m_document;
+}
+
+int PDFAnnotation::page() const
+{
+    return d->m_page;
+}
+
+QRectF PDFAnnotation::boundary() const
+{
+    return d->m_annotation->boundary();
+}
+
+void PDFAnnotation::attachOnce(PDFDocument *document, int page)
+{
+    /* Attach annotation only once. */
+    if (d->m_document != nullptr || document == nullptr)
+        return;
+    if (d->m_page >= 0 || page < 0 || page >= document->pageCount())
+        return;
+
+    d->m_document = document;
+    d->m_page = page;
+}
+
+void PDFAnnotation::attach(PDFDocument *document, PDFSelection *selection)
+{
+    if (!selection || selection->count() == 0)
+        return;
+
+    QPair<int, QRectF> word = selection->rectAt(0);
+    attachOnce(document, word.first);
+    
+    QRectF bounds = word.second;
+    for (int i = 1; i < selection->count(); i++)
+        bounds.united(selection->rectAt(i).second);
+
+    d->m_annotation->setBoundary(bounds);
+    d->m_document->addAnnotation(d->m_annotation, d->m_page);
+    emit attached();
+}
+
+void PDFAnnotation::remove()
+{
+    if (d->m_document && d->m_annotation) {
+        d->m_document->removeAnnotation(d->m_annotation, d->m_page);
+        d->m_annotation = nullptr;
+    }
+
+    deleteLater();
+}
+
+QString PDFAnnotation::author() const
+{
+    return d->m_annotation->author();
+}
+
+void PDFAnnotation::setAuthor(const QString &value)
+{
+    // Store last given author to use it as default for
+    // newly created annotations.
+    currentAuthor = value;
+
+    if (d->m_annotation->author() == value)
+        return;
+    d->m_annotation->setAuthor(value);
+    emit authorChanged();
+
+    d->m_annotation->setModificationDate(QDateTime());
+    emit modificationDateChanged();
+
+    if (d->m_document != nullptr)
+        d->m_document->setDocumentModified();
+}
+
+QString PDFAnnotation::contents() const
+{
+    return d->m_annotation->contents();
+}
+
+void PDFAnnotation::setContents(const QString &value)
+{
+    if (d->m_annotation->contents() == value)
+        return;
+
+    d->m_annotation->setContents(value);
+    emit contentsChanged();
+
+    d->m_annotation->setModificationDate(QDateTime());
+    emit modificationDateChanged();
+
+    if (d->m_document != nullptr)
+        d->m_document->setDocumentModified();
+}
+
+QDateTime PDFAnnotation::creationDate() const
+{
+    return d->m_annotation->creationDate();
+}
+
+QDateTime PDFAnnotation::modificationDate() const
+{
+    return d->m_annotation->modificationDate();
+}
+
+QColor PDFAnnotation::color() const
+{
+    return d->m_annotation->style().color();
+}
+
+void PDFAnnotation::setColor(const QColor &value)
+{
+    if (color() == value)
+        return;
+
+    Poppler::Annotation::Style style = d->m_annotation->style();
+    style.setColor(value);
+    d->m_annotation->setStyle(style);
+    emit colorChanged();
+
+    if (d->m_document != nullptr && d->m_page >= 0)
+        d->m_document->onPageModified(d->m_page, d->m_annotation->boundary());
+}
+
+PDFAnnotation::SubType PDFAnnotation::type() const
+{
+    return PDFAnnotation::SubType(d->m_annotation->subType());
+}
+
+
+static QStringList iconNames = (QStringList() << "Note" << "Comment" << "Key" << "Help" << "NewParagraph" << "Paragraph" << "Insert" << "Cross" << "Circle");
+
+PDFTextAnnotation::PDFTextAnnotation(QObject *parent)
+    : PDFAnnotation(new Poppler::TextAnnotation(Poppler::TextAnnotation::Linked), parent)
+{
+}
+
+PDFTextAnnotation::PDFTextAnnotation(Poppler::TextAnnotation *annotation,
+                                     PDFDocument *document, int ipage, QObject *parent)
+    : PDFAnnotation(annotation, document, ipage, parent)
+{
+}
+
+PDFTextAnnotation::~PDFTextAnnotation()
+{
+}
+
+void PDFTextAnnotation::attach(PDFDocument *document, PDFSelection *selection)
+{
+    if (!selection)
+        return;
+
+    QPair<int, QRectF> word = selection->rectAt(0);
+    attachAt(document, word.first, word.second.x(), word.second.y());
+}
+
+void PDFTextAnnotation::attachAt(PDFDocument *document,
+                                 unsigned int page, qreal x, qreal y)
+{
+    attachOnce(document, page);
+
+    // The size of the icon used to render the note on the document
+    // is hard-coded in Poppler to 24 1/72 of inch, see AnnotText::draw()
+    // of poppler/Annot.cc file. So we use the same size for the bounding box.
+    d->m_annotation->setBoundary(QRectF(x, y, 24., 24.));
+    d->m_document->addAnnotation(d->m_annotation, d->m_page, true);
+    emit attached();
+}
+
+PDFTextAnnotation::IconType PDFTextAnnotation::icon() const
+{
+    Poppler::TextAnnotation *anno = static_cast<Poppler::TextAnnotation*>(d->m_annotation);
+    return PDFTextAnnotation::IconType(iconNames.indexOf(anno->textIcon()));
+}
+
+void PDFTextAnnotation::setIcon(PDFTextAnnotation::IconType value)
+{
+    if (icon() == value)
+        return;
+
+    int index = int(value);
+    if (index > iconNames.length() || index < 0) {
+        qWarning() << QStringLiteral("Wrong icon for text annotation (%1).").arg(index);
+        return;
+    }
+    
+    Poppler::TextAnnotation *anno = static_cast<Poppler::TextAnnotation*>(d->m_annotation);
+    anno->setTextIcon(iconNames[value]);
+    emit iconChanged();
+    
+    if (d->m_document != nullptr && d->m_page >= 0)
+        d->m_document->onPageModified(d->m_page, d->m_annotation->boundary());
+}
+
+PDFCaretAnnotation::PDFCaretAnnotation(QObject *parent)
+    : PDFAnnotation(new Poppler::CaretAnnotation(), parent)
+{
+    Poppler::CaretAnnotation *anno = static_cast<Poppler::CaretAnnotation*>(d->m_annotation);
+    anno->setCaretSymbol(Poppler::CaretAnnotation::CaretSymbol::P);
+}
+
+PDFCaretAnnotation::PDFCaretAnnotation(Poppler::CaretAnnotation *annotation,
+                                       PDFDocument *document, int ipage, QObject *parent)
+    : PDFAnnotation(annotation, document, ipage, parent)
+{
+}
+
+PDFCaretAnnotation::~PDFCaretAnnotation()
+{
+}
+
+PDFHighlightAnnotation::PDFHighlightAnnotation(QObject *parent)
+    : PDFAnnotation(new Poppler::HighlightAnnotation(), parent)
+{
+    setColor(QColor(255, 255, 0, 255));
+}
+
+PDFHighlightAnnotation::PDFHighlightAnnotation(Poppler::HighlightAnnotation *annotation,
+                                               PDFDocument *document, int ipage,
+                                               QObject *parent)
+    : PDFAnnotation(annotation, document, ipage, parent)
+{
+}
+
+PDFHighlightAnnotation::~PDFHighlightAnnotation()
+{
+}
+
+PDFHighlightAnnotation::HighlightType PDFHighlightAnnotation::style() const
+{
+    Poppler::HighlightAnnotation *anno = static_cast<Poppler::HighlightAnnotation*>(d->m_annotation);
+    return PDFHighlightAnnotation::HighlightType(anno->highlightType());
+}
+
+void PDFHighlightAnnotation::setStyle(PDFHighlightAnnotation::HighlightType value)
+{
+    if (style() == value)
+        return;
+    
+    Poppler::HighlightAnnotation *anno = static_cast<Poppler::HighlightAnnotation*>(d->m_annotation);
+    anno->setHighlightType(Poppler::HighlightAnnotation::HighlightType(value));
+    emit styleChanged();
+    
+    if (d->m_document != nullptr && d->m_page >= 0)
+        d->m_document->onPageModified(d->m_page);
+}
+
+void PDFHighlightAnnotation::attach(PDFDocument *document, PDFSelection *selection)
+{
+    if (!selection)
+        return;
+
+    QPair<int, QRectF> word = selection->rectAt(0);
+    attachOnce(document, word.first);
+
+    QRectF bounds = word.second;
+    QList<Poppler::HighlightAnnotation::Quad> quads;
+    int count = selection->count();
+
+    for (int i = 0; i < count; i++) {
+        word = selection->rectAt(i);
+        Poppler::HighlightAnnotation::Quad quad;
+        if (word.first == d->m_page) {
+            quad.capEnd = (i == count - 1);
+            quad.capStart = (i == 0);
+            quad.feather = 0.;
+            quad.points[0] = QPointF(word.second.left(), word.second.top());
+            quad.points[1] = QPointF(word.second.right(), word.second.top());
+            quad.points[2] = QPointF(word.second.right(), word.second.bottom());
+            quad.points[3] = QPointF(word.second.left(), word.second.bottom());
+            quads.append(quad);
+            bounds |= word.second;
+        }
+    }
+
+    Poppler::HighlightAnnotation *annotation =
+        static_cast<Poppler::HighlightAnnotation*>(d->m_annotation);
+    annotation->setHighlightQuads(quads);
+    d->m_annotation->setBoundary(bounds);
+    d->m_document->addAnnotation(d->m_annotation, d->m_page);
+    emit attached();
+}
--- a/pdf/pdfannotation.h
+++ b/pdf/pdfannotation.h
@@ -1,0 +1,185 @@
+/*
+ * Copyright (C) 2016 Caliste Damien.
+ * Contact: Damien Caliste <dcaliste@free.fr>
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; version 2 only.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+#ifndef PDFANNOTATION_H
+#define PDFANNOTATION_H
+
+#include <poppler-qt5.h>
+
+#include "pdfdocument.h"
+#include "pdfselection.h"
+
+class PDFTextAnnotation;
+class PDFHighlightAnnotation;
+
+class PDFAnnotation : public QObject
+{
+    Q_OBJECT
+
+    Q_ENUMS(SubType)
+
+    Q_PROPERTY(PDFDocument* document READ document NOTIFY attached)
+    Q_PROPERTY(int page READ page NOTIFY attached)
+    Q_PROPERTY(QRectF boundary READ boundary NOTIFY attached)
+
+    Q_PROPERTY(QString author READ author WRITE setAuthor NOTIFY authorChanged)
+    Q_PROPERTY(QString contents READ contents WRITE setContents NOTIFY contentsChanged)
+    Q_PROPERTY(QDateTime creationDate READ creationDate CONSTANT)
+    Q_PROPERTY(QDateTime modificationDate READ modificationDate NOTIFY modificationDateChanged)
+    Q_PROPERTY(QColor color READ color WRITE setColor NOTIFY colorChanged)
+    Q_PROPERTY(SubType type READ type CONSTANT)
+
+public:
+    enum SubType {
+        Base,
+        Text,
+        Line,
+        GeometricFigure,
+        Highlight,
+        Stamp,
+        InkPath,
+        Link,
+        Caret,
+        FileAttachment,
+        Sound,
+        Movie,
+        Screen,
+        Widget
+    };
+
+    PDFAnnotation(QObject *parent = 0);
+    PDFAnnotation(Poppler::Annotation *annotation, QObject *parent = 0);
+    PDFAnnotation(Poppler::Annotation *annotation,
+                  PDFDocument *document, int ipage, QObject *parent = 0);
+    ~PDFAnnotation();
+
+    SubType type() const;
+
+    PDFDocument* document() const;
+    int page() const;
+    QRectF boundary() const;
+    Q_INVOKABLE void attach(PDFDocument *document, PDFSelection *selection);
+    Q_INVOKABLE void remove();
+
+    QString author() const;
+    void setAuthor(const QString &value);
+
+    QString contents() const;
+    void setContents(const QString &value);
+
+    QDateTime creationDate() const;
+
+    QDateTime modificationDate() const;
+
+    QColor color() const;
+    void setColor(const QColor &value);
+
+Q_SIGNALS:
+    void attached();
+    void authorChanged();
+    void contentsChanged();
+    void creationDateChanged();
+    void modificationDateChanged();
+    void colorChanged();
+
+protected:
+    class Private;
+    Private *d;
+
+    void attachOnce(PDFDocument *document, int page);
+};
+
+class PDFTextAnnotation : public PDFAnnotation
+{
+    Q_OBJECT
+
+    Q_ENUMS(IconType)
+
+    Q_PROPERTY(IconType icon READ icon WRITE setIcon NOTIFY iconChanged)
+
+public:
+    enum IconType {
+        Note,
+        Comment,
+        Key,
+        Help,
+        NewParagraph,
+        Paragraph,
+        Insert,
+        Cross,
+        Circle
+    };
+
+    PDFTextAnnotation(QObject *parent = 0);
+    PDFTextAnnotation(Poppler::TextAnnotation *annotation,
+                      PDFDocument *document, int ipage, QObject *parent = 0);
+    ~PDFTextAnnotation();
+
+    Q_INVOKABLE void attach(PDFDocument *document, PDFSelection *selection);
+    Q_INVOKABLE void attachAt(PDFDocument *document,
+                              unsigned int page, qreal x, qreal y);
+
+    IconType icon() const;
+    void setIcon(IconType value);
+
+Q_SIGNALS:
+    void iconChanged();
+};
+
+class PDFCaretAnnotation : public PDFAnnotation
+{
+    Q_OBJECT
+
+public:
+    PDFCaretAnnotation(QObject *parent = 0);
+    PDFCaretAnnotation(Poppler::CaretAnnotation *annotation,
+                       PDFDocument *document, int ipage, QObject *parent = 0);
+    ~PDFCaretAnnotation();
+};
+
+class PDFHighlightAnnotation : public PDFAnnotation
+{
+    Q_OBJECT
+
+    Q_ENUMS(HighlightType)
+
+    Q_PROPERTY(HighlightType style READ style WRITE setStyle NOTIFY styleChanged)
+
+public:
+    enum HighlightType {
+        Highlight,
+        Squiggly,
+        Underline,
+        StrikeOut
+    };
+
+    PDFHighlightAnnotation(QObject *parent = 0);
+    PDFHighlightAnnotation(Poppler::HighlightAnnotation *annotation,
+                           PDFDocument *document, int ipage, QObject *parent = 0);
+    ~PDFHighlightAnnotation();
+
+    Q_INVOKABLE void attach(PDFDocument *document, PDFSelection *selection);
+
+    HighlightType style() const;
+    void setStyle(HighlightType value);
+
+Q_SIGNALS:
+    void styleChanged();
+};
+
+#endif // PDFANNOTATION_H
--- a/pdf/pdfcanvas.cpp
+++ b/pdf/pdfcanvas.cpp
@@ -380,6 +380,57 @@ QPair<QUrl, PDFCanvas::ReducedBox> PDFCanvas::urlAtPoint(const QPointF &point) c
     return QPair<QUrl, PDFCanvas::ReducedBox>();
 }
 
+QPair<Poppler::Annotation*, PDFCanvas::ReducedBox> PDFCanvas::annotationAtPoint(const QPointF &point) const
+{
+    for (int i = 0; i < d->pageCount; ++i) {
+        const PDFPage &page = d->pages.value(i);
+        if (page.rect.contains(point)) {
+            qreal squaredDistanceMin = d->linkWiggle * d->linkWiggle;
+            Poppler::Annotation *result = nullptr;
+            QRectF at;
+            for (Poppler::Annotation *annotation : d->document->annotations(i)) {
+                switch (annotation->subType()) {
+                case (Poppler::Annotation::ALink):
+                    // Ignore link annotation for the moment since
+                    // real link are reported as annotation also.
+                    break;
+                case (Poppler::Annotation::AHighlight): {
+                    QList<Poppler::HighlightAnnotation::Quad> quads =
+                        static_cast<Poppler::HighlightAnnotation*>(annotation)->highlightQuads();
+                    for (QList<Poppler::HighlightAnnotation::Quad>::iterator quad = quads.begin();
+                         quad != quads.end(); quad++) {
+                        // Assuming rectangular quad...
+                        qreal squaredDistance =
+                            squaredDistanceFromRect(page.rect, QRectF(quad->points[0], quad->points[2]), point);
+                
+                        if (squaredDistance < squaredDistanceMin) {
+                            result = annotation;
+                            at = QRectF(quad->points[0], quad->points[2]);
+                            squaredDistanceMin = squaredDistance;
+                        }
+                    }
+                    break;
+                }
+                default: {
+                    qreal squaredDistance =
+                        squaredDistanceFromRect(page.rect, annotation->boundary(), point);
+                
+                    if (squaredDistance < squaredDistanceMin) {
+                        result = annotation;
+                        at = annotation->boundary();
+                        squaredDistanceMin = squaredDistance;
+                    }
+                    break;
+                }
+                }
+            }
+            return QPair<Poppler::Annotation*, PDFCanvas::ReducedBox>{result, {i, at}};
+        }
+    }
+
+    return QPair<Poppler::Annotation*, PDFCanvas::ReducedBox>();
+}
+
 QRectF PDFCanvas::fromPageToItem(int index, const QRectF &rect) const
 {
     if (index < 0 || index >= d->pageCount)
--- a/pdf/pdfcanvas.cpp
+++ b/pdf/pdfcanvas.cpp
@@ -177,6 +177,7 @@ void PDFCanvas::setDocument(PDFDocument *doc)
         connect(d->document, &PDFDocument::pageFinished, this, &PDFCanvas::pageFinished);
         connect(d->document, &PDFDocument::pageSizesFinished, this, &PDFCanvas::pageSizesFinished);
         connect(d->document, &PDFDocument::documentLockedChanged, this, &PDFCanvas::documentLoaded);
+        connect(d->document, &PDFDocument::pageModified, this, &PDFCanvas::pageModified);
 
         if (d->document->isLoaded())
             documentLoaded();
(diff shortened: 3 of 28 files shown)


anandrkris:
Maybe we  we have an rpm it openrepos or elsewhere for users to test and give feedback?
Also, there is a TJC for this.
https://together.jolla.com/question/25613/feature-request-annotating-and-highlighting-documents/?answer=136640#post-id-136640

dcaliste:
Hello, indeed I thought to build an RPM but differ it because:
* I've installed a more recent version of Poppler in my SDK than available on the phone, so build executable won't install properlly on the phone.
* My PR is built on top of master version of sailfish-office which contains some dependencies that are not present on the phone (system icons for documents).

The coming 2.0.2.x version may contain the two prerequisitions mentioned above. Then, I can build an RPM package for everyone to test.

anandrkris:
Oh okies. Thanks for clarifying. Appreciate all your good work on this. :smile:

pdf/pdflinkarea.cpp
```diff
@@ -94,18 +156,68 @@ void PDFLinkArea::mouseReleaseEvent(QMouseEvent *event)
...
+    // - click.
+
+    if (d->clickOnSelection) {
+      emit selectionClicked();
+      return;
+    }
+
+    if (d->selection && d->selection->count() > 0) {
+      d->selection->unselect();
+      return;
```
pvuorela: Nitpick: some bad indentation here and in mousePress.

pvuorela:
I'll do a limited review while (still) waiting for new UI plan. Also vacation coming up soon.

One thing noted is clicking on highlighted text. Shows press effect of filled rectangle which can be quite much outside the highlight with two column layout. Adobe seems to show only bounding rect edges. Maybe could do something similar.

Related to above, maybe could show only the edges for links too. Now image as link doesn't show as it's supposed to.

pvuorela:
Or at least filled rectangle could be more subtle, more transparency.

pdf/pdflinkarea.cpp
dcaliste: Good catch, emacs wrongly configured ;)

dcaliste:
Thank you for taking time looking at this. I understand that's it's a limited review waiting for design suggestions. Have nice coming holidays, I'll be on vacation also first part of July !

By the way, your remark on visual feedback when tapping on a sensitive part (link or annotation) is exactly the kind of remark that I was waiting for by submitting this early like that. Indeed, you're right, I didn't thought about the two columns spaning annotation but it's really ugly as it is now. Having the visual feedback exactly on each rectangle will imply an important rework for this because as you've seen, currently it's easily implemented with a rectangle property exported to QML.

I'll have to think of something else, but I don't want to through away this visual feedback, I think it brings something additional even for simple links, don't you think ?

dcaliste:
Just in case, these commits correspond to work in progress. I still need to polish the highlight context menu and highlight selection from toolbar.

dcaliste:
@pvuorela, ok, I think the PR is ready for review when you have time. It implements the following according to design (or what I've understand from design paper, which hopefully corresponds !):
* long press on an empty area open a context menu to create a new text note, opening a dialog.
* an icon in the toolbar is dedicated for text annotations. When tapped, becomes active, keep the toolbar open, display once a helping notification and waits for tap anywhere to create a new note.
* tapping on a note open a page to view it and edit it on tap. It also has a pull down menu to delete the note.
* long tap on a text note opens a context menu to edit or delete the note.
* selecting a text opens the toolbar and an icon can be tapped to highlight the selected text.
* the same icon can be tapped to become active. When active, moving the finger on text select it and when released, the text becomes highlighted. A helping notification is displayed once.
* a short or long tap on an highlighted area opens a context menu to allow to change color or remove the highlight.
* document PDF is updated with modification in place.

In addition to these, I would like to propose some additional features. They are implemented, but they can also be easily removed if desired. I've put them to show case what can be done with these features and hopefully convince that they are usefull and don't clutter the interface too much. It includes:
* displaying the author of a text annotation as title of the text annotation page, leaving it empty as design recommends when there is no author.
* display as two lines maximum the text content (*i.e.* comment) of an highlight, when any, at the top of the highlight context menu. So one can easily have a clue of the highlight reason by tapping on the highlight. Tapping on this menu item opens the annotation page so one can view the full text and edit it if necessary. When there is no text content for an highlight, I've added a menu item that proposes to add a comment on  the highlight, opening the annotation dialog.
* I've added the highlight style possibilities (*e.g.* underlines, strike through…) in the highlight context menu under the color choices.

I've taken your first remark into account for visual feedback on large annotations. Now, just the tapped rectangle inside an annotation is highlighted when tapped.

Finally, I've one or two questions: what do you recommend concerning the size of PDFDocumentPage.qml file ? The toolbar and the various context menus take quite a big number of lines. Would you recommend to move them in dedicated qml files, or is it just fine line that ?

pvuorela:
Starting to look quite nice with the new design!

...But did find a bunch of little details which work funny or we could adjust the behavior :)

- Press & hold should allow to move finger down and select menu items
- Selecting "Add note" from menu flashes "Delete" while dialog is transitioning visible
- Many times got menu into state where it thinks I'm editing existing annotation even though where I press&hold there is none
- Deletion remorse timer can be removed
- Note editing dialog should have editor size until keyboard top even if empty -> clicking near "Write a note" should focus the editor.
- Opening an existing note looks actually quite empty. Talked with Martin that we could perhaps have then some page header like "Note".
- Clicking on new note button in toolbar should clear selection. Now it leads into strange state when there's at the same time selection and tap creates a note.
- Need new icons for note button, but I can handle that later.
- If user changes color of a highlight, we could store the color in dconf to be user for new highlights coming after that. If user prefers something else than yellow, doesn't need to keep selecting it.
- Maybe same for underlining style if that's kept
- Could maybe mark selected underlining style with highlight color?
- Create annotation on first page and try dragging down -> page content dims except lighter rectangle around the annotation.
- Ideally highlight should continue over words in a row without bumps/spaces between words. Not sure how well Poppler supports, but at least Evince does that. Roundings on highlight edges are not specified but can be kept. For one detail the press effect is smaller than word's highlight.
- Zoom near annotation and click it -> black rectangle appears instead of context menu.
- Comment in context menu not entirely according to UI conventions. If it's clickable it should be primary color. Also ponder if it needs some header/icon to indicate what it is. Might need some thinking more.

Size of file doesn't matter to me that much.

dcaliste:
Great, thank you for all the interesting notes. I'll address the simple minor (but relevent) details tomorrow or during the week-end and report here when done. But before doing so, some remarks about some of your concerns:
> Press & hold should allow to move finger down and select menu items

I agree, and have noticed this also, but didn't find any solution. Ok, I'll try harder ! Something to do with mouse.accept in the press signal handler of pdflinkarea maybe ?

> Many times got menu into state where it thinks I'm editing existing annotation even though where I press&hold there is none

The wiggle for detection is quite large (Theme.itemSizeMedium / 2), do you notice this behaviour within that range, or far outside ?

> Opening an existing note looks actually quite empty. Talked with Martin that we could perhaps have then some page header like "Note".

I agree, "note" or something if no author, or the author.

> Create annotation on first page and try dragging down -> page content dims except lighter rectangle around the annotation.

The most interesting bug. I've noticed it also, but only when rotating the device after having done an annotation. During rotation, the buggy state is not kept for long so I didn't understand the origin of it. But with your way to trigger it, I've now understood. The annotation, for quick rendering, is done by patches on the big page texture. Patches are thrown away when the page is rerendered. In the patch state, the page is thus the sum of several GSSimpleTextureNode. When the screen is dimed, each textures are rendered with an opacity < 1, thus patches are more visible, because they are the sum of the page behind and themselves. I have no idea how to simply solve this… Random thoughts giving: making holes in the page texture where the patches are, asking to rerender the full page after the patch. But none of these is satisfactory.

> Ideally highlight should continue over words in a row without bumps/spaces between words. Not sure how well Poppler supports, but at least Evince does that.

That's the old selection issue we have already discussed together. Currently, the highlight annotation implementation I've done, relies on the selection rectangles to define the quads used by Poppler for the highlight. But the selection rectangles are currently on words only due to Poppler-Qt limitations. There are routines in Poppler-Glib to obtain continuous selections between words, but we're not using them (not available from Poppler-Qt). The idea is to later on, import the code of Poppler-Glib which is doing plain Poppler to get continuous selection and use this instead of the selection by word of Poppler-Qt. After having continuous selection, the highlight will automatically works out of the box. Ideally, I would like to upstream this to Poppler-Qt, but that's another story.

> If it's clickable it should be primary color. Also ponder if it needs some header/icon to indicate what it is. Might need some thinking more.

The other solution, would be to keep the "add/edit a comment" menu item all the time (with a different label). Like that, we keep the small summary on top of the menu, close to the highlight (a bit like the URL when cliking on links), giving a quick and unobstrusive view on the reason of the highlight, if any.

dcaliste:
> Zoom near annotation and click it -> black rectangle appears instead of context menu.

This is the case for links also. It's due to ContextMenuHook, so I've created a specific pull request to correct this independently from the annotation PR, see #90.

pvuorela:
> > Press & hold should allow to move finger down and select menu items

> I agree, and have noticed this also, but didn't find any solution. Ok, I'll try harder ! Something to do with mouse.accept in the press signal handler of pdflinkarea maybe ?

Context Menu should support when inside a MouseArea. Though it might get tricky here when menu parent is emulated with the hook thing. Won't hold back merging because of this.

> > Many times got menu into state where it thinks I'm editing existing annotation even though where I press&hold there is none

> The wiggle for detection is quite large (Theme.itemSizeMedium / 2), do you notice this behaviour within that range, or far outside ?

Far outside. E.g. press and hold on top of first page to create note, after creating press and hold on the next page.

> >  If it's clickable it should be primary color. Also ponder if it needs some header/icon to indicate what it is. Might need some thinking more.

> The other solution, would be to keep the "add/edit a comment" menu item all the time (with a different label). Like that, we keep the small summary on top of the menu, close to the highlight (a bit like the URL when cliking on links), giving a quick and unobstrusive view on the reason of the highlight, if any.

Yea, option too.

dcaliste:
I've addressed the following remarks:
> Press & hold should allow to move finger down and select menu items

I've added a special commit for this. For it to work, as you said, the PDF.LinkArea object needs to emit the positionChanged and released signals. Then, they are catched by the ContextMenuHook and reemit there for the ContextMenu to work. Special notice: it's not working when the background item is not occupying the full width (like the highlight style) because ContextMenu.qml detect the item only with its y coordinate.

> Selecting "Add note" from menu flashes "Delete" while dialog is transitioning visible

Corrected.

> Deletion remorse timer can be removed

Removed.

> Note editing dialog should have editor size until keyboard top even if empty -> clicking near "Write a note" should focus the editor.

Done like in the note application.

> Opening an existing note looks actually quite empty. Talked with Martin that we could perhaps have then some page header like "Note".

Done like that. "Note" is displayed if no author or the author name if exists.

> Clicking on new note button in toolbar should clear selection. Now it leads into strange state when there's at the same time selection and tap creates a note.

Now any action on toolbar clear the selection.

> If user changes color of a highlight, we could store the color in dconf to be user for new highlights coming after that. If user prefers something else than yellow, doesn't need to keep selecting it.

Added a dconf entry and reusing it.

> Maybe same for underlining style if that's kept

Same for style.

> Could maybe mark selected underlining style with highlight color?

Corrected.

> Comment in context menu not entirely according to UI conventions. If it's clickable it should be primary color. Also ponder if it needs some header/icon to indicate what it is. Might need some thinking more.

Corrected by enabling the add / edit menu entry all the time and removing the background item under the info label.

This item is still pending:

> Many times got menu into state where it thinks I'm editing existing annotation even though where I press&hold there is none

Sorry, even with your comment before, I cannot reproduce this behaviour. Is it only on specific PDF documents or on all ?

I think I've otherwise addressed all the remarks you've done. I will continue to test this to check the various git-rebase didn't introduce any regressions and see if there are some unseen corner cases.

pdf/pdfannotation.cpp
```diff
@@ -0,0 +1,353 @@
...
+    Poppler::TextAnnotation *anno = static_cast<Poppler::TextAnnotation*>(d->m_annotation);
+    return (PDFTextAnnotation::IconType)iconNames.indexOf(anno->textIcon());
+}
+
+void PDFTextAnnotation::setIcon(PDFTextAnnotation::IconType value)
+{
+    if (icon() == value)
+        return;
+
+    int index = (int)value;
```
pvuorela: Cosmetics but generally c++ style casts preferred. static_cast<int>(value). Few other instances elsewhere too.

pdf/pdfannotation.cpp
```diff
@@ -0,0 +1,353 @@
...
+    if (index > iconNames.length() || index < 0) {
+        qWarning() << QString::fromLocal8Bit("Wrong icon for text annotation (%1).").arg(index);
```
pvuorela: QStringLiteral could be better here and elsewhere.

pdf/pdfannotation.h
```diff
@@ -0,0 +1,189 @@
...
+    void attached();
+    void authorChanged();
+    void contentsChanged();
+    void creationDateChanged();
+    void modificationDateChanged();
+    void colorChanged();
+
+private:
+    class Private;
+    Private *d;
```
pvuorela: No change needs, but just to note that private data classes can be just omitted since it doesn't matter on qml interfaces nor inside one application.

pdf/pdfdocument.cpp
```diff
@@ -65,6 +68,11 @@ QString PDFDocument::source() const
     return d->source;
 }
 
+QString PDFDocument::autoSave() const
```
pvuorela: Suggestion: autoSaveName wouldn't sound as an action.

plugin/PDFAnnotationEdit.qml
```diff
@@ -0,0 +1,71 @@
...
+                width: parent.width
+                height: Math.max(flickable.height - pageHeader.height, implicitHeight)
+                background: null
+                focus: false
+                text: annotation ? annotation.contents : ""
+                //% "Write a note…"
+                placeholderText: qsTrId("sailfish-office-ta-text-annotation")
+                onTextChanged: {
+                    if (annotation) {
+                        annotation.contents = text
```
pvuorela: Cosmetics: trailing spaces :)

dcaliste:
I've rebased all commits on master.

pdf/pdfannotation.cpp
dcaliste: That's done here and in pdfrenderthraed.cpp also.

plugin/PDFView.qml
```diff
@@ -214,17 +220,32 @@ SilicaFlickable {
                 }
 
                 canvas: pdfCanvas
+                selection: pdfSelection
+
+                onPositionChanged: {
+                    console.log(at)
```
pvuorela: Looks like a leftover.

pdf/pdfannotation.h
dcaliste: Ok, I take note. I think you understand now that I'm learning C++ with this project ;)

pvuorela: Sure :) The reason why it's usually used is keeping binary compatibility, added/removed member variables change size of a class instance. But since this is not a library external software links against, it doesn't provide much.

plugin/PDFSelectionView.qml
```diff
@@ -23,6 +23,7 @@ Repeater {
     id: root
 
     property Item flickable
+    property bool dragable: true
```
pvuorela: draggable

pdf/pdfdocument.cpp
dcaliste: I've changed it for ```autoSavePath```, because it receives something starting with file:///.

plugin/PDFAnnotationEdit.qml
dcaliste: Removed.

plugin/PDFAnnotationEdit.qml
```diff
@@ -0,0 +1,71 @@
...
+                       : qsTrId("sailfish-office-hd-text-annotation")
+            }
+            TextArea {
+                id: areaContents
+                width: parent.width
+                height: Math.max(flickable.height - pageHeader.height, implicitHeight)
+                background: null
+                focus: false
+                text: annotation ? annotation.contents : ""
+                //% "Write a note…"
```
pvuorela: For highlight this and header should probably say "comment", action is "Add a comment".

plugin/PDFDocumentPage.qml
```diff
@@ -108,30 +109,105 @@ DocumentPage {
...
+                default:
+            }
+        }
+        onLongPress: {
+            base.open = false
+            contextMenuText.at = pressAt
+            hook.showMenu(contextMenuText)
+        }
+        clip: anchors.bottomMargin > 0 || base.status !== PageStatus.Active
+        RemorsePopup {
```
pvuorela: Looks unused.

plugin/PDFDocumentPage.qml
```diff
@@ -189,9 +375,182 @@ DocumentPage {
...
+                }
+            }
+        }
+        Row {
+            height: Theme.itemSizeExtraSmall
+            Repeater {
+                id: colors
+                model: ["#db431c", "#ffff00", "#8afa72", "#00ffff",
+                "#3828f9", "#a328c7", "#ffffff", "#989898",
+                "#000000"]
```
pvuorela: Cosmetics, but line splitting with same indentation as property declaration looks strange. Ideally would match with first item in list.

plugin/PDFView.qml
dcaliste: Indeed !

plugin/PDFSelectionView.qml
dcaliste: Oups, sorry, should have paid better attention to English at school…

plugin/PDFDocumentPage.qml
dcaliste: That's true, another leftover, so removed.

plugin/PDFDocumentPage.qml
dcaliste: Corrected. Normally, I do like that, but maybe after several code moves Emacs decided differently and I didn't correct it afterwards.

plugin/PDFDocumentPage.qml
```diff
@@ -108,30 +109,105 @@ DocumentPage {
...
+            base.open = false
+            switch (annotation.type) {
+                case PDF.Annotation.Highlight:
+                contextMenuHighlight.annotation = annotation
+                hook.showMenu(contextMenuHighlight)
+                break
+                case PDF.Annotation.Caret:
+                case PDF.Annotation.Text:
+                contextMenuText.annotation = annotation
+                hook.showMenu(contextMenuText)
```
pvuorela: Cases and actual code shouldn't start on same point. Qt convention would be moving case to match with switch.

pdf/sailfishofficepdfplugin.cpp
```diff
@@ -35,4 +36,8 @@ void SailfishOfficePDFPlugin::registerTypes(const char *uri)
     qmlRegisterType<PDFCanvas>(uri, 1, 0, "Canvas");
     qmlRegisterType<PDFLinkArea>(uri, 1, 0, "LinkArea");
     qmlRegisterType<PDFSelection>(uri, 1, 0, "Selection");
+    qmlRegisterType<PDFAnnotation>(uri, 1, 1, "Annotation");
+    qmlRegisterType<PDFTextAnnotation>(uri, 1, 1, "TextAnnotation");
+    qmlRegisterType<PDFCaretAnnotation>(uri, 1, 1, "CaretAnnotation");
+    qmlRegisterType<PDFHighlightAnnotation>(uri, 1, 1, "HighlightAnnotation");
```
pvuorela: I think I'd just keep 1.0 for simplicity. Shouldn't be imported by anyone except this app. But doesn't really matter.

plugin/PDFAnnotationEdit.qml
dcaliste: That's logical. I've changed the title and the placeholder to be refering to "note" when dealing with Text or Caret annotations and refering to "comment" for any other annotations.

pdf/pdfannotation.h
dcaliste: Thanks for the explanation. It makes sense.

plugin/PDFDocumentPage.qml
dcaliste: Realigned at the three places in PDFDocumentPage.qml.

plugin/PDFDocumentPage.qml
```diff
@@ -189,9 +375,182 @@ DocumentPage {
...
+            text: qsTrId("sailfish-office-me-pdf-txt-anno-clear")
+            onClicked: contextMenuText.annotation.remove()
+        }
+        Component {
+            id: textComponent
+            PDF.TextAnnotation { }
+        }
+    }
+
+    ContextMenu {
```
pvuorela: Can do later, but having this created dynamically when needed would save a bit memory and cpu time on the common case when annotations are not used. Additionally in separate file would save even more as less qml parsing would be needed then.

plugin/PDFDocumentPage.qml
```diff
@@ -189,9 +375,182 @@ DocumentPage {
...
+            }
+        }
+        Row {
+            height: Theme.itemSizeExtraSmall
+            Repeater {
+                id: styles
+                model: [{"style": PDF.HighlightAnnotation.Highlight,
+                         "label": "abc"},
+                        {"style": PDF.HighlightAnnotation.Squiggly,
+                         "label": "a&#816;b&#816;c&#816;"},
```
pvuorela: Hmh, wonder if &#x330; would be easier to check what the character is, unicode more commonly presented in hex.

pdf/sailfishofficepdfplugin.cpp
dcaliste: Ok, I think you're right. Since I was adding new QML object in the library, I increased the version number, but SailfishOffice being the only user of this library, it makes no sense to bother with the versioning. Switched back to 1.0.

pdf/pdflinkarea.cpp
```diff
@@ -148,8 +192,50 @@ void PDFLinkArea::mouseReleaseEvent(QMouseEvent *event)
...
+    // - click.
+
     if (!d->canvas) {
         emit clicked(d->clickLocation);
+    } else if (d->clickOnSelection) {
+        emit selectionClicked();
+    } else if (d->selection && d->selection->count() > 0) {
+        d->selection->unselect();
+    } else if (d->annotation != nullptr) {
+        // proxy will be child of this object to avoid memory leak.
```
pvuorela: Means also that annotation proxies will be accumulated as long as the document is open, right? Wonder would this work better if ownership is passed to qml engine (QQmlEngine::setObjectOwnership() etc). Didn't test though.

pdf/pdfannotation.cpp
```diff
@@ -0,0 +1,353 @@
...
+    if (d->m_document != nullptr && d->m_page >= 0)
+        d->m_document->onPageChanged(d->m_page, d->m_annotation->boundary());
+
+    if (d->m_document != nullptr)
+        d->m_document->documentModified();
+}
+
+PDFAnnotation::SubType PDFAnnotation::type() const
+{
+    return (PDFAnnotation::SubType)d->m_annotation->subType();
```
pvuorela: Cast style.

pvuorela: Also icon() and style() later.

pdf/pdfannotation.cpp
```diff
@@ -0,0 +1,353 @@
...
+    QPair<int, QRectF> word = selection->rectAt(0);
+    attachAt(document, word.first, word.second.x(), word.second.y());
+}
+
+void PDFTextAnnotation::attachAt(PDFDocument *document,
+                                 unsigned int page, qreal x, qreal y)
+{
+    attachOnce(document, page);
+
+    d->m_annotation->setBoundary(QRectF(x, y, 24., 24.));
```
pvuorela: Magic value.

pdf/pdfannotation.h
```diff
@@ -0,0 +1,189 @@
...
+    Q_PROPERTY(int page READ page NOTIFY attached)
+    Q_PROPERTY(QRectF boundary READ boundary NOTIFY attached)
+
+    Q_PROPERTY(QString author READ author WRITE setAuthor NOTIFY authorChanged)
+    Q_PROPERTY(QString contents READ contents WRITE setContents NOTIFY contentsChanged)
+    Q_PROPERTY(QDateTime creationDate READ creationDate CONSTANT)
+    Q_PROPERTY(QDateTime modificationDate READ modificationDate NOTIFY modificationDateChanged)
+    Q_PROPERTY(QColor color READ color WRITE setColor NOTIFY colorChanged)
+    Q_PROPERTY(SubType type READ type CONSTANT)
+
```
pvuorela: Some unused properties here. Could just remove? Also iconType on text annotation.

pdf/pdflinkarea.cpp
pvuorela: May or may not work, but some "activeAnnotation" property might be an option too.

plugin/PDFDocumentPage.qml
pvuorela: Ehm, was supposed to be 816 -> x330

plugin/PDFDocumentPage.qml
```diff
@@ -108,30 +109,102 @@ DocumentPage {
...
+                     || contextMenuText.active
         autoShowHide: !row.active
 
+        Connections {
+            target: view.selection
+            onSelectedChanged: if (view.selection.selected) toolbar.show()
+        }
+
+        Notification {
+            id: notice
```
pvuorela: Also a potential candidate for lazy creation.

plugin/PDFDocumentPage.qml
```diff
@@ -144,6 +217,113 @@ DocumentPage {
...
+                    }
+                }
+                IconButton {
+                    id: highlightButton
+                    anchors.centerIn: parent
+                    highlighted: pressed || highlightTool.pressed || row.activeItem === highlightTool
+                    icon.source: "image://theme/icon-m-edit"
+                    onClicked: highlightTool.clicked(mouse)
+                }
+                Component {
```
pvuorela: Suggest moving this as first or last child element, now between the directly instantiated ones.

pdf/pdfselection.cpp
```diff
@@ -570,6 +572,34 @@ void PDFSelection::onLayoutChanged()
                      QVector<int>{Rect});
 }
 
+bool PDFSelection::selectionAtPoint(const QPointF &point) const
+{
+    if (d->pageIndexStart < 0 || d->pageIndexStop < 0 ||
+        d->boxIndexStart < 0 || d->boxIndexStop < 0)
+        return false;
```
pvuorela: Should have curly braces if conditions wrap multiple lines. || at start of line.

pdf/pdflinkarea.cpp
dcaliste: That's right. At the time I wrote this, I read extensively the Qt documentation and I remember reading that transfer ownership to QML was impossible: http://doc.qt.io/qt-5/qtqml-cppintegration-data.html <quote>When data is transferred from C++ to QML, the ownership of the data always remains with C++.</quote>

So yes, all the annnotation wrappers, generated when tapping on an annotation, will stay alive for the duration of the document. I don't think that's really a memory penalty, even on small memory systems. Indeed, the wrapper is only sizeof(QObject) + some pointers and one int and one boolean. So we're speaking of some kilobytes at most even if the user is tapping on dozen of annotations. Don't you think so ?

pdf/pdfannotation.cpp
dcaliste: Yes, I've explained this in one of the issue #80 comments. It is the size Poppler gives to the text annotation icon that is added in the document. It's also hard-coded in Poppler, and I didn't find any mean to retrieve this value.

I'll add a comment in our code about this.

pdf/pdfannotation.h
dcaliste: You're right. Initially, I've bound as much of Poppler-Qt Annotation object as possible from the C++ side, because it was 1:1 code. Then, the UI part is finally not using all capabilities exposed from C++. I didn't have envy to just erase all this code, even if not use, in case it will be later.

From code development point of view, it's dead code, indeed. I will remove it if you think it's better like that (*i.e.* less maintenance burden).

plugin/PDFDocumentPage.qml
dcaliste: That's true.

plugin/PDFDocumentPage.qml
dcaliste: Ok, moved it first because it's called from the function that is on top of this BackgroundItem.

dcaliste:
Ok, hopefully, I should have addressed all your remarks (thanks a lot for your full afternoon review !), except these ones:
* annotation wrappers piling up on C++ side for the full life duration of the document object (I can test the currentAnnotation property solution if you think it's worth).
* unused properties in pdfannotation.h (wait for your decision to remove them or not).
* add a commit to dynamically create the contextmenus and notifications (need to further work on this).

dcaliste:
As a side note, the unused properties concern this (for you to have a better view of what could be valuable in futur evolutions):
* in PDFAnnotation, creation and modification dates, may be displayed in the PDFAnnotationEdit.qml page;
* in PDFTextAnnotation, the icon, may be used to display a different icon instead of the note one (can be a question mark, a help icon…).

If I'm not doing any mistake, all other properties are currently used here and there.

pdf/pdflinkarea.cpp
pvuorela: Yea, not big objects so could perhaps tolerate, but should have some remark about that at least.

plugin/PDFAnnotationNew.qml
```diff
@@ -0,0 +1,53 @@
...
+                acceptText: qsTrId("sailfish-office-he-txt-anno-save")
+                //% "Cancel"
+                cancelText: qsTrId("sailfish-office-he-txt-anno-cancel")
+            }
+            TextArea {
+                id: areaContents
+                width: parent.width
+                height: Math.max(flickable.height - dialogHeader.height, implicitHeight)
+                //% "Write a note…"
+                placeholderText: qsTrId("sailfish-office-ta-text-annotation")
```
pvuorela: Comment text version here too.

plugin/PDFView.qml
```diff
@@ -214,17 +220,29 @@ SilicaFlickable {
...
+                onSelectionClicked: base.selectionClicked(selection, contextHook)
+                onAnnotationClicked: base.annotationClicked(annotation, contextHook)
+                onTextAnnotationClicked: base.annotationClicked(annotation, contextHook)
+                onCaretAnnotationClicked: base.annotationClicked(annotation, contextHook)
+                onHighlightAnnotationClicked: base.annotationClicked(annotation, contextHook)
+                onClicked: base.clicked()
+                onAnnotationLongPress: base.annotationLongPress(annotation, contextHook)
+                onTextAnnotationLongPress: base.annotationLongPress(annotation, contextHook)
+                onCaretAnnotationLongPress: base.annotationLongPress(annotation, contextHook)
+                onHighlightAnnotationLongPress: base.annotationLongPress(annotation, contextHook)
```
pvuorela: Why different signals for different types of annotations if they all end up handled the same and annotation itself can also identify the type?

dcaliste: I've tested the following, but it didn't work. Maybe you can spot a mistake somewhere.
pdflinkarea.h:
signal annotationPress(PDFAnnotation *anno);

pdflinkarea.cpp:
        switch (d->annotation->subType()) {
        case Poppler::Annotation::SubType::AText: {
            PDFTextAnnotation *proxy = new PDFTextAnnotation(…);
            emit annotationPress(static_cast<PDFAnnotation*>(proxy));
            return;
        }
…

PDFDocumentPage.qml:
onAnnotationPress: if (anno.type == PDF.Annotation.Text) console.log(anno.icon)

This raises a runtime error, saying that icon is not a property of type PDFAnnotation. So what I've deduced is that, having a signal with signature (Annotation *anno), make the QML part loose the information that this anno object is indeed a specialized text version of annotation. Thus, I decided to create as many signals with detailled signature as I have of specialized types, then I store all these different QML object into a single variant QML property, so the type is preserved even if accessed from the same name.

plugin/PDFAnnotationNew.qml
dcaliste: That's right, I thought this one was used only for text annotations. It's corrected.

plugin/PDFView.qml
pvuorela: Changed myself textAnnotationClicked to annotationClicked and works for me, click handler outputs icon type. Casting unnecessary, but don't think it should matter.

dcaliste: Ok, I will retry after lunch to use the unified signal emission from pdflinkarea and report if I stumble on the same issue as before.

pdf/pdfannotation.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+void PDFAnnotation::remove()
+{
+    if (d->m_document == nullptr)
+        return;
+
+    d->m_document->removeAnnotation(d->m_annotation, d->m_page);
+    d->m_annotation = nullptr;
+
+    deleteLater();
```
pvuorela: Since invokable, could consider checking m_annotation null pointer and always deleting.

pdf/pdfannotation.cpp
```diff
@@ -0,0 +1,356 @@
...
+        return;
+
+    d->m_document = document;
+    d->m_page = page;
+}
+
+void PDFAnnotation::attach(PDFDocument *document, PDFSelection *selection)
+{
+    if (!selection)
+        return;
```
pvuorela: For robustness: || selection->count() == 0?

pdf/pdfdocument.cpp
```diff
@@ -144,6 +157,48 @@ void PDFDocument::setSource(const QString &source)
...
+    return d->thread->annotations(page);
+}
+
+void PDFDocument::removeAnnotation(Poppler::Annotation *annotation, int pageIndex)
+{
+    d->thread->removeAnnotation(annotation, pageIndex);
+    documentModified();
+}
+
+void PDFDocument::documentModified()
```
pvuorela: Suggestion: setDocumentModified(), modified alone sounds like a signal.

pdf/pdfannotation.cpp
```diff
@@ -0,0 +1,356 @@
...
+    if (color() == value)
+        return;
+
+    Poppler::Annotation::Style style = d->m_annotation->style();
+    style.setColor(value);
+    d->m_annotation->setStyle(style);
+    emit colorChanged();
+
+    if (d->m_document != nullptr && d->m_page >= 0)
+        d->m_document->onPageChanged(d->m_page, d->m_annotation->boundary());
```
pvuorela: Cannot modify page without touching the document it belongs to, right? Could then make this method trigger modification handling too and remove all documentModified() calls that are done afterwards.

pdf/pdfcanvas.cpp
```diff
@@ -350,6 +376,57 @@ QPair<QUrl, PDFCanvas::ReducedBox> PDFCanvas::urlAtPoint(const QPointF &point) c
     return QPair<QUrl, PDFCanvas::ReducedBox>();
 }
 
+QPair<Poppler::Annotation*, PDFCanvas::ReducedBox> PDFCanvas::annotationAtPoint(const QPointF &point) const
+{
+    for (int i = 0; i < d->pageCount; ++i) {
+        const PDFPage &page = d->pages.value(i);
```
pvuorela: Guess in reality doesn't matter that much, but a bit concerned that now on clicks the document pages are traversed for both urls and links. Can do later, but should consider how to make page lookup and files with ridiculous amount of pages in general more efficient.

pdf/pdfdocument.cpp
```diff
@@ -162,6 +217,16 @@ void PDFDocument::requestPage(int index, int size, QQuickWindow *window, QRect s
     d->thread->queueJob(job);
 }
 
+void PDFDocument::requestArea(int index, int size, QQuickWindow *window, QRect subpart)
+{
+    if (!isLoaded() || isLocked() || subpart.isEmpty())
+        return;
+
+    RenderPageJob* job = new RenderPageJob(index, size, window, subpart,
+                                           RenderPageJob::RenderPageType::Area);
```
pvuorela: This seems a bit silly that there are separate methods for requesting arbitrary areas of pages that are considered root texture and arbitrary areas that are considered patches. Could be nice if patch approach would be implementation detail of pdfcanvas. E.g. add some int extraData to requestPage and return that on finish signal.

dcaliste:
While reviewing your remarks, I noticed that I introduced a regression on the last commit, *i.e.* the tap for long press and move is not working anymore and code to allow this disappear in rebase. I'll try to recover from home backup this evening. So currently don't pay attention to this not working anymore. Sorry for the inconvenience.

pvuorela:
Hmh, now press&hold and then move finger on top of menu doesn't allow to select "add note". Think it worked yesterday at some point? On existing annotation there seems to be some offset. Was earlier seeing also that moving finger on top of color or highlight type kept below item selected.

The case of creating patch on first page and dragging against document start remains. But yeah, don't have good ideas at the moment what to do. Could at least document the shortcoming as some fixme.

plugin/PDFView.qml
dcaliste: Indeed, it's working. So I've removed all the differentiated signals from pdflinkarea.h. I don't remember what specific code implementation I did that triggered the above issue.

pdf/pdfdocument.cpp
```diff
@@ -37,7 +37,9 @@ class PDFDocument::Private
     PDFSearchModel *searchModel;
 
     QString source;
+    QString autoSavePath;
     bool completed;
+    bool modified;
```
pvuorela: Missing initialization?

pdf/pdflinkarea.cpp
```diff
@@ -30,14 +31,18 @@ class PDFLinkArea::Private
...
     QPointF clickLocation;
     int wiggleFactor;
 
     QTimer pressTimer;
     bool pressed;
     PDFCanvas::ReducedBox pressedBox;
+
     QUrl link;
+    Poppler::Annotation *annotation;
+    bool clickOnSelection;
```
pvuorela: Better initialize these too on Private class ctor.

pdf/pdfannotation.cpp
dcaliste: Very true, added.

pdf/pdfannotation.cpp
dcaliste: Good reasoning. I've simplified the implementation by calling the setDocumentModified() in the onPageModified() slot, so setDocumentModified() is not needed anymore in pdfrenderthread when adding or removing an annotation and in the three places where modification on annotation modify the page in pdfannotation.cpp.

pdf/pdfcanvas.cpp
dcaliste: Yes, I guess a dichotomic algorithm should switch from O(N) to O(ln(N)) since the rectangles of pages are sorted (and almost all of the same size). I mean, int(at.y / height) should give a good guess of page index i0, then trying i0 + int((at.y - i0.y)/(height-i0.height)…

pdf/pdfdocument.cpp
dcaliste: true, corrected.

pdf/pdflinkarea.cpp
dcaliste: Ok, the selection attribute was missing initialisation also.

pdf/pdflinkarea.cpp
dcaliste: I've added the following comment:
>  Todo: transfer ownership to QML, if possible. Before that,
     all created proxy objects will be alive up to the moment
     the document object is released.

pdf/pdfdocument.cpp
dcaliste: I did it like that to put more sense into the API, even if a bit more clumpsy. In fact the rendering job is already unified with an enum to decide between Page and Area. The best would have been to expose this enum in pdfcanvas which does the request and pdfjob that stores it, but pdfjob is not visible from pdfcanvas which only address pdfdocument.

What you suggest would have the following API (correct me if I misinterpret your suggestion):
pdfcanvas.cpp:
enum {Page, Area};
m_document->requestPage(…, Page) or m_document->requestPage(…, Area)

pdfdocument.h:
requestPage(…, int flag);

pdfjob.h:
RenderPageJob(…, int flag);

IMHO, when reading this API, the pro is that its simple, to request a page, call requestPage() and details are implementation specific hidden in flag argument. The con is that flag has no mean when not explained from pdfcanvas. What do you prefer ?

pdf/pdflinkarea.cpp
```diff
@@ -148,8 +196,53 @@ void PDFLinkArea::mouseReleaseEvent(QMouseEvent *event)
...
+                (static_cast<Poppler::HighlightAnnotation*>(d->annotation),
+                 d->canvas->document(), d->pressedBox.first, this);
+            emit annotationClicked(proxy);
+            return;
+        }
+        default: {
+            PDFAnnotation *proxy = new PDFAnnotation
+                (d->annotation, d->canvas->document(), d->pressedBox.first, this);
+            emit annotationClicked(proxy);
+            return;
```
pvuorela: Just a minor thing, but all cases emit same signal, can as well do that outside the switch.

pdf/pdfdocument.cpp
dcaliste: Ok, let's go for the ```, gpointer user_data)``` that I know well ! Joke aside, I suppose in your response that you mean « pdfjob doesn't need to know meaning if it only provides a method for storing random data », instead of « pdfcanvas…»

I'll switch for an API like suggested before this evening, thank you for your answer.

pdf/pdfannotation.cpp
```diff
@@ -0,0 +1,346 @@
...
+        bounds.united(selection->rectAt(i).second);
+
+    d->m_annotation->setBoundary(bounds);
+    d->m_document->addAnnotation(d->m_annotation, d->m_page);
+    emit attached();
+}
+
+void PDFAnnotation::remove()
+{
+    if (d->m_document  && d->m_annotation) {
```
dcaliste: Oops, deleted.

pdf/pdfannotation.h
```diff
@@ -0,0 +1,189 @@
...
+private:
+    class Private;
+    Private *d;
+
+    void attachOnce(PDFDocument *document, int page);
+
+    friend class PDFTextAnnotation;
+    friend class PDFCaretAnnotation;
+    friend class PDFHighlightAnnotation;
```
pvuorela: Also option to use protected instead of private.

pdf/pdflinkarea.cpp
```diff
@@ -132,6 +180,36 @@ void PDFLinkArea::mouseMoveEvent(QMouseEvent *event)
     }
 }
 
+PDFAnnotation* PDFLinkArea::newProxyForAnnotation()
+{
+    PDFAnnotation *proxy;
```
pvuorela: Unused.

pdf/pdfselection.cpp
```diff
@@ -570,6 +572,35 @@ void PDFSelection::onLayoutChanged()
...
+    }
+
+    if (!d->canvas)
+        return false;
+    PDFDocument *doc = d->canvas->document();
+    if (!doc)
+        return false;
+    
+    int i;
```
pvuorela: No need to have outside for loop?

dcaliste:
Ok, back at home, I found my backup and restored the commit allowing to continuously move the finger to select item in the context menu.

Besides, I've also found the root of the issue you reported that long press anywhere seems to believe that an annotation was pressed. To reproduce:
* first tap on a note and discard the edit page ;
* long tap anywhere, the context menu was kept showing at the right place but for the previously tapped note.
It's because the annotation property of the context menu was not reseted.

I'll do the protected declarations and the change in requestPage API later this evening.

pdf/pdfannotation.h
dcaliste: Change from friend to protected to allow inheritage.

pdf/pdfdocument.cpp
dcaliste: Ok, done.

dcaliste:
I've added an additional commit where the notifications and the context menus are loaded dynamically on demand. Context menus have also been moved to separated files.

dcaliste:
Last remark: I noticed also that the highlight of the menu items in the highlight context menu is not working well when moving the finger above the color line or the style line. Indeed, it comes from the code in Silica/ContextMenu.qml. There, the highlight logic is done by looking for a child of the context menu at a given y position *and* that has a "__silica_menu_item" property. There is an asumption there that a child menu item should always be found, so the there is no highlight removing in case none is found.

I can cheat a bit by adding a dummy "__silica_menu_item" property on the color line and style line. But even like that, for the style line, the code in ContextMenu.qml is missing the x information, so we cannot select a given style inside the line.

plugin/PDFContextMenuHighlight.qml
```diff
@@ -0,0 +1,137 @@
...
+                    contextMenuHighlight.annotation.style = modelData["style"]
+                    highlightStyleConfig.value = highlightStyleConfig.fromEnum(modelData["style"])
+                }
+                Label {
+                    anchors.centerIn: parent
+                    text: modelData["label"]
+                    textFormat: Text.RichText
+                    color: bgStyle.highlighted
+                           || (contextMenuHighlight.annotation !== undefined
+                               && contextMenuHighlight.aannotation !== null
```
pvuorela: Now that you mentioned these, property variant is deprecated and should be changed to property var: https://wiki.qt.io/Property-var - besides annotations, there's property variant _menu on ContextMenuHook.

For import, don't think there's anything requiring "as PDF", so just drop it if you want.

plugin/PDFContextMenuHighlight.qml
```diff
@@ -0,0 +1,137 @@
...
+
+ContextMenu {
+    id: contextMenuHighlight
+    property variant annotation
+
+    InfoLabel {
+        id: infoContents
+        visible: infoContents.text != ""
+        width: parent.width
+        anchors.topMargin: Theme.paddingSmall
```
pvuorela: Not top anchored, so don't think this does anything. And cannot be vertically anchored anyway since ContextMenu puts this inside a Column. verticalAlignment: Text.AlignVCenter could do what's intended.

plugin/PDFDocumentPage.qml
```diff
@@ -105,33 +109,138 @@ DocumentPage {
...
         flickable: view
         forceHidden: base.open || pdfDocument.failure || pdfDocument.locked
+                     || (contextMenuLinks && contextMenuLinks.active)
+                     || (contextMenuHighlight && contextMenuHighlight.active)
+                     || (contextMenuText && contextMenuText.active)
         autoShowHide: !row.active
 
+        function noticeHide() {
+            if (!notice) {
+                notice = noticeComponent.createObject(toolbar)
```
pvuorela: If no notification existed, there's nothing to hide?

pvuorela:
> I can cheat a bit by adding a dummy "__silica_menu_item" property on the color line and style line. But even like that, for the style line, the code in ContextMenu.qml is missing the x information, so we cannot select a given style inside the line.

Yea, not necessarily important to be able to select color or style by dragging, highlight not very visible anyway with such small items, fitting better for full width ones. Not selecting wrong items would be good enough, but even that is quite minor.

pvuorela:
I think this is starting to be ready, but were you still planning on adjusting those "as PDF" / property variant -> var things?

dcaliste:
> I think this is starting to be ready

Great ! I understand that I still have a long way to go before proposing PR without so many tiny errors or mistakes. Thank you again for the long and deep review.

> were you still planning on adjusting those "as PDF" / property variant -> var things?

Sorry, I missed your reply to the question above, so I was waiting. It's updated now. I've switch from variant to var in ContextMenuHook and I've used property Annotation everywhere in the PDFContextMenu*.qml, simplifying the tests also.

pvuorela:
> Great ! I understand that I still have a long way to go before proposing PR without so many tiny errors or mistakes. Thank you again for the long and deep review.

Also difficult to even avoid review changes with this big bunch of code.

> Sorry, I missed your reply to the question above, so I was waiting. It's updated now. I've switch from variant to var in ContextMenuHook and I've used property Annotation everywhere in the PDFContextMenu*.qml, simplifying the tests also.

Hmh, looks still variants to me?

dcaliste:
Arg, made a second time the mistake of rebasing the wrong tree when switching between machines… Corrected now.

pvuorela:
Still found another issue: I have big pdf, 238 pages, highlight annotation on the second page. Changed the color, quickly popped the page, reopened the document -> old color shown. If then reopen again, the color is changed. Suppose writing the new pdf file is still unfinished when I open the document fast. Can also see smaller file sizes if I do change, pop, reopen, pop, etc.

But I think that's corner case enough not to block merging. Great feature here, huge thanks for implementation!

PR #84, opened 2016-07-25
Make search in PDF cancellable (and not using the rendering thread)

dcaliste:
Initial implementation of search capability in PDF document is using the job queue that is created to load and render PDF documents. That was done for reasons of simplicity and reuse of existing code. But it has three main drawbacks:
* it is using the rendering thread, so a long search postpone any new rendering job (*i.e.* scrolling while searching does not update the view);
* it is not easily cancellable without adding cancellation in PDFJob class that has no sense for other purposes of PDFJob (except loading maybe). So searches in long documents can last for minutes;
* it can not without additional modifications report search progress.

The present PR corrects the first two points by introducing a new ```SearchThread``` class that is dedicated for search. Like that any code specific to search (cancel or progress report) can be implemented there without poluting the rendering thread machinery. Nevertheless its code still resides in ```pdfrenderthread.cpp``` since it is there that the ```Poppler::Document``` is, but if you think it's better, I can create a new ```pdfsearchthread.cpp``` file.

The third point, implying to report progress, can be implemeted later by signaling every ten pages or so, that the list ```m_matches``` has been expanded and modify the search model accordingly.

### DIFF
--- a/pdf/pdfrenderthread.cpp
+++ b/pdf/pdfrenderthread.cpp
@@ -43,6 +43,7 @@ public:
     }
     ~SearchThread()
     {
+        requestInterruption();
         wait();
     }
 
@@ -50,8 +51,8 @@ public:
 
     void start(const QString& search, uint startPage = 0)
     {
-        if (isRunning())
-            return;
+        requestInterruption();
+        wait();
 
         m_search = search;
         m_startPage = startPage;
@@ -60,7 +61,7 @@ public:
 
     void run() {
         m_matches.clear();
-        for (int i = 0; i < m_document->numPages(); ++i) {
+        for (int i = 0; i < m_document->numPages() && !isInterruptionRequested(); ++i) {
             int ipage = (m_startPage + i) % m_document->numPages();
             Poppler::Page *page = m_document->page(ipage);
 
@@ -85,7 +86,12 @@ public:
 
             delete page;
         }
+        if (!isInterruptionRequested())
+            emit searchFinished();
     }
+
+signals:
+    void searchFinished();
 
 private:
     Poppler::Document *m_document;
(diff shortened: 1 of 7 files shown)


pdf/pdfrenderthread.cpp
```diff
@@ -33,6 +33,72 @@ class PDFRenderThreadQueue;
...
+    }
+    ~SearchThread()
+    {
+        requestInterruption();
+        wait();
+    }
+
+    QList<QPair<int, QRectF>> m_matches;
+
+    void start(const QString& search, uint startPage = 0)
```
pvuorela: Maybe startSearch() to avoid name used in parent class for different thing?

pdf/pdfrenderthread.cpp
```diff
@@ -33,6 +33,72 @@ class PDFRenderThreadQueue;
...
+            Poppler::Page *page = m_document->page(ipage);
+
+            double sLeft, sTop, sRight, sBottom;
+            float scaleW = 1.f / page->pageSizeF().width();
+            float scaleH = 1.f / page->pageSizeF().height();
+            bool found;
+            found = page->search(m_search, sLeft, sTop, sRight, sBottom,
+                                 Poppler::Page::FromTop,
+                                 Poppler::Page::IgnoreCase);
+            while (found) {
```
pvuorela: Could later consider if it'd make sense to report matches while search is not yet finished.

pvuorela: Ah, you had already mentioned the same thing.

pvuorela:
Just a minor naming suggestion, LGTM.

pdf/pdfrenderthread.cpp
dcaliste: Yes, I would like to add this functionality also, but I'm wondering currently how to simply signal from the search thread that some additional matches are available. Maybe by signaling the length of the m_matches list every ten pages or so and do the beginRowInsert() things in SearchModel from that simple integer information.

Let's see later !

pdf/pdfrenderthread.cpp
dcaliste: Well, I've named it on purpose, because it has the meaning of starting the work thread from the main one like QThread::start() is doing but doing here for the specialised SearchThread class, so with additional arguments.

Besides, if you think it's misleading as your initial remark implies, I'll change it to startSearch() without any trouble ;)

dcaliste:
Nice, thank you.

Issue #83, opened 2016-07-20
Crash with busy indicator


dcaliste:
Steps to reproduce:
* open a PDF document with quite many pages (~25) with text.
* search for any string.
* while the busy indicator is running, tap to open the drawer and go back to the list of documents.
Repeat this procedure up to the moment it stops with a segfault (within four or five attempts in my testing).

I've rooted the trigger of the bug to the busy indicator running while transitioning the page. Indeed, if in PDFDocumentPage.qml, the ```busy``` attribute is set to ```true```, to always run, it reproduces the bug easily within four or five attempts on any document. So it's not related to the search feature, IMHO.

@pvuorela What's your opinion ? Do you agree that it's the busy indicator that trigger the crash, or do you think there is another cause ? Is it a Silica issue ? How to treat it in sailfish-office ? Should we guard the busy indicator running with a ```&& !open``` condition, so it is never running while the page can transition ?

chriadam:
Do you have a backtrace?  I'm about to head home so won't try to repro this until tomorrow at the earliest, but a backtrace would be very helpful.

dcaliste:
Hi @chriadam Thanks for your reply. I've run it in gdb but the best I have is:
<pre>
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x4601e390 (LWP 5456)]
0x4862c4b0 in ?? (Cannot access memory at address 0x1
)
(gdb) bt
#0  0x4862c4b0 in ?? ()
#1  0x404a7246 in QSGOpaqueTextureMaterialShader::updateState(QSGMaterialShader::RenderState const&, QSGMaterial*, QSGMaterial*) () from /usr/lib/libQt5Quick.so.5
#2  0x40499278 in QSGBatchRenderer::Renderer::renderMergedBatch(QSGBatchRenderer::Batch const*) () from /usr/lib/libQt5Quick.so.5
#3  0x4049a54e in QSGBatchRenderer::Renderer::renderBatches() () from /usr/lib/libQt5Quick.so.5
#4  0x0000000e in ?? ()
#5  0x0000000e in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
</pre>
I didn't try to install libraries with debug symbols. If you need further information, please do ask.

pvuorela:
Getting:
[W] qsg_safeguard_texture:284 - Invalid texture accessed: 0xb061e480 
[...]

PR #85, opened 2016-08-08
[office] Postpone texture deletion, avoid crash on quit with busy running.

dcaliste:
Should correct issue #83 . The cullprit was the texture that was delete as soon as the PDFCanvas object was delete while they may be used a bit longer by the render thread.

Seems to work, but should be tested further.

### DIFF
--- a/pdf/pdfcanvas.cpp
+++ b/pdf/pdfcanvas.cpp
@@ -127,7 +127,7 @@ PDFCanvas::~PDFCanvas()
 {
     for (int i = 0; i < d->pageCount; ++i) {
         PDFPage &page = d->pages[i];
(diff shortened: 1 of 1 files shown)


pvuorela:
@sletta or perhaps @martinjones - this the right way to protect the texture or just making it harder to trigger crash?

dcaliste:
If I may add something, I've read in the Qt documentation yesterday, that it's possible to give the texture ownership to the GSNode. IMHO, that would be the right way to go, but it's available only from Qt5.4.

pvuorela:
Noticed that myself too. Qt update coming up, but not there yet.

sletta:
lgtm

PR #86, opened 2016-08-17
Design update of the toolbar

dcaliste:
This pull request should match the new ui design for current capabilities of the toolbar, *i.e* page counter and search. The toolbar content is modified to contain BackgroundItem which width are the page width divided by the number of children in the toolbar. Like that, we can later easily add items for highlight and annotate.

* The first commit modifies the page counter to use this new design and also match the design mockups (replace « / » by « | » and don't use any icon).
* The second commit modifies the search tool by introducing a new Item called SearchBar that is a modified version of SearchField.qml (I kept the header for licences reasons) to add the two prev / next buttons and add an iconized state when in the tool bar. Also handle the number of matches with the label property (not shown in the mockups but present in the text).
* The third commit is a bug correction commit, to avoid the toolbar to continue hiding on action. For instance, when taping on the search icon while the bar is transitioning to hide, the search field was not displayed because the toolbar was finishing its transition regardless of autoShowHide becoming false.

Implementation is working on all orientations.

### DIFF
--- a/plugin/SearchBarItem.qml
+++ b/plugin/SearchBarItem.qml
@@ -42,28 +42,31 @@ import Sailfish.Silica 1.0
 BackgroundItem {
     id: root
 
-    property alias iconized: searchField.iconized
-    property alias modelCount: searchField.modelCount
+    property bool iconized: true
+    property int modelCount: -1
     property real iconizedWidth
+
+    property real _margin: Math.max((iconizedWidth - searchIcon.width) / 2., 0.)
 
     signal requestSearch(string text)
     signal requestPreviousMatch()
     signal requestNextMatch()
     signal requestCancel()
 
-    onClicked: searchField.iconized = false
+    onClicked: iconized = false
+    onIconizedChanged: if (!iconized) searchField.forceActiveFocus()
 
     states: [State {
                  name: "extended"
-                 when: !searchField.iconized
+                 when: !root.iconized
                  PropertyChanges {
-                     target: searchField
-                     _margin: 0.
+                     target: root
+                     _margin: Theme.horizontalPageMargin
                  }
              },
              State {
                  name: "iconized"
-                 when: searchField.iconized
+                 when: root.iconized
                  PropertyChanges {
                      target: root
                      width: iconizedWidth
@@ -78,26 +81,35 @@ BackgroundItem {
     }
     highlighted: down || searchIcon.down
 
+    IconButton {
+        id: searchIcon
+        anchors {
+            left: parent.left
+            leftMargin: _margin
+        }
+        width: icon.width
+        height: parent.height
+        icon.source: "image://theme/icon-m-search"
+        highlighted: down || root.down || searchField.activeFocus
+
+        onClicked: {
+            root.iconized = false
+            searchField.forceActiveFocus()
+        }
+    }
+
     TextField {
         id: searchField
 
-        property bool iconized: true
-        property int modelCount: -1
+        property string _searchText
 
-        property real _prevNextWidth: modelCount > 0 ? (Theme.itemSizeSmall + Theme.paddingMedium)*2 : 0.
-        implicitWidth: !iconized ? _editor.implicitWidth + Theme.paddingSmall
-                                   + Theme.itemSizeSmall*2  // width of two icons
-                                   + _prevNextWidth // width of prev / next icons
-                                 : Theme.itemSizeSmall
         height: Math.max(Theme.itemSizeMedium, _editor.height + Theme.paddingMedium + Theme.paddingSmall)
 
-        property real _margin: Math.max((root.iconizedWidth - Theme.itemSizeSmall) / 2., 0.)
         anchors {
             verticalCenter: parent.verticalCenter
-            left: parent.left
+            left: searchIcon.right
             right: parent.right
-            leftMargin: _margin
-            rightMargin: _margin
+            rightMargin: Theme.horizontalPageMargin
         }
 
         focusOutBehavior: FocusBehavior.ClearPageFocus
@@ -106,31 +118,39 @@ BackgroundItem {
             family: Theme.fontFamilyHeading
         }
 
-        textLeftMargin: Theme.itemSizeSmall + Theme.paddingMedium + Theme.horizontalPageMargin - Theme.paddingLarge
-        textRightMargin: Theme.itemSizeSmall + Theme.paddingMedium + Theme.horizontalPageMargin - Theme.paddingLarge + _prevNextWidth
+        textRightMargin: clearButton.width
+            + (searchPrev.visible ? searchPrev.width + Theme.paddingLarge : 0.)
+            + (searchNext.visible ? searchNext.width + Theme.paddingLarge : 0.)
         textTopMargin: labelVisible ? Theme.paddingSmall : (height/2 - _editor.implicitHeight/2)
-        labelVisible: modelCount > 0 && !searchField._editor.activeFocus
-        Binding on label {
-            target: searchField
-            //% "%n item(s) found"
-            value: qsTrId("sailfish-office-lb-%n-matches", modelCount)
-            when: modelCount > 0
-        }
 
-        onModelCountChanged: if (modelCount == 0) text = "" // Allow the placeholder
-        // to be visible for no results.
-        onIconizedChanged: if (!iconized) _editor.forceActiveFocus()
+        labelVisible: root.modelCount > 0 && !searchField.activeFocus
+        //% "%n item(s) found"
+        label: qsTrId("sailfish-office-lb-%n-matches", root.modelCount)
 
-        //: Placeholder text of SearchField
-        placeholderText: (modelCount == 0)
+        placeholderText: (root.modelCount == 0 && !activeFocus)
             //% "No result"
             ? qsTrId("sailfish-office-search-no-result")
             //% "Search on document"
             : qsTrId("sailfish-office-search-document")
-        onFocusChanged: {
-            if (focus) {
+
+        Connections {
+            target: root
+            onModelCountChanged: if (modelCount == 0) {
+                searchField.text = "" // Allow the placeholder
+                searchField._searchText = ""
+            }
+        }
+
+        onActiveFocusChanged: {
+            if (activeFocus) {
                 cursorPosition = text.length
-                searchField.iconized = false
+            } else {
+                if (text != _searchText) {
+                    text = _searchText
+                }
+                if (!text) {
+                    root.iconized = true
+                }
             }
         }
 
@@ -138,35 +158,27 @@ BackgroundItem {
         EnterKey.iconSource: text != "" ? "image://theme/icon-m-enter-accept"
                                         : "image://theme/icon-m-enter-close"
         EnterKey.onClicked: {
-            focus = false
             if (text != "") {
+                _searchText = text
                 root.requestSearch(text)
             } else {
-                iconized = true
+                root.iconized = true
             }
+            focus = false
         }
 
         background: null
 
+        visible: !root.iconized && opacity > 0.
+
+        opacity: root._margin == Theme.horizontalPageMargin ? 1. : 0.
+        Behavior on opacity { FadeAnimation {} }
+
         Item {
-            parent: searchField // avoid TextBase contentItem auto-parenting
-            anchors.fill: parent
+            parent: searchField
+            anchors.right: parent.right
 
-            IconButton {
-                id: searchIcon
-                x: searchField.textLeftMargin - width - Theme.paddingSmall
-                width: icon.width
-                height: parent.height
-                icon.source: "image://theme/icon-m-search"
-                highlighted: down || root.down || searchField._editor.activeFocus
-
-                enabled: searchField.enabled
-
-                onClicked: {
-                    searchField.iconized = false
-                    searchField._editor.forceActiveFocus()
-                }
-            }
+            height: parent.height
 
             IconButton {
                 id: searchPrev
@@ -178,10 +190,9 @@ BackgroundItem {
                 height: parent.height
                 icon.source: "image://theme/icon-m-left"
 
-                enabled: searchField.enabled
-                visible: !iconized && opacity > 0.
+                visible: opacity > 0.
 
-                opacity: modelCount > 0 && !searchField._editor.activeFocus ? 1 : 0
+                opacity: root.modelCount > 0 && !searchField.activeFocus ? 1 : 0
                 Behavior on opacity { FadeAnimation {} }
 
                 onClicked: root.requestPreviousMatch()
@@ -197,10 +208,9 @@ BackgroundItem {
                 height: parent.height
                 icon.source: "image://theme/icon-m-right"
 
-                enabled: searchField.enabled
-                visible: !iconized && opacity > 0.
+                visible: opacity > 0.
 
-                opacity: modelCount > 0 && !searchField._editor.activeFocus ? 1 : 0
+                opacity: root.modelCount > 0 && !searchField.activeFocus ? 1 : 0
                 Behavior on opacity { FadeAnimation {} }
 
                 onClicked: root.requestNextMatch()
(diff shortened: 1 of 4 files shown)


pvuorela:
Nice!

Talked with design and there are few details:
- on active search the close button should switch to initial state of the toolbar, not refocus search field
- could have a bit more padding between active search buttons on the right
- transition between toolbar basic state and focused search field is quite busy, field moves a lot and on closing the X button get abruptly repositioned. Some of that could be solved if search button would be in main state where it is in design, moving left just a bit.

dcaliste:
Thanks for your quick reply.

> could have a bit more padding between active search buttons on the right

I've put paddingMedium now, instead of paddingSmall, between the three buttons.

> on active search the close button should switch to initial state of the toolbar

That's clearer, I agree. If the search entry has the focus (and text is not empty), the X button acts like a clear button, otherwise, it's a close button. In any case, it cancels any pending search.

> transition between toolbar basic state and focused search field is quite busy

I agree. I've tried to clean this up a bit, by removing the fade animations of the three buttons while closing (the visual transition is handled by the behaviour on width anyway), it avoid the blinking of these buttons. I've moved also the behaviour on width directly on the BackgroundItem which is responsible to handle the size of the search field anyway.

Hopefully it should help to be more in the spirit of the design team.

dcaliste:
By the way, would it be good to alternate the icon used on the clear / close button ? Using something like this:
<pre>icon.source: (!searchField._editor.activeFocus || searchField.text == "")
                         ? "image://theme/icon-m-close"
                         : "image://theme/icon-m-clear"</pre>

The button purpose is clearer like that, I think, but maybe it makes the UI blinking too much. What would you and design advice ?

dcaliste:
I've modified a bit the commits to add a way to prevent the toolbar to automatically hide when pressing a background item. Also highlight the search icon when the background item is pressed.

pvuorela:
> I've put paddingMedium now, instead of paddingSmall, between the three buttons.

Got feedback that it could be even still sparser. Maybe paddingLarge then.

> I agree. I've tried to clean this up a bit, by removing the fade animations of the three buttons while closing (the visual transition is handled by the behaviour on width anyway), it avoid the blinking of these buttons. I've moved also the behaviour on width directly on the BackgroundItem which is responsible to handle the size of the search field anyway.

Lot better, but the icon still moves around quite much. Was suggested that if we instead keep search action as leftmost item. Then the icon can stay in place. Page number then to be rightmost item, annotation actions later on in the middle. Was thinking if toolbar should consist of five or six cells, then icon can stay in one place better in search active. Or just have padding.

> By the way, would it be good to alternate the icon used on the clear / close button ? Using something like this:

The icons don't don't differ and state the actual action that well, so let's not do that.

pvuorela:
Oh and now there's toolbar size transition when changing orientation.

dcaliste:
> Maybe paddingLarge then.

Let's go for paddingLarge !

> Was suggested that if we instead keep search action as leftmost item.

I've tried that, and it looks ok. I share the view that having the page count at the rightmost place is good, particularly if it opens an attached page on the right. I've modified the transition logic also, not to act on width but on margins. Like that there is no animation when screen is rotated and the result while transitioning from iconized to active for the search bar is quite nice IMHO.

> so let's not do that.

Ok, reverted to clear icon always.

pvuorela:
Lot better again! And sorry for delay, missed the update notifications here.

Still nitpicking on some transition details: toolbar open in basic state and rotating the device makes the icon move to its new position (page number directly where it should be). Could try fixing at least this. Maybe separate state + transition only with that instead of Behavior?

Then closing keyboard with enter key briefly shows ... eliding on placeholder text and close icon replacement with page number in different position is a bit abrupt. Could live with or fix later when adding more controls to the toolbar.

Mentioned cell number increase as possibility, but yeah, let's start with only these two. Tested myself adding three placeholders in the middle and at that point could then try not to move icons at all.

dcaliste:
> sorry for delay

No trouble, nothing is in a hurry and you may have other things to do also !

> Still nitpicking on some transition details

Good, we should aim the best result.

> Maybe separate state + transition only with that instead of Behavior?

I didn't see before the difference in purpose of Behaviour and State / Transition, the later seems to be a verbose way to achieve the former. Now, I see the use case, it's clever. So, with states it's indeed better and we can transition only when switching from iconized to extended and vice versa. I've modified a bit further SearchBar to make it inherit of BackgroundItem, so I can play with the centering stuff and width stuff without implying something on the parent of the TextField. And it's cleaner in PDFDocumentPage also now.

> Then closing keyboard with enter key briefly shows ... eliding on placeholder text

That's a bit of an issue because we don't have access to the placeholder item without fiddling with internals of TextBase which is not recommended, I guess. Besides, I've (yet again) modified a bit the transitions to make the width and the margin to be resized together. It should make the placeholder less blink when transitioning between iconized and extended. Give me your feelings.

plugin/ToolBar.qml
```diff
@@ -35,6 +35,7 @@ PanelBackground {
             autoHideTimer.start()
         } else {
             autoHideTimer.stop()
+            _active = true
```
pvuorela: Seems surprising from toolbar point of view, setting autoShowHide as false activates a hidden toolbar. If intention is to allow reactivating while toolbar is transitioning away, maybe check offset > 0?

plugin/SearchBarItem.qml
```diff
@@ -0,0 +1,243 @@
...
+            properties: "_margin, width"
+            easing.type: Easing.InOutQuad
+            duration: 400
+        }
+    }
+    Binding on highlighted {
+        target: root
+        when: searchIcon.down
+        value: true
+    }
```
pvuorela: Why does this need Binding? And isn't target redundant with 'on'?

plugin/SearchBarItem.qml
```diff
@@ -0,0 +1,243 @@
...
+        implicitWidth: !iconized ? _editor.implicitWidth + Theme.paddingSmall
+                                   + Theme.itemSizeSmall*2  // width of two icons
+                                   + _prevNextWidth // width of prev / next icons
+                                 : Theme.itemSizeSmall
+        height: Math.max(Theme.itemSizeMedium, _editor.height + Theme.paddingMedium + Theme.paddingSmall)
+
+        anchors.verticalCenter: parent.verticalCenter
+        anchors.left: parent.left
+        anchors.right: parent.right
+        property real _margin: Math.max((root.iconizedWidth - Theme.itemSizeSmall) / 2., 0.)
```
pvuorela: Nitpick: properties commonly grouped together.

plugin/SearchBarItem.qml
dcaliste: The reason is that both the BackgroundItem and the IconButton (the search icon) have a highlighted property that can be triggered independently. We want the IconButton to be highlighted when the BackgroundItem is down or when itself is down and vice versa. I was thinking that writing something like this would result in a loop:
<code>
BackgroundItem {
  id: root
  highlighted: down || icon.down
  IconButton {
    id: icon
    highlighted: down || root.down
  }
}
</code>
But well, now that I'm writing it (and not like <code>highlighted: down || other.highlighted</code> as firstly tried), You're right that it should be done like that and not with a Binding element.

plugin/ToolBar.qml
dcaliste: You're right, something like ```_active = (offset > 0)``` would allow to change autoShowHide property without changing the visibility of the toolbar: a hidden one stays hidden, and a transiting one (either showing or masking) will stay visible.

plugin/SearchBarItem.qml
```diff
@@ -0,0 +1,243 @@
...
+            } else {
+                iconized = true
+            }
+        }
+
+        background: null
+
+        Item {
+            parent: searchField // avoid TextBase contentItem auto-parenting
+            anchors.fill: parent
```
pvuorela: Is there reason for putting these inside the TextField rather than directly in the parent?

dcaliste: Yes, at first, the idea was to stay as close to the source code of the SearchField item as possible to allow a hypothetical merge in future. The main variations are the introduction of the prev / next buttons and the ability to catch the press callback on the clear button and adapt its behaviour. All the rest (more or less) can be moved outside such a modified SearchField (the use of the label, the iconized property…).

So here, depends where we want to go:
- either we keep the code as close as possible to the SearchField item, which means putting the icons in the TextField and play with text margins.
- or we implement a SearchBarItem made from a TextField and separated IconButtons, which would allow to be able to play easily with the visibility of each elements.

plugin/SearchBarItem.qml
dcaliste: Ok, grouped.

plugin/ToolBar.qml
```diff
@@ -35,6 +35,8 @@ PanelBackground {
             autoHideTimer.start()
         } else {
             autoHideTimer.stop()
+            // Keep a transiting toolbar visible.
+            _active = (offset > 0)
```
pvuorela: Should also start autoHideTimer if autoShowHide is true?

dcaliste: Sorry, I don't understand. I think it's the case with the first part of the ```if``` statement, isn't it ?

When ```autoShowHide``` becomes ```true``` and the toolbar is shown (*i.e.* ```_active``` is also ```true```), the timer is started.

plugin/SearchBarItem.qml
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        onModelCountChanged: if (modelCount == 0) text = "" // Allow the placeholder
+        // to be visible for no results.
+        onIconizedChanged: if (!iconized) _editor.forceActiveFocus()
+
+        //: Placeholder text of SearchField
+        placeholderText: (modelCount == 0)
+            //% "No result"
+            ? qsTrId("sailfish-office-search-no-result")
```
pvuorela: Search with something that's not found and refocus text field -> No result as placeholder text. Would expect "Search on document", no? Text also as "No results" on spec jpg.

plugin/SearchBarItem.qml
```diff
@@ -0,0 +1,241 @@
...
+        id: searchField
+
+        property bool iconized: true
+        property int modelCount: -1
+
+        property real _prevNextWidth: modelCount > 0 ? (Theme.itemSizeSmall + Theme.paddingMedium)*2 : 0.
+        implicitWidth: !iconized ? _editor.implicitWidth + Theme.paddingSmall
+                                   + Theme.itemSizeSmall*2  // width of two icons
+                                   + _prevNextWidth // width of prev / next icons
+                                 : Theme.itemSizeSmall
```
pvuorela: What should this affect? Not referenced and anchors are used for horizontal edges.

Also: search with a string found in document, click on field when search is done -> width is only half the screen.

plugin/SearchBarItem.qml
```diff
@@ -0,0 +1,241 @@
...
+        inputMethodHints: Qt.ImhNoAutoUppercase | Qt.ImhPreferLowercase | Qt.ImhNoPredictiveText
+        EnterKey.iconSource: text != "" ? "image://theme/icon-m-enter-accept"
+                                        : "image://theme/icon-m-enter-close"
+        EnterKey.onClicked: {
+            focus = false
+            if (text != "") {
+                root.requestSearch(text)
+            } else {
+                iconized = true
+            }
```
pvuorela: Click on field, swipe the keyboard down -> placeholder text "Search on document" stays.

plugin/SearchBarItem.qml
pvuorela: I think it could be a bit simpler if buttons would be under BackgroundItem, though doesn't matter that much. 

But did find a few glitches on state handling. Perhaps could be more robust if the single field wouldn't be reused for all of them. Just an idea.

plugin/SearchBarItem.qml
dcaliste: Yes, I've noticed this before also but had no specific opinionon what would be expected ? Search being cancelled and iconized, maybe.

plugin/SearchBarItem.qml
dcaliste: > width is only half the screen.

Ah, ah, fun ! I didn't notice. I'll correct.

plugin/SearchBarItem.qml
pvuorela: Returning to previous state could be logical, if there was basic view then that would be reached. If there was search content available, then that would be reached. 

Currently also if there's matches found with, say, "word", clicking on it and typing more, "wordfoo", swipe down -> wordfoo is said to have matches. More of a minor detail, but resulting of using the single field for all.

plugin/SearchBarItem.qml
dcaliste: Corrected.

plugin/SearchBarItem.qml
dcaliste: Corrected.

plugin/SearchBarItem.qml
dcaliste: Now, the displayed text is always the one of the search. If the text is modified but the keyboard dismissed, the text is changed back to its previous value.

dcaliste:
I've tried to address all wrong behaviours you mentioned previously:
* when taping on "no results", the "search on document" is shown again.
* dismissing the keyboard when no text close the search.
* dismissing the keyboard with text but no search result close the search.
* dismissing the keyboard with text with search result display the three buttons.

I've tried to move all the icon buttons out of the textfield, but didn't succeed because the textfield loose the focus when clicking on the clear icon. So I've moved out only the search icon but left the three others inside. I've put this in a different commit in case you would like to revert.

plugin/SearchBarItem.qml
```diff
@@ -0,0 +1,248 @@
...
+        anchors {
+            left: parent.left
+            leftMargin: _margin
+        }
+        width: icon.width
+        height: parent.height
+        icon.source: "image://theme/icon-m-search"
+        highlighted: down || root.down || searchField.activeFocus
+
+        enabled: root.enabled
```
pvuorela: Enabled is a special property that has by default inherited value from parent, so these lines shouldn't have any effect. Also enabled is not set false anywhere?

dcaliste: It's an old inheritance from SearchField.qml that I didn't removed. I will.

plugin/SearchBarItem.qml
```diff
@@ -0,0 +1,248 @@
...
+            searchField.forceActiveFocus()
+        }
+    }
+
+    TextField {
+        id: searchField
+
+        property string _searchText
+
+        height: Math.max(Theme.itemSizeMedium, _editor.height + Theme.paddingMedium + Theme.paddingSmall)
```
pvuorela: Usually not a good idea to refer to internal properties, but oh well.

pvuorela:
On moving out buttons, it would be possible to still reserve space for them with text margins, but as said, good already. 

Added some minor comments, but on general seems to work and should be good to go. I'll anyhow wait until next week before merging to match our schedules.

plugin/SearchBarItem.qml
dcaliste: Here again, it's a legacy from SearchField.qml, but I didn't find how to center the input text when there is no label without knowing the _editor height.

dcaliste:
> On moving out buttons, it would be possible to still reserve space for them with text margins

Sorry, maybe I don't understand properly your remark. I mean, moving the buttons out creates an issue with focus, not space on the right. If the clear button is out, when clicking on it the textfield loses the focus. I didn't find how to keep the focus on the text field (I try to encapsulate the whole in a FocusScope with no success).

> I'll anyhow wait until next week before merging to match our schedules.

Good, good. That's fine with me. If you have a moment to look at the context menu PR#87 that would be great also and I can then begin to migrate the annotation code to the new design ! But if you're busy, I can wait.

pvuorela:
> Sorry, maybe I don't understand properly your remark. I mean, moving the buttons out creates an issue with focus, not space on the right. If the clear button is out, when clicking on it the textfield loses the focus. I didn't find how to keep the focus on the text field (I try to encapsulate the whole in a FocusScope with no success).

Could be misunderstanding here too. Assumed click to remove focus would only count area outside the editor.

dcaliste:
> Assumed click to remove focus would only count area outside the editor.

Ok, now I got it. That's worth to try. I will.

pvuorela:
I'll merge now. Can adjust the editor details then separately later.

PR #87, opened 2016-08-18
Update context menu with new design

dcaliste:
This PR introduces two commits to follow design guide-lines for context menus, *i.e.* having menu that opens locally near the item they are related to. Currently, context menus are used to display actions when clicked on a PDF link.
* first commit was taken from my annotate branch. It brings the knowledge of the bounding box when clicking on a link. This bounding box is required to position the context menu below the link (before the menu was full screen, so no need at that time of the link position and shape).
* second commit replaces PDFContextMenu with a FloatingMenu implementation which is nothing else than a ContextMenu with a dimming background and a function to position it at a given y value in the flicked item.

This works also when taping on a link while the toolbar is opened. The toolbar is automatically disabled by the context menu.

Besides, I think that your remark on the first commit about strange bounding boxes for highlighted regions padding on two columns can be solved by giving a bounding box, only around the local tapped area (in that example, it would be the tapped line, and not the full highlighted region). But anyway, that will come later when using these FloatingMenu items for highlighted regions.

Finally, a last remark, the current implementation suffers from a bug when the menu should expand below the end of the document (below the end of the current view is fine, the flickable is moved to allow full view of the menu). In that case, since the PDF.canvas item has a fixed height, it cannot scroll further to include the context menu. I have no simple idea how to solve this at the moment.

Anyway implementation for most cases is there and we can discuss it.


pvuorela:
Wasn't explicitly mentioned in the new design what to do with existing links menus. Martin was liking the idea of having them consistent, but both ways are fine. I think it's then up to the little details how well it works. Unfortunately I'm having some problem testing this. Clicking on a link I get black screen and:

[D] QOpenGLFramebufferObjectPrivate::init:482 - [opengl/qopenglframebufferobject.cpp line 482] OpenGL Error: 1281
[D] QOpenGLFramebufferObjectPrivate::checkFramebufferStatus:356 - QOpenGLFramebufferObject: Framebuffer incomplete attachment.
[D] QOpenGLFramebufferObjectPrivate::checkFramebufferStatus:356 - QOpenGLFramebufferObject: Framebuffer incomplete attachment.
[W] QQuickShaderEffectTexture::bind:202 - ShaderEffectSource: 'recursive' must be set to true when rendering recursively.

Last repeated many times. No idea why that happens.

dcaliste:
Ah, annoying. I've built on the lastest public SDK and run it on a 2.0.2.50 installation, I don't have this issue. I have no clue neither where it can come from. There is no strange things in the new code, just a ContextMenu displayed on top of a SilicaFlickable…

Concerning this PR, applying it on PDF links was a way to test the implementation of the FloatingMenu to be used later for the annotation stuff and having this context menu rendering consistent inside the application. But as you said both ways are fine (as long as the last one can run :| ).

dcaliste:
I've found some PDF documents that trigger such a behaviour every time a link is tapped, while some others never make the bug appears. Quite strange, but since I can reproduce it easily, I'm going to investigate.

dcaliste:
Ok, I've found my mistake. I've attached the context menu to the PDF.canvas item itself (of course, the PDFView flickable is made of a long single item and not a list of ListItem like in usual SilicaListView). When the context menu is shown, it applies a dimming effect on all the window and add a ShaderEffect to the parent item of the menu to undo this dimming effect by displaying the original values of the item, see ContextMenu.qml the removeOpacityEffect component.

In my case, the item where the dimming effect should be removed is the full PDF document. So the maximum texture size when doing a layer.effect on it, is easily reached for zoomed-in documents or documents with several pages.

My solution, instead of copying sources of the ContextMenu.qml and removing the dimming effect, is to provide an Item with the size and position of the tapped link (or later annotation) and show a standard context menu on this hooking item. I've called this item ContextMenuHook. It is also responsible to display a rectangle under the menu to help reading on light documents. What do you think ?

The only drawback of this solution is that to avoid the dimming effect, after reading the source of ContextMenu, the only way was to fake a MouseArea in ContextMenuHook that is always pressed. A bit ugly IMHO, but I don't want to duplicate full ContextMenu just to remove the dimming effect.

plugin/ContextMenuHook.qml
```diff
@@ -0,0 +1,59 @@
...
+
+    Rectangle {
+        id: background
+        parent: _menu ? _menu : null
+        anchors.fill: parent ? parent : undefined
+        color: Theme.highlightDimmerColor
+        opacity: 0.91
+        Binding on z {
+            when: parent
+            value: parent.z - 1
```
pvuorela: I think z: -1 should provide the same result as this binding. Negative values stacked behind parent.

pvuorela:
Found some positioning glitches when clicking on link and changing between orientations, depending on where the link was. Also on landscape it's possible to open split view and then click on a link, as result the view is zoomed and menu opens to wrong position. Although hoping that we could soon get rid of the split view, which could fix that kind of issues as side-effect.

Emulating a mouse area for ContextMenu is a bit suspicious, but since this is the first place where we use it as overlay, can accept as proving the concept. Can then later add better support to ContextMenu for living in more different kind of contexts.

plugin/ContextMenuHook.qml
dcaliste: You're right. I've simplified with z: -1.

dcaliste:
Thanks for the review.

> Found some positioning glitches when clicking on link and changing between orientations

Ok, I didn't notice before, it's when rotating make the zoom level change. I'll try to figure out a way to fix this. The issue with the split view has the same cause.

> can accept as proving the concept

Ok, thanks, at least we can test the new design for annotation support also and see if it is indeed convenient. Let's see also if the browser adopts this new design and how they will choose to implement the UI mockups.

dcaliste:
> Found some positioning glitches when clicking on link and changing between orientations

I've modified a bit the implementation, so PDFLinkArea object can signal changes in clickedBox coordinates when the page layout is modified. As a result, this PR now introduces two new properties in PDFLinkArea:
* ```clickedBox``` which is the rectangle in canvas coordinates of the last activated clickable area in the document. These coordinates are still valid after the mouse has been released and are updated on page layout changes.
* ```pressed``` which indicates when the mouse is pressed.

The conjunction of these two properties allow to:
* highlight the pressed area ;
* do something around the tapped area after click has been fired and the finger moved away from screen.

pvuorela:
Not sure how much latest changes fixed glitches, but those seem still easy to trigger. Cases like orientation change not having menu visible, or having menu visible but closing it to suddenly move view elsewhere. Similar with split view. If you want and it helps you, can also disable link click handling on landscape split view.

Small additional case is if url wrapping changes due to orientation change width change. But that should be ContextMenu's transition and can be ignored for now. 

Another detail I noticed is context menu showing in app cover. Selection gets there too. Can ignore for now, but rendering pdf separately for cover could help also on that.

dcaliste:
> Not sure how much latest changes fixed glitches
Well, it did, now the menu is always displayed below the link even when the link is displayed elsewhere. The issue appears when link becomes rendered out of the page when rotating.

Sorry, it appears that I tested the fix only for links near the top of the page, so I didn't notice. I was satisfied because the menu was moved with the link when rotating and I didn't push further. Sorry.

I need to think how to handle this. Currently, the fixed point during relayout or rotation is the top left corner (as defined by contentX and contentY). Maybe it would be interesting to change this to the link centre when a context menu is opened ?

> But that should be ContextMenu's transition

Yes, I don't know much how I can handle this.

> Another detail I noticed is context menu showing in app cover

That may be a complete different PR, that would solve also the issue of the scaling of the cover when rotating.

dcaliste:
In a SilicaListView, having a contextmenu opened and rotate the device always handle these cases properly:
* no context: the top listitem is the fixed point when rotating;
* a context in portrait: the view is scrolled so the menu is always visible, and when rotating back the list is scrolled back to its old position.

I've done a bit of QML code reading this week-end. There is nothing specific in SilicaListView.qml, and I think, correct me if I'm wrong, that the code responsible for this is in ContextMenu.qml. More specifically, I think it comes from the fact that flickable.contentY is bound to an apropriated arithmetic formula when the menu is shown.

Unfortunately, when the device is rotated (or when the split view is closed), the PDFView item calls ```adjust()``` which does the trick on contentX and contentY to keep the top left corner fixed when rotating. I think, this JS affectation removes the binding set from ContextMenu, so the flickable is not properlly positioned later on.

I don't know how ListView (inherited by SilicaListView) is handling the device rotation (or the width change in fact) not to break any binding in contentY. I didn't find any ListView.qml in /usr/lib/qt5/qml. Maybe in the code is in C++ in Qt sources.

Any way, what do you think about calling ```adjust()``` on width changed *only* if interactive is ```true``` ? That may solve the issue, but is it to much of a trick ? Or I'm still missing something and it won't help ?

dcaliste:
It seems that positioning issues come from the fact that ConetxtMenu.qml assumes that ```contentHeight``` of the flickable is constant while the menu is shown. My mistake, but of course, it's true for any SilicaListView, since changing orientation just changes the item widths, not their heights. But, it's not the case for flickables where the height / width ratio should be preserved, like images or PDF documents.

Indeed, ContextMenu.qml assumes constant contentHeight, because, it is saving contentY in _flickableContentYAtOpen to ensure correct placement of the menu, even if opening imply to shift a bit up the flickable to display the menu entirely.

So first action is to update _flickableContentYAtOpen whenever the document is zoomed by rotation changes. But this is not enough. Indeed, a binding on _flickable.contentY is set by PropertyChanges when the menu is open in ContextMenu.qml. Then, this PropertyChanges automatically set back contentY to its initial value when the state is changed to closed. But the initial value may be out of date if a change in the zoom has happened.

So second action is to force contentY to reset to _flickableContentYAtOpen when _open becomes false. One cannot use the closed signal here, because closed() signal is issued before _open is set to false in ContextMenu.qml.

Now, opening a link on the bottom of a page in portrait indeed move the flickable up. Then, rotating the device scroll again to ensure the visibility of the menu and finally, dismissing the menu moves up the view as much as the link is still visible.

I acknowledge that this uses a lot of the internals of ContextMenu (_open, _flickable and _flickableContentYAtOpen) and maybe (maybe) this will be pushed in ContextMenu.qml later if this component is to be reused in contexts where the flickable contentHeight may be allow to change while the menu is opened.

What's your opinion ?

plugin/PDFView.qml
```diff
@@ -218,6 +228,18 @@ SilicaFlickable {
...
+        Rectangle {
+            x: linkArea.clickedBox.x
+            y: linkArea.clickedBox.y
+            width: linkArea.clickedBox.width
+            height: linkArea.clickedBox.height
+            radius: Theme.paddingSmall
+            color: Theme.highlightColor
+            opacity: 0.75
+            visible: linkArea.pressed
+            Behavior on opacity { FadeAnimation { duration: 100 } }
```
pvuorela: Opacity doesn't change anywhere? Maybe something like: opacity: linkArea.pressed ? 0.75 : 0; visible: opacity > 0 ?

pvuorela:
Nice job in figuring out the details of ContextMenu here.  Seems to work quite well now.

Indeed it's not nice using those internal properties, but something to get started with, so ok from me. One small comment, but otherwise LGTM.

One other detail I noticed: if toolbar was visible when clicking a link, it's visible for a while but not interactive and doesn't indicate that in any way. Could consider if dimming that's normally done for full screen should be emulated there, or then just hide immediately. I'll mention to design. Can be changed separately.

plugin/PDFView.qml
dcaliste: Oups, that's right. I've slightly changed this Rectangle showing bindings during previous iterations but didn't notice this inconsistency… Corrected now.

dcaliste:
Thank you for all the feedback. Next week I have two long train travels, I will try to apply the new design to the annotation implementation.

Besides, I've also noticed that the toolbar easily gets in the way when tapping on a link, producing annoying visual moves. I like the consistency of the dimming effect like anywhere else, but it may look strange to be applied just on the small part of the screen used by the toolbar. Folding the toolbar in a synchronized way (*i.e.* same duration) while the context menu appears, may be better. Let's see what design can propose.

PR #91, opened 2016-09-27
Fix crash on password protected pdf files

pvuorela:
Broken by 919030e00b51571, Poppler doesn't like numPages() asked from locked document.

Don't have pdf with 0 pages, so didn't test how that continues working. 

@dcaliste


dcaliste:
Fine with me. Sorry, I cannot put my hand on the PDF file that pushed me to create the erroneous pull request. Stupid, I should have attached it to the PR.

PR #93, opened 2016-10-03
[office] Avoid warnings for unexpected null receiver. Fixes JB#36476

pvuorela:
98dfd3a610c6 made use deleteLater without checking null pointer.

@dcaliste


dcaliste:
Looks ok. Good correction. I did notice the messages appeared after the deleteLater() trick was commited, but I didn't think of this for the reason (and falsely thought at the time that the reason was within Qt depending on how the finilisation of the application was done internally).

PR #94, opened 2016-10-04
[sailfish-office] Calculate number of visible cover list items dynami…

jpetrell:
…cally. Contributes to JB#14363

### DIFF
--- a/qml/CoverPage.qml
+++ b/qml/CoverPage.qml
@@ -45,21 +45,28 @@ CoverBackground {
     ListView {
         id: fileListView
 
-        property real itemHeight: Theme.iconSizeSmall + Theme.paddingSmall*2
-
+        property int itemHeight: height/maxItemCount
+        property int maxItemCount: Math.round(height/(dummy.height + Theme.paddingSmall))
         clip: true
         interactive: false
         model: window.fileListModel
         visible: window.documentItem === null
-        y: Theme.paddingLarge
-        width: parent.width
-        height: 7*itemHeight
+        anchors {
+            fill: parent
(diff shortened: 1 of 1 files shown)


qml/CoverPage.qml
```diff
@@ -45,22 +45,29 @@ CoverBackground {
...
+            bottomMargin: Theme.paddingLarge
+        }
 
         delegate: CoverFileItem {
             width: fileListView.width
             height: fileListView.itemHeight
             text: model.fileName
             iconSource: window.mimeToIcon(model.fileMimeType)
         }
+        Label {
```
dcaliste: I don't know if it's a good suggestion, but anyway:

Would it be better to create here a dummy CoverFileItem instance, so like that if later on, someone who makes changes in the layout of CoverFileItem (and its height is not a label height anymore), won't have to update here also ?

This suggestion, though requires to add ```height: label.height``` in CoverFileItem.qml.

pvuorela:
LGTM

qml/CoverPage.qml
jpetrell: Currently height is explicitly determined by the user of CoverFileItem so I think deriving label height from external item is also ok. That being said I wouldn't normally write layouts like this (implicit height is normally defined by the item itself), but rewriting the layout feels a bit of an overkill too. Adding invisible dummy CoverFileItem would add more loading overhead than adding just a label, I will replace the label with FontMetrics QML element once Qt 5.6 is in use.

dcaliste: I thank you for the explanation.

PR #95, opened 2016-10-10
[fontsize] Avoid creating dummy labels for determining line height. C…

jpetrell:
…ontributes to JB#14363

### DIFF
--- a/qml/CoverPage.qml
+++ b/qml/CoverPage.qml
@@ -16,7 +16,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-import QtQuick 2.0
+import QtQuick 2.4
 import Sailfish.Silica 1.0
 
 CoverBackground {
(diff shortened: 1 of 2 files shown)


qml/CoverPage.qml
```diff
@@ -63,10 +63,9 @@ CoverBackground {
             text: model.fileName
             iconSource: window.mimeToIcon(model.fileMimeType)
         }
-        Label {
-            id: dummy
-            visible: false
-            text: "Dummy"
+        FontMetrics {
+            id: fontMetrics
+            font.pixelSize: Theme.fontSizeMedium
```
rainemak: Will ```FontMetrics``` read default font family or should that be defined here as well?
```
font {
    pixelSize: Theme.fontSizeMedium
    family: Theme.fontFamily
```

rainemak: Replying to myself... of course FontMetrics must read default font family.

rpm/sailfish-office.spec
```diff
@@ -16,7 +16,7 @@ BuildRequires: cmake
 BuildRequires: qt5-qttools-linguist
 Requires: calligra-components >= 2.7.9+git4
 Requires: calligra-filters >= 2.7.9+git4
-Requires: sailfishsilica-qt5 >= 0.21.63
+Requires: sailfishsilica-qt5 >= 0.23.14
```
rainemak: I don't see that this commit/change would need this requirement. Qt5.6 (quick) on the other hand is required. (I'm seeing indirect reasoning).

pvuorela:
Concur that dependency is really on qt quick, not Silica, but indirectly it ensures new enough version already so good enough for me. LGTM.

jpetrell:
I'll add correct dependency.

Issue #89, opened 2016-09-07
Go back after internal link


dcaliste:
That would be nice to have a UI action to go back to previous position in a PDF document after an internal goto link has been tapped. Use case is typically a goto link for a figure or a chapter reference that send some pages further, and after one has seen the figure, one would like to go back to his previous reading position in the text.

Would it be a good idea to have a back arrow in the new toolbar design that appears after internal link clicking ? IMHO, it would not be easily discoverable like that but less space eating than a back arrow always present.

Should we ask design or they are quite occupied already and this is a minor request ?

pvuorela:
Should consult with design as it uses space from the toolbar.

After annotation buttons, there could still be space left for more buttons, but need to be careful. For comparison Adobe reader doesn't provide back/forward, but has support for creating bookmarks on the document.

If done, maybe with just one level of history, skipping forward button and maybe back button even disappearing when moving +-1 page away from link target. Just quick ideas.

dcaliste:
> If done, maybe with just one level of history, skipping forward button and maybe back button even disappearing when moving +-1 page away from link target. Just quick ideas.

That's exactly my opinion, indeed. The use case is really the one described before : following a link for a side reading, and then go back to the main reading. If reading continue after the goto link, then the back button disappears.

> support for creating bookmarks on the document.

I was also thinking about this yesterday evening. They may have their places in the table of content page (*i.e.* the attached page) where one can navigate quickly in a document, but I didn't figure out how to intuitively create them. I was also wondering how to properly label them to be usefull. Possible use case: one read a 10 document and notice that a figure on page 4 is quite important. One selects some words in the figure caption (or the full caption) and make a bookmark for this selection. Then a listItem with the elipsized text of the bookmark is available in the ToC page for instance. The same if a paragraph in a long document looks of particular importance: selection -> bookmark. But it looks a bit of a double usage with the highlight annotation (except for the quick access in the ToC page). Should these two things be united (heighlight creating quick accesses) ?

What would be interesting also about bookmarks, would be the ability for the search field (in the document list) to search also on the bookmark labels. It converges with the discussion we had some time ago about the tagging of documents.

Well, well, well, all ideas and possibilities about bookmarks are not all clear for me, and I need to continue to think about them. I'll open a dedicated issue entry when these will be clearer in my mind.

PR #96, opened 2016-10-17
[office] Add a back button after an internal link has been taped.

dcaliste:
Add a back button in the toolbar when tapped on an internal goto link, see #89. Also add smooth scroll to the link location.

When tapping on an internal goto link, the view scrolls to the link target location and the toolbar is shown. The toolbar is kept visible, up to the moment the back button is either tapped, the view is scrolled more than the height of a screen, or the zoom changes. Going back to the previous location is also smoothly animated.

### DIFF
--- a/plugin/PDFView.qml
+++ b/plugin/PDFView.qml
@@ -30,7 +30,8 @@ SilicaFlickable {
     property alias itemWidth: pdfCanvas.width
     property alias itemHeight: pdfCanvas.height
     property alias document: pdfCanvas.document
-    property alias currentPage: pdfCanvas.currentPage
+    property int currentPage: !quickScrollAnimation.running
+                              ? pdfCanvas.currentPage : quickScrollAnimation.pageTo
     property alias selection: pdfSelection
     property alias selectionDraggable: selectionView.draggable
     property bool canMoveBack: (_contentYAtGotoLink >= 0)
@@ -39,6 +40,7 @@ SilicaFlickable {
     property QtObject _feedbackEffect
 
     property int _pageAtLinkTarget
+    property int _pageAtGotoLink
     property real _contentXAtGotoLink: -1.
     property real _contentYAtGotoLink: -1.
 
@@ -120,10 +122,27 @@ SilicaFlickable {
         }
     }
 
-    function scrollTo(pt) {
-        scrollX.to = pt.x
-        scrollY.to = pt.y
-        scrollAnimation.start()
+    function scrollTo(pt, pageId) {
+        if ((pt.y < base.contentY + base.height && pt.y > base.contentY - base.height)
+            && (pt.x < base.contentX + base.width && pt.x > base.contentX - base.width)) {
+            scrollX.to = pt.x
+            scrollY.to = pt.y
+            scrollAnimation.start()
+        } else {
+            var deltaY = pt.y - base.contentY
+            if (deltaY < 0) {
+                deltaY = Math.max(deltaY / 2., -base.height / 2.)
+            } else {
+                deltaY = Math.min(deltaY / 2., base.height / 2.)
+            }
+            leaveX.to = (base.contentX + pt.x) / 2
+            leaveY.to = base.contentY + deltaY
+            returnX.to = pt.x
+            returnY.from = pt.y - deltaY
+            returnY.to = pt.y
+            quickScrollAnimation.pageTo = pageId
+            quickScrollAnimation.start()
+        }
     }
 
     function moveBack() {
@@ -131,7 +150,7 @@ SilicaFlickable {
             return
         }
 
-        scrollTo(Qt.point(_contentXAtGotoLink, _contentYAtGotoLink))
+        scrollTo(Qt.point(_contentXAtGotoLink, _contentYAtGotoLink), _pageAtGotoLink)
 
         _pageAtLinkTarget = 0
         _contentXAtGotoLink = -1.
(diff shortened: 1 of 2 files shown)


pvuorela:
Seems a nice feature. Couple details, though:
- talked with design about scroll transition, seemed a bit restless by going past unloaded pages. Would be better following fast scroll style, but that's under Silica private. If difficult to use, could just stick with old style for now.
- page number item doesn't seem to scale well and back button makes it worse. Take a bit document (hello, pdf reference) and text goes beyond item boundaries and screen. Could perhaps check minimum size for label, e.g. "[lastPage] | [lastPage]"  + little padding? Could make other buttons more left aligned, but that should be lesser evil.

pvuorela:
As last resort for text fitting can also use on Label:
    width: Math.min(parent.width - Theme.paddingSmall, implicitWidth)
    fontSizeMode: Text.HorizontalFit

Seems good enough when there is no back button. Could create a pull request to handle current situation with that already. For back button shown, maybe could keep page number item same width as normally, but scale all the others?

dcaliste:
I agree that scroll (either for goto links or search results) should follow the QuickScroll style when going further than one or two pages. I didn't think about it but it's a very good suggestion. I'll do what I can achieve with what is under BSD in Sailfish/Silica and report here later.

dcaliste:
I've added a commit trying to mimick what happen in ```private/FastScrollAnimator.qml```. It's not as sophisticated as implemented there, but it should keep the spirit and durations. What do you think ?

pvuorela:
Didn't yet go to code side, but scrolling looks so much nicer now. As small detail the page counter can briefly show start+1 and end-1 while animating. Could consider avoiding those intermediate page numbers?

Having this on top of my text fitting, can see how the text size changes if the document is long enough. Could maybe use minimum width of Screen.width / 4 on that? (Extra points for omitting on screen.sizeCategory > Screen.Medium which likely just has enough space)

dcaliste:
> As small detail the page counter can briefly show start+1 and end-1 while animating. Could consider avoiding those intermediate page numbers?

That's true, but I have no idea how to prevent this easily (signal being emitted by PDF.Canvas object). But, I propose an alternative solution : show the toolbar only after the scroll is complete and hide the toolbar as soon as the scroll back is started (instead of waiting for the timeout). Like that, it's not obvious that the page numbers are changing.

> Could maybe use minimum width of Screen.width / 4 on that?

I've added a simple commit implementing this, separated from the back button commits. The commit may go by itself if you think so, or together with the back button PR.

pvuorela:
> That's true, but I have no idea how to prevent this easily (signal being emitted by PDF.Canvas object). But, I propose an alternative solution : show the toolbar only after the scroll is complete and hide the toolbar as soon as the scroll back is started (instead of waiting for the timeout). Like that, it's not obvious that the page numbers are changing.

For example something like PDFView: property int currentPage: _transitioningTo >= 0 ? _transitioningTo : pdfCanvas.currentPage, transition to set overriding page number when starting and setting -1 when done. 

Temporary hiding doesn't sound good to me, intermediate page number could be lesser evil.

dcaliste:
> For example something like PDFView: property int currentPage

Of course, silly me. I don't know why I was focussing on the currentPage not being emitted in such a scheme when transition finishes, which is wrong of course. Anyway, I've updated the last commit (the one implementing quick scroll) to display directly the destination page in case of quick jump.

I've also reestablished the toolbar being shown as soon as link is activated, but I keep the automatic hiding as soon as back button is clicked. If you prefer, I can put it back to the previous behaviour with the toolbar being hidden after timeout after going back.

plugin/PDFView.qml
```diff
@@ -223,8 +311,7 @@ SilicaFlickable {
                 selection: pdfSelection
 
                 onLinkClicked: base.linkClicked(linkTarget, contextHook)
-                onGotoClicked: base.goToPage(page - 1, top, left,
-                                             Theme.paddingLarge, Theme.paddingLarge)
+                onGotoClicked: base.gotoClicked(page, top, left)
```
pvuorela: Only handler for signal is in the same file. Could just have some internal function called here?

plugin/PDFView.qml
```diff
@@ -142,6 +208,28 @@ SilicaFlickable {
...
+    onGotoClicked: {
+        var pt = base.contentAt(page - 1, top, left,
+                                Theme.paddingLarge, Theme.paddingLarge)
+        _contentYAtLinkTarget = pt.y
+        _pageAtGotoLink = pdfCanvas.currentPage
+        _contentXAtGotoLink = contentX
+        _contentYAtGotoLink = contentY
+        scrollTo(pt, page)
+    }
+    onContentYChanged: {
```
pvuorela: Generally I'd avoid binding to contentY unless it's really needed, will be executed quite many times. How about storing target page instead of y position, and then follow how far pdfCanvas.currentPage is?

plugin/PDFView.qml
dcaliste: That's right. The toolbar was made visible on gotoLink signal before, but then, I switched to listen to change of canMoveBack which more logical IMHO. Forgot to remove the signal then. Corrected.

pvuorela:
LGTM.

But still nitpicking a little: second commit includes a merge conflict. If you want, I can also squash last two commits and push that.

dcaliste:
> second commit includes a merge conflict.

Arf, really poorly done rebase… It's corrected.

PR #97, opened 2016-10-21
[office] Make sure PDF page number's text doesn't go past item area.

pvuorela:
@jpetrell @dcaliste


dcaliste:
That's a good idea. LGTM.

PR #100, opened 2016-11-23
[office] Use the same scroll animator as for goto link for search matching.

dcaliste:
Factorise the code in PDFView.qml to use the same scrolling animator for goto link and search matching. This solves also the issue of seing all empty pages scroll when the next match is far away.

### DIFF
--- a/plugin/PDFView.qml
+++ b/plugin/PDFView.qml
@@ -96,14 +96,11 @@ SilicaFlickable {
         
         var match = searchDisplay.itemAt(index)
         var cX = match.x + match.width / 2. - width / 2.
-        transX.to = (cX < 0) ? 0 : (cX > pdfCanvas.width - width) ? pdfCanvas.width - width : cX
+        cX = Math.max(0, Math.min(cX, pdfCanvas.width - width))
         var cY = match.y + match.height / 2. - height / 2.
-        transY.to = (cY < 0) ? 0 : (cY > pdfCanvas.height - height) ? pdfCanvas.height - height : cY
+        cY = Math.max(0, Math.min(cY, pdfCanvas.height - height))
 
-        scaleIn.target = match
-        scaleOut.target = match
-
-        focusSearchMatch.start()
+        scrollTo(Qt.point(cX, cY), match.page, match)
     }
 
     function nextSearchMatch() {
@@ -122,11 +119,12 @@ SilicaFlickable {
         }
     }
 
-    function scrollTo(pt, pageId) {
+    function scrollTo(pt, pageId, focusItem) {
         if ((pt.y < base.contentY + base.height && pt.y > base.contentY - base.height)
             && (pt.x < base.contentX + base.width && pt.x > base.contentX - base.width)) {
             scrollX.to = pt.x
             scrollY.to = pt.y
+            scrollAnimation.focusItem = (focusItem !== undefined) ? focusItem : null
             scrollAnimation.start()
         } else {
             var deltaY = pt.y - base.contentY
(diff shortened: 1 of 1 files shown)


plugin/PDFView.qml
```diff
@@ -96,14 +96,11 @@ SilicaFlickable {
         
         var match = searchDisplay.itemAt(index)
         var cX = match.x + match.width / 2. - width / 2.
-        transX.to = (cX < 0) ? 0 : (cX > pdfCanvas.width - width) ? pdfCanvas.width - width : cX
+        cX = (cX < 0) ? 0 : (cX > pdfCanvas.width - width) ? pdfCanvas.width - width : cX
```
pvuorela: Old code, but something like cx = Math.max(0, Math.min(cx, pdfCanvas.width - width) could be more readable.

plugin/PDFView.qml
```diff
@@ -141,6 +143,11 @@ SilicaFlickable {
             returnY.from = pt.y - deltaY
             returnY.to = pt.y
             quickScrollAnimation.pageTo = pageId
+            if (focusItem !== undefined) {
+                quickScrollAnimation.focusItem = focusItem
```
pvuorela: Just thinking, but since both animations only pass the property further to focusAnimation, could just always assign the directly and save some lines. Though now this function doesn't need to know animations trigger another. Either way ok.

pvuorela:
Couple very minor comments, generally LGTM. Can already merge if you want.

Noted one thing more: quick scroll mode can trigger between two subsequent pages if the search results are enough far away. Could perhaps consider increasing distance required to enter quick scroll?

plugin/PDFView.qml
dcaliste: That's right, I prefer also the Math.min/max notation, fixed.

plugin/PDFView.qml
dcaliste: I prefer to delegate to the scroll() function, the task to activate the focus animation because otherwise, the caller needs to set the focusItem property for both scrollAnimation and quickScrollAnimation. IMHO, the caller should not know that implementation internally uses two different scroll animators.

dcaliste:
> Could perhaps consider increasing distance required to enter quick scroll?

Well, I find it correct like that, we use quick scroll if the move is more than one screen height, which is more or less the limit for the texture anyway, so we're currently masking any void region (except on arrival). If the next match is on next page but within flickable.height pixel reach, then it's still normal scroll.

But if use have feeling that quick scroll is triggered too often, we can increase the limit indeed to 1.5 height or something like that.

pvuorela:
> Well, I find it correct like that, we use quick scroll if the move is more than one screen height, which is more or less the limit for the texture anyway, so we're currently masking any void region (except on arrival). If the next match is on next page but within flickable.height pixel reach, then it's still normal scroll.

Well, if the matches are on following pages, quick scroll won't anyway save rendering work. Even if there's one page between, that should have been pre-rendered, right? Think normal scroll would help understand better how far user got by moving to next match as search mode doesn't have page number indication.

But adjusting the threshold should be independent of these changes, so I'll just merge now.

dcaliste:
Thank you for the merge. I've done this as you understand I guess, for the same concern than we had for the goto link action, *i.e* avoid seing a lot of blank page shapes scrolling fast when the next match is far away.

> as search mode doesn't have page number indication.

That is a relevant concern I think. The situation before the PR was that we know that we have scrolled a lot of pages. The situation is slightly degraded now, quick scroll masking the quantity of scroll (*e.g.* only one page if the zoom level is high).

One solution would be to display the page id of the current match in the label (beside the number of matches). But for now, I find this solution not very convenient, making the UI overcrowded. Any idea is welcome !

PR #102, opened 2016-12-02
[office] Fix pdf search string not fully visible on bigger font size.

pvuorela:
A little hacky, but does its job.

@jpetrell


jpetrell:
Opacity transition combined with boolean visibility binding often results in flicker?

pvuorela:
> Opacity transition combined with boolean visibility binding often results in flicker?

What do you mean? No visibility or opacity changes here.

jpetrell:
Sorry now I understand. Looks good on device. LGTM.

PR #103, opened 2016-12-08
[office] Use incremental search.

dcaliste:
Implement an incremental text search by signaling every 3 pages the additional matches since the last update. A singular modification is that now the search model advertised by the document is available since the beginning of the request with a zero count. A fraction of the job done is also signaled so one can implement a progress bar.

I had to change various things in the UI because some of the logic in SearchBarItem.qml was based on the assumption that the search model was given at the end of the search and that a count of zero was meaning no match. The clear / cancel button is now also a stop button that can stop the current search while keeping the results already found if pressed when searching.

I think, the biggest other change I made is that the search model is not anymore a reference on the search list owned by the thread but a copy that is slowly increased.

### DIFF
--- a/pdf/pdfdocument.cpp
+++ b/pdf/pdfdocument.cpp
@@ -29,7 +29,13 @@
 class PDFDocument::Private
 {
 public:
-    Private() : searching(false), completed(false), modified(false) { }
+    Private()
+        : searching(false)
+        , searchModel(nullptr)
+        , completed(false)
+        , modified(false)
+    {
+    }
 
     PDFRenderThread *thread;
 
@@ -50,10 +56,9 @@ PDFDocument::PDFDocument(QObject *parent)
     connect(d->thread, &PDFRenderThread::loadFinished, this, &PDFDocument::pageCountChanged);
     connect(d->thread, &PDFRenderThread::loadFinished, this, &PDFDocument::loadFinished);
     connect(d->thread, &PDFRenderThread::jobFinished, this, &PDFDocument::jobFinished);
-    connect(d->thread, &PDFRenderThread::searchFinished, this, &PDFDocument::searchFinished);
+    connect(d->thread, &PDFRenderThread::searchFinished, this, &PDFDocument::onSearchFinished);
+    connect(d->thread, &PDFRenderThread::searchProgress, this, &PDFDocument::onSearchProgress);
     connect(d->thread, &PDFRenderThread::pageModified, this, &PDFDocument::onPageModified);
-
-    d->searchModel = nullptr;
 }
 
 PDFDocument::~PDFDocument()
@@ -244,13 +249,16 @@ void PDFDocument::search(const QString &search, uint startPage)
     if (!isLoaded() || isLocked())
         return;
 
-    if (d->searchModel != nullptr) {
+    if (search.length() > 0) {
         delete d->searchModel;
+        // Ensure that QML is updated in the repeater displaying
+        // the search results, even in case the later 'new' allocates
+        // a new object on the same address.
         d->searchModel = nullptr;
         emit searchModelChanged();
-    }
+        d->searchModel = new PDFSearchModel;
+        emit searchModelChanged();
 
-    if (search.length() > 0) {
         d->searching = true;
         emit searchingChanged();
         d->thread->search(search, startPage);
@@ -259,9 +267,9 @@ void PDFDocument::search(const QString &search, uint startPage)
     }
 }
 
-void PDFDocument::cancelSearch()
+void PDFDocument::cancelSearch(bool resetModel)
 {
-    if (d->searchModel != nullptr) {
+    if (resetModel && d->searchModel != nullptr) {
         delete d->searchModel;
         d->searchModel = nullptr;
         emit searchModelChanged();
@@ -313,11 +321,16 @@ void PDFDocument::jobFinished(PDFJob *job)
     job->deleteLater();
 }
 
-void PDFDocument::searchFinished(const QList<QPair<int, QRectF>> &matches)
+void PDFDocument::onSearchFinished()
 {
-    delete d->searchModel;
-    d->searchModel = new PDFSearchModel(matches);
-    emit searchModelChanged();
     d->searching = false;
     emit searchingChanged();
+}
+
+void PDFDocument::onSearchProgress(float fraction, const QList<QPair<int, QRectF>> &matches)
+{
+    if (!d->searchModel)
+        return;
+
+    d->searchModel->addMatches(fraction, matches);
 }
--- a/pdf/pdfrenderthread.cpp
+++ b/pdf/pdfrenderthread.cpp
@@ -62,6 +62,7 @@ public:
 
     void run() {
         m_matches.clear();
+        m_prevSearchSize = 0;
         for (int i = 0; i < m_document->numPages() && !isInterruptionRequested(); ++i) {
             int ipage = (m_startPage + i) % m_document->numPages();
             Poppler::Page *page = m_document->page(ipage);
@@ -86,6 +87,14 @@ public:
             }
 
             delete page;
+
+            if ((i + 1) % 3 == 0 || i + 1 == m_document->numPages()) {
+                emit searchProgress((m_document->numPages() > 1)
+                                    ? float(i) / float(m_document->numPages() - 1)
+                                    : 1.f,
+                                    m_prevSearchSize, m_matches.size() - m_prevSearchSize);
+                m_prevSearchSize = m_matches.size();
+            }
         }
         if (!isInterruptionRequested())
             emit searchFinished();
@@ -93,11 +102,12 @@ public:
 
 signals:
     void searchFinished();
+    void searchProgress(float fraction, uint beginIndex, uint nNewMatches);
 
 private:
     Poppler::Document *m_document;
     QString m_search;
-    uint m_startPage;
+    uint m_startPage, m_prevSearchSize;
 };
 
 class Thread : public QThread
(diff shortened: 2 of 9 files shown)


pdf/pdfdocument.cpp
```diff
@@ -315,9 +318,17 @@ void PDFDocument::jobFinished(PDFJob *job)
 
 void PDFDocument::searchFinished(const QList<QPair<int, QRectF>> &matches)
 {
-    delete d->searchModel;
-    d->searchModel = new PDFSearchModel(matches);
-    emit searchModelChanged();
+    if (matches.count() != d->searchModel->count())
```
pvuorela: Is there any reason to pass or even keep matches elsewhere?

pdf/pdfrenderthread.cpp
```diff
@@ -86,18 +87,25 @@ class SearchThread: public QThread
             }
 
             delete page;
+
+            if ((i + 1) % 3 == 0 || i + 1 == m_document->numPages()) {
+                emit searchProgress(float(i) / float(m_document->numPages() - 1),
```
pvuorela: Divide by zero with documents of one page? 

Btw, currently getting toolbar on those requires zooming so it can be scrolled :)

plugin/SearchBarItem.qml
```diff
@@ -162,11 +183,12 @@ BackgroundItem {
         EnterKey.onClicked: {
             if (text != "") {
                 _searchText = text
+                focus = false
                 root.requestSearch(text)
             } else {
+                focus = false
```
dcaliste: I've added a comment. I don't like much having it in the two branches, but the call in the first branch must appear before requesting search, because loosing the focus will use matchCount property which may be changed by the request. But at the same time, it must be set after saving current text into _searchText, because loosing the focus will reset text (matchCount being zero on a starting search).

plugin/PDFView.qml
```diff
@@ -349,7 +349,13 @@ SilicaFlickable {
             property int currentIndex
 
             model: pdfCanvas.document.searchModel
-            onModelChanged: moveToSearchMatch(0)
+            onModelChanged: if (model) model.onCountChanged.connect(moveToFirstMatch)
```
dcaliste: I agree, currentIndex is only used in the root item functions, so I've moved it there and renamed it searchIndex.

But, I can't find how to move the moveToFirstMatch() stuff to the root item. I've done something like this:
```
Connections {
   target: searchDisplay
   onModelChanged: if (model) model.onCountChanged.connect(movetoFirstMatch))
}
function moveToFirstMatch() {
   if (!searchDisplay.count) return
   moveToSearchMatch(0)
   model.onCiountChanged.disconnect(moveToFirstMatch)
}
```
I'm stuck on the last line of the function. How to disconnect the signal handler for it to run only once (precisely, when the first found match is available) ?

pdf/pdfdocument.cpp
```diff
@@ -51,9 +57,8 @@ PDFDocument::PDFDocument(QObject *parent)
     connect(d->thread, &PDFRenderThread::loadFinished, this, &PDFDocument::loadFinished);
     connect(d->thread, &PDFRenderThread::jobFinished, this, &PDFDocument::jobFinished);
     connect(d->thread, &PDFRenderThread::searchFinished, this, &PDFDocument::searchFinished);
+    connect(d->thread, &PDFRenderThread::searchProgress, this, &PDFDocument::searchProgress);
```
dcaliste: Sure, renamed.

pdf/pdfdocument.cpp
dcaliste: No, I've kept passing a reference to the complete list as a remnant of the previous way to do. So there is two way to catch the search results, either build it on the way with the progress signal, or at the end with the finished signal. The latter is not used anymore. If you prefer, I can remove it.

Additionally, we can also delete the list in the thread after the finished signal has been emitted, since the client have now kept a copy of it. I'm not sure this represent a big memory print though.

pdf/pdfrenderthread.cpp
dcaliste: Ah, my bad. The if initially didn't have the ```|| i + 1== numPages()``` part, so the division by zero case cannot happen. Later, when I added the or case in the if, I didn't remember the division by zero case…

Thanks for spoting, Corrected.

plugin/SearchBarItem.qml
```diff
@@ -162,11 +184,14 @@ BackgroundItem {
         EnterKey.onClicked: {
             if (text != "") {
                 _searchText = text
+                focus = false // Need to be called after saving text
+                              // but before the request, see the code
+                              // in onActiveFocusChanged, when activeFocus is false.
```
pvuorela: Not sure I'm still following. onActiveFocusChanged clears text in some cases, so by this requestSearch("") gets emitted? Maybe related but now searching for bogus, "asdf", every other time when I focus the editor and click enter it either searches or just dismisses. Would work better by just leaving focus where it was and emitting _searchText with requestSearch?

Nitpick: suggest having multiline comments before the line, otherwise automatic indentation mangles this at some point.

pdf/pdfsearchmodel.h
```diff
@@ -25,12 +25,14 @@ class PDFSearchModel : public QAbstractListModel
...
     Q_OBJECT
     Q_PROPERTY(int count READ count NOTIFY countChanged)
+    Q_PROPERTY(float fraction READ fraction NOTIFY fractionChanged)
 
 public:
     enum PDFSearchModelRoles {
         Page = Qt::UserRole + 1,
         Rect
     };
+    explicit PDFSearchModel(QObject *parent = 0);
```
pvuorela: The old ctor below isn't used anymore? Remove?

pdf/pdfsearchmodel.cpp
```diff
@@ -23,16 +23,28 @@
...
         roles.insert(Rect, "rect");
     }
-    const QList<QPair<int, QRectF> > &m_matches;
+    QList<QPair<int, QRectF> > m_matches;
     QHash<int, QByteArray> roles;
+    float m_fraction;
 };
 
+PDFSearchModel::PDFSearchModel(QObject *parent)
+  : QAbstractListModel(parent), d(new Private())
```
pvuorela: Cosmetics: () on new are unnecessary, same earlier with "new PDFSearchModel()"

pdf/pdfdocument.cpp
pvuorela: I think it could make sense for simplicity to remove things that are not used. Would remove need to keep track of previous index etc if there thread didn't store any matches itself.

dcaliste: Alright, I've removed from the searchFinished signal the reference to the complete list, since not used anymore. It will be easy to readd it later if necessary.

> if there thread didn't store any matches itself.

Well, currently the search thread stores the full list, while building it and I keep it like that (as you said for index handling simplicity). But after searchFinished has been emitted, the list in the thread is not needed anymore, so may be cleared. Nevertheless, since it's the main thread that copies the list items from the search thread list, I'm afraid the searchFinished may be reached in the search thread before all the copying has been finished. So I think it's easier now to keep the list alive in the search thread to avoid using mutex to access it.

plugin/SearchBarItem.qml
dcaliste: > it either searches or just dismisses. 

Yes, I can reproduce. I've tried to correct this also. I've follow your advice, and simplified the logic there, using _searchText for the request (don't know why I didn't do it before…), even using a connection on_SearchTextChanged to emit the request indeed.

pdf/pdfsearchmodel.h
dcaliste: Yes, removed.

pdf/pdfsearchmodel.cpp
dcaliste: Cosmetically improved here and there ;)

dcaliste:
I've addressed all your suggestions of improvement. In addition, I've also modified:
* the implementation of the move to first match action, moving it outside the display model to the root document, as you suggested once.
* the PdfDocument::search() function to emit a searchModelChanged() signal with a null value before emitting it with the true new object. I've encountered cases while testing when the call to new provides the same memory address than before, making the onModelChanged signal of the display model not emitted.

pvuorela:
> the PdfDocument::search() function to emit a searchModelChanged() signal with a null value before emitting it with the true new object. I've encountered cases while testing when the call to new provides the same memory address than before, making the onModelChanged signal of the display model not emitted.

Sounds strange, sure about that? Should be c++ responsibility checking when signals need to be emitted.

plugin/PDFView.qml
```diff
@@ -218,6 +227,15 @@ SilicaFlickable {
         }
     }
 
+    Connections {
+        target: document
+        onSearchModelChanged: {
+            if (document.searchModel) {
+                document.searchModel.onCountChanged.connect(moveOnceToFirstMatch)
```
pvuorela: Just thinking as this connect()/disconnect() feels a bit hacky, would this be any better:

```
Connections {
   target: document
   onSearchModelChanged: initialMatch.moveToFirst = true
}
Connections {
   id: initialMatch
   property bool moveToFirst: true
   target: document.searchModel
   onTargetChanged: moveToFirst = true
   onCountChanged: {
      if (document.searchModel.count == 0) {
        moveToFirst = true
      } else if (moveToFirst) {
        moveToSearchMatch(0)
        moveToFirst = false
      }
   }
}
```

Would be agnostic whether model is changed or cleared and shouldn't care about null models.

dcaliste:
> Sounds strange, sure about that? Should be c++ responsibility checking when signals need to be emitted.

Sorry, I may be inaccurate in my description. The signal is indeed emitted each time by the C++ side, even on the same address. 

The issue is on QML side in the Repeater, where we have model: doc.searchModel. There, the onModelChanged signal is emitted (not the searchModelChanged from the document), only if doc.searchModel comes from a different memory address.

You may test this with the previous version, by searching 'fgjgfg', see that it fails, then reopen the searchfield and clear with back key (not the clear button), and search for another valid string. Maybe add a console.log(doc.searchModel) on searchModelChanged to check that the memory address is the same. It's not happening all the time.

pvuorela:
> The issue is on QML side in the Repeater, where we have model: doc.searchModel. There, the onModelChanged signal is emitted (not the searchModelChanged from the document), only if doc.searchModel comes from a different memory address.

Oh right, so in other words it makes no difference here since onModelChanged is now removed? In that case up to you whether to have intermediate null pointer state. Maybe comment could be slightly adjusted to cover the specific case.

On general, think this is looking good. Can merge whenever you see fit.

dcaliste:
> it makes no difference here since onModelChanged is now removed?

Not exactly, because in my tests, the repeater itself was broken because from my understanding it was not reacting anymore to any count changes.

I guess the C++ implementation look something like this:
```
repeater = new Repeater;
connect(document, searchModelChanged, repeater, onModelChanged);

onModelChanged(model)
{
   if (model == this.model) return;

   connect(model, countChanged, this, update);
}
```
We can see with that implementation that when document.model dies, the update connection is removed. Then when document.model is new but with the same memory address, the if line in the slot raises the bug with the repeater not being updated and not reacting anymore to count changes.

I've tried to make the comment a bit more explicit.

PR #104, opened 2016-12-08
[office] Defer link scan on a page on request instead at load time.

dcaliste:
Try to correct the long loading time of lengthy PDF files (like the PDF reference paper). This loading time is mainly due to the link scanning in all pages and in a smaller proportion to the construction of the table of content.

To remove the link loading time, I've made this done on demand for pages that are displayed and that has not yet retrieved the link rectangles. This works quite well IMHO, the initial loading time being well reduced. But currently there is one issue: the rendering jobs having the priority, the links appear on page after the pages themselves have been rendered.

Do you think this is an issue ? Should I find a way to better interlaced texture rendering and link scanning in the job queue ?

### DIFF
--- a/pdf/pdfcanvas.cpp
+++ b/pdf/pdfcanvas.cpp
@@ -39,6 +39,7 @@ struct PDFPage {
         , requested(false)
         , renderWidth(0)
         , texture(nullptr)
+        , linksLoaded(false)
     { }
 
     int index;
@@ -53,7 +54,8 @@ struct PDFPage {
 
     QList<Patch> patches;
 
-    QList<QPair<QRectF, QUrl> > links;
+    bool linksLoaded;
+    PDFDocument::LinkList links;
 };
 
 class PDFCanvas::Private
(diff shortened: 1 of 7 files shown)


rburchell:
How significant is the delay between rendering and linkage?

dcaliste:
Hello, the issue here is that if page rendering takes, let say 1.5 seconds, then the links (that takes only tens of ms to be retrieved) are displayed after these 1.5 s. If only one page is rendered, you don't see the flickering, but if two pages are rendered or more, the links magically appear on the already rendered pages *after* all the pages have been calculated.

I don't like much the situation introduced by this PR (but I like the situation improvement for lengthy PDFs), and I submit this PR to discuss about it and find together the best solution before including the code in master.

pvuorela:
Quickly testing it doesn't look like a big issue. And for link coloring we should anyway make them more subtle. 

Loading pdf reference came from ridiculously slow to just slowish. Excellent!

dcaliste:
> Loading pdf reference came from ridiculously slow to just slowish. Excellent!

The remaining slowness comes from the construction of the ToC (if I remember well my measurements). I've in my ToDo list to move this to a specific thread, or the job thread just after loading, with a placeholder saying it is building in the ToC page if one lands there before completion…

> it doesn't look like a big issue.

Interesting, that's what I wanted to discuss by submitting it here. I don't like much the fact to degrade the experimence for simple PDFs, but if the feeling is that it's not much of an issue, I prefer the code as it is introduced by this PR : quite simple. Any other solution I can think about to solve the ordering issue will make the base code either much more complex (reordering solution) or ridiculously verbose (add a specific thread for link parsing purposes) or dangerous for UI responsiveness (synchronous link parsing call for required pages in the layout routine).

pdf/pdfcanvas.cpp
```diff
@@ -453,6 +454,16 @@ QPointF PDFCanvas::fromPageToItem(int index, const QPointF &point) const
                    point.y() * page.rect.height() + page.rect.y());
 }
 
+void PDFCanvas::linksFinished(int id, const QList<QPair<QRectF, QUrl> > &links)
+{
+    PDFPage &page = d->pages[id];
+
+    page.linksAvailable = true;
```
pvuorela: Just a suggestion, but would linksLoaded or something on that direction be slightly better? Available could be understood as links exist on a page.

PR #105, opened 2016-12-08
Jb35554

jpetrell:
Documents app main page header no longer opens outside the view (caused by Binding element changing search field height to zero too late, after the content y had already been calculated).

Info label "No documents" no longer flickers on app launch even if user has documents. During search filtering "No documents found" label no longer incorrectly flickers to "No documents" during the hide transition. Info label vertical position no longer flickers during orientation change. Also moved the label a bit lower on portrait orientation where there is more space.

### DIFF
--- a/qml/FileListPage.qml
+++ b/qml/FileListPage.qml
@@ -135,14 +135,14 @@ Page {
 
         InfoLabel {
             parent: listView.contentItem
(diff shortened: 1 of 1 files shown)


jpetrell:
@pvuorela

pvuorela:
LGTM.

Could consider also removing the dummy header item. Think the item shouldn't get recreated unnecessarily anymore.

PR #107, opened 2016-12-16
[office] Delegate table of content build to a dedicated thread.

dcaliste:
Second PR for PDF document load time improvement.

This time, the idea is to move the creation of the table of content model into a dedicated thread, to avoid doing it in the load thread.

The busy indicator and the associated ready property may look a bit overkill, because they almost never appear, but one can see them for the reference pdf document, by taping on the page count in the toolbar. Otherwise, I've never seen them appear…

As a result, the reference PDF document, now opens in the same duration as in Evince on desktop (almost).

### DIFF
--- a/pdf/pdfrenderthread.cpp
+++ b/pdf/pdfrenderthread.cpp
@@ -120,11 +120,11 @@ public:
 
     void run() {
         QThread::exec();
-        // Delete pending search that may use document.
+        // Delete pending search and toc that may use document.
         delete searchThread;
+        delete tocModel;
         autoSaveTo();
         delete document;
-        delete tocModel;
         deleteLater();
     }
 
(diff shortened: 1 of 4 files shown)


pvuorela:
Now PDF reference loading is blazingly fast, nice!

...but Office crashes if that file is opened and user pops the page immediately :)

dcaliste:
> ...but Office crashes if that file is opened and user pops the page immediately :)

Ok, I can reproduce the crash. I will investigate, sorry for not noticing it before.

dcaliste:
I've moved the toc model deletion to happen before the document deletion (see pdf/pdfrenderthread.cpp#125, so the toc thread can finish its work before ressources are freed, see pdf/pdftocmodel.cpp#46.

pvuorela:
Working better now. Quite nice impact for relatively simple change again.

dcaliste:
Thank you.

PR #108, opened 2018-01-31
[sailfish-office] Adapt to changed ShareMethodList behavior. Contributes to JB#35260

jpetrell:
Don't pop sharing page like normally done.


pvuorela:
Think this was no longer needed.

PR #114, opened 2018-04-03
[office] Removed accidental Content property from the sharingpage. Contributes to JB#41068

Nrsisti:


pvuorela:
Codewise LGTM, but commit message could document not only what is changed but why it's changed. In this case because the content wasn't proper and was there by accident. Referenced bug is also closed, so maybe could use a still open one, 41068 perhaps as it's about sharing changes too.

pvuorela:
Looking good.

PR #120, opened 2018-08-17
[office] Avoid flickering empty state placeholder on app startup. Contributes to JB#42702

jpetrell:


jpetrell:
Good points. Will fix. :)

jpetrell:
Changed to using ready property. Now ready won't become true until the initial SQL query completes. Previous ready didn't mean much (basically just that componentComplete had happened), also nothing emitted readyChanged() signal.

dcaliste:
Looks good.

pvuorela:
Looking good to me too.

PR #121, opened 2018-10-29
Remove zoomtimeout

pvuorela:
ZoomTimeout was creating warnings on both old and new Calligra, and don't think it should have ever had any other effect as it's a private slot in the controller.

@jpetrell @dcaliste


jpetrell:
LGTM

PR #124, opened 2019-03-12
Silence debug unless explicitly asked with -d. 

pvuorela:
Calligra has tons of debug messages, seems like everything on by default.
Cannot silence just Calligra side separately as lot is done with plain
qDebug instead of the categorized version, so now disabling all but
qml/js debug by default.

@jpetrell @dcaliste

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1,6 +1,6 @@
 /*
- * Copyright (C) 2013-2014 Jolla Ltd.
- * Contact: Robin Burchell <robin.burchell@jolla.com>
+ * Copyright (C) 2013-2019 Jolla Ltd.
+ * Contact: Pekka Vuorela <pekka.vuorela@jolla.com>
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
@@ -22,13 +22,13 @@
 #include <QtQml>
 #include <QQuickView>
 #include <QQmlError>
-#include <QQmlContext>
 #include <QQmlEngine>
 #include <QQuickItem>
 #include <QDBusConnection>
 #include <QTranslator>
 #include <QLocale>
 #include <QDebug>
+#include <QLoggingCategory>
 
 #include <MDeclarativeCache>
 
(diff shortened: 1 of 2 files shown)


jpetrell:
Seems a bit like a shotgun approach, the correct solution would be to search and replace qDebugs in upstream to use logging category.

dcaliste:
Yes, indeed it's using a sledgehammer, but I agree with @pvuorela. This is a well done short term solution to the flood when opening an office document while keeping QML info. It will help to improve sailfish-office itself by viewing the QML messages at least! So LGTM.

Besides, I agree with you @jpetrell, that it's an easy task to conduct upstream. I've done it for the sheet Excel parser (because based on kDebug which is deprecated from KF5) and it was welcomed by upstream team. The flood mainly comes from the Word parser if I remember well. It's a task that I can try to do in the coming weeks and restricted to the Word parser, it's not looking so overwhelming than touching all the code base.

pvuorela:
Yea, this is quite rough. Got frustrated by the number of qdebugs and first on the list by the document I was testing using rtf-qt "temporary checkout" under 3rdparty dir, with upstream haven't been touched since 2011 or so.

I could try to adjust the logging upstream, but maybe have this now or within a couple of days so the version on next sailfish release wouldn't be so noisy?

dcaliste:
> I could try to adjust the logging upstream, but maybe have this now or within a couple of days so the version on next sailfish release wouldn't be so noisy?

Yes, this is a good short term solution.

Upstreaming proper logged categories is for the next step imho.

pvuorela:
To add: uh, it's not even just qDebugs but also std::cout which still gets through.

pvuorela:
Fixed couple plugins in calligra (waiting for merge) after which just silencing calligra categories ends up lot better already. Changed the rules accordingly as less intrusive.

dcaliste:
Great job, indeed. For all the accepted patches (see https://phabricator.kde.org/people/revisions/14024/), you can merge and push them yourself to master (scary the first time).

pvuorela:
> Great job, indeed.

Take that as lgmt still holds? :)

> For all the accepted patches (see https://phabricator.kde.org/people/revisions/14024/), you can merge and push them yourself to master (scary the first time).

Hm, not spotting any option on the review page, or do you mean I have push access to calligra git directly?

pvuorela:
> Hm, not spotting any option on the review page, or do you mean I have push access to calligra git directly?

Think this would require me to apply for a kde developer account. Perhaps should sooner or later, but do hope I could get the initial changes in without :)

dcaliste:
> Think this would require me to apply for a kde developer account.
Yes, that's it. You indeed have to do yourself a `git push upstream master` which I consider scary ;) Particularly when considering that you can actually push anything. Nothing is enforcing the "reviewed by:" and "differential revision:" mentions in the commit message. Nothing is enforcing to push in relation with the actual patch in phabricator…

PR #125, opened 2019-03-21
[office] Get rid of split view. Fixes JB#41071

jpetrell:


jpetrell:
Still WIP

- The PDF and Text Document viewer designs mostly follow Email viewer design
- Not happy with Presentation and Spreadsheet designs, but not yet sure what to do there. Now using Gallery-style toolbar there
- Controller zoom proxy provides decent performance during zooming, but the scaling calculations (negative scale values, proxy and normal pinch going out of sync) there have bugs I still need to fix
- To make scrolling smooth would be nice to get rid of ViewController in SpreadsheetPage and instead use Presentation -style rendering of the page to image and if user zooms rendering higher-resolution version
- To make zooming smooth would be nice to streamline ImageDataItem, e.g. avoid recreating the texture on each frame
- Tap to zoom out gesture is missing on spreadsheet and text document viewers, former is fixed if I get rid of ViewController, for the latter I need a some way to utilize zoom proxy during the zoom out transition

jpetrell:
I think the pull requests are now ready for review. Got another green light from design on the latest version.

Dropped the use of zoom proxy for text and spreadsheet documents, it is only marginally better performance-wise and comes with many issues of its own. The performance is painfully slow, but IMO would require major rewrite and out of scope of split view pattern removal task.

Presentation performance is not good either (also due to broken architecture), but managed to do few improvements there. First of all postponed loading of next slide until the scroll movement has ended to avoid hiccups in animation. Also pinch zooming is smooth if we avoid texture creation on every frame: https://git.merproject.org/mer-core/calligra/merge_requests/5

Tried to avoid use of legacy ViewController for spreadsheets (and use similar approach to presentation page), but the ContentsModel wasn't very good at rendering spreadsheets in correct dimensions for ImageDataItem so gave up.

plugin/DeleteButton.qml
```diff
@@ -0,0 +1,53 @@
...
+        if (!_remorsePopup.active) {
+            _remorsePopup.execute(text, action)
+        }
+    }
+
+    onSourceChanged: if (_remorsePopup && _remorsePopup.active) _remorsePopup.trigger()
+    onClicked: {
+        //: Delete an image
+        //% "Deleting"
+        remorseAction(qsTrId("gallery-la-deleting"), function() {
```
pvuorela: gallery?

jpetrell: Fixed.

jpetrell:
Fixed few review findings by Pekka:
- Text document index indicator sometimes showed wrong page index
- Fixed label coloring on presentation thumbnail navigation to work with light ambiences
- Moved delete and share actions on PDF viewer to pulley, 6 action toolbar didn't fit on smaller devices

plugin/DeleteButton.qml
```diff
@@ -0,0 +1,53 @@
...
+        if (!_remorsePopup.active) {
+            _remorsePopup.execute(text, action)
+        }
+    }
+
+    onSourceChanged: if (_remorsePopup && _remorsePopup.active) _remorsePopup.trigger()
+    onClicked: {
+        //: Delete an image
+        //% "Deleting"
+        remorseAction(qsTrId("sailfish-office-la-deleting"), function() {
```
pvuorela: This id is used in FileListPage.qml with same string, but bit different //: and intention.

plugin/ControllerFlickable.qml
```diff
@@ -0,0 +1,83 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import QtQuick 2.0
```
pvuorela: Just a suggestion, but could aim to introduce new QtQuick imports as 2.6 at this point.

plugin/DocumentFlickable.qml
```diff
@@ -0,0 +1,67 @@
...
+    // Make sure that _noGrabbing will be reset back to false (JB#42531)
+    Component.onDestruction: if (!visible) pageStack._noGrabbing = false
+
+    // Override SilicaFlickable's pressDelay because otherwise it will
+    // block touch events going to PinchArea in certain cases.
+    pressDelay: 0
+    interactive: !dragDetector.horizontalDragUnused
+    ScrollDecorator { color: Theme.highlightDimmerColor }
+
+    Binding { // Allow page navigation when panning the image near the top or bottom edge
```
pvuorela: The image or the document or something else?

plugin/CMakeLists.txt
```diff
@@ -12,29 +12,35 @@ install(TARGETS sailfishofficeplugin DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/qt5
 install(FILES
     qmldir
     ContextMenuHook.qml
+    ControllerFlickable.qml
+    DeleteButton.qml
+    DetailsPage.qml
+    DocumentFlickable.qml
+    DocumentHeader.qml
     DocumentPage.qml
     DocumentsSharingList.qml
```
pvuorela: This file is no longer used?

plugin/PDFSelectionHandle.qml
```diff
@@ -16,7 +16,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-import QtQuick 2.0
+import QtQuick 2.6
```
pvuorela: Not related to this line, but bug: have another handle near the left side of the display, try to grab and move it horizontally -> the whole page gets dragged.

plugin/PDFView.qml
```diff
@@ -168,44 +181,44 @@ SilicaFlickable {
...
+            NumberAnimation { id: scrollY; target: root; property: "contentY"; duration: 300; easing.type: Easing.InOutQuad }
         }
         ScriptAction {
             script: if (scrollAnimation.focusItem) {
-                focusAnimation.targetItem = scrollAnimation.focusItem
-                focusAnimation.start()
-            }
+                        focusAnimation.targetItem = scrollAnimation.focusItem
+                        focusAnimation.start()
+                    }
```
pvuorela: Stylistic thing, but could perhaps have a block of its own for script value?

plugin/PDFView.qml
```diff
@@ -278,29 +292,29 @@ SilicaFlickable {
...
-
+        y: header.height
+        width: root.width
         spacing: Theme.paddingLarge
-        flickable: base
+        flickable: root
         linkWiggle: Theme.itemSizeMedium / 2
         linkColor: Theme.highlightColor
-        pagePlaceholderColor: Theme.highlightColor
+        pagePlaceholderColor: "white"
```
pvuorela: I think I liked the older version better myself, at least on light ambience it nicely shows where the pages start and stop when they are not yet loaded. For another detail the placeholder transparency is hard-coded in canvas.

plugin/ShareButton.qml
```diff
@@ -0,0 +1,35 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+
+IconButton {
+    property DocumentPage page
```
pvuorela: Ok too, but a little thinking if the page's relevant property values would be better explicitly passed to avoid coupling.

plugin/DeleteButton.qml
jpetrell: Fixed.

plugin/PDFDocumentPage.qml
```diff
@@ -155,273 +150,267 @@ DocumentPage {
...
+                        pageStack.pop()
+                    })
+                }
+                Component {
+                    id: remorsePopupComponent
+                    RemorsePopup {}
+                }
+            }
+            MenuItem {
+                //: This menu option can be used to share the note via Bluetooth
```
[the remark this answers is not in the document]
jpetrell: Fixed.

plugin/PDFView.qml
jpetrell: The highlight color is often much more saturated than background, and the document contents are unlikely to be colored in highlight color. Documents are predominantly white, thus white placeholder minimizes flicker and calms down the scrolling.

plugin/ShareButton.qml
jpetrell: I had the properties initially exposed separately, but passing page resulted in cleaner code. As app's internal implementation detail doesn't matter much.

plugin/CMakeLists.txt
jpetrell: Removed.

plugin/PDFSelectionHandle.qml
jpetrell: Fixed in Silica side. DragDetectorItem didn't respect preventStealing.

plugin/TextDocumentToCPage.qml
```diff
@@ -49,7 +49,7 @@ Page {
                     verticalCenter: parent.verticalCenter
                 }
                 text: model.title
-                color: (model.contentIndex == page.document.currentIndex || bg.highlighted) ? Theme.highlightColor
+                color: (model.contentIndex + 1 == page.document.currentIndex || bg.highlighted) ? Theme.highlightColor
                                                                                             : Theme.primaryColor
```
pvuorela: Nitpick: indentation

plugin/SpreadsheetPage.qml
```diff
@@ -18,81 +18,128 @@
...
+    busy: doc.status != Calligra.DocumentStatus.Loaded
+
+    FadeBlocker {
+        id: fadeBlocker
+        color: "white"
+        Binding {
+            when: fadeBlocker.fullscreen
+            target: __silica_applicationwindow_instance.__quickWindow
+            property: "color"
+            value: fadeBlocker.color
```
pvuorela: Suppose it's somewhere along these lines that break spread sheet when app gets background and then foreground again. Black screen and eglErrors.

plugin/PresentationPage.qml
```diff
@@ -18,55 +18,142 @@
...
+        id: overlay
+        property bool active: true
+
+        enabled: active && !deleteButton.remorseActive
+        anchors.fill: parent
+        opacity: enabled ? 1.0 : 0.0
+        Behavior on opacity { FadeAnimator {}}
+
+        FadeGradient {
+            id: headerBackground
```
pvuorela: No references.

plugin/TextDocumentPage.qml
```diff
@@ -18,76 +18,113 @@
...
+            y: -height
         }
     }
 
+    ToolBar {
+        id: toolbar
+
+        flickable: flickable
+        anchors.top: flickable.bottom
+        forceHidden: doc.failure
```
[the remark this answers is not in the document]
jpetrell: Fixed.

plugin/DocumentPage.qml
```diff
@@ -20,90 +20,22 @@ import QtQuick 2.0
...
-                                                      })
-                }
-            }
-        }
-    }
-
     onStatusChanged: {
         if (status == PageStatus.Active) {
-            drawer.open = false
             window.documentItem = documentItem
```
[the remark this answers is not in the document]
jpetrell: Fixed.

plugin/TextDocumentPage.qml
```diff
@@ -18,76 +18,113 @@
...
+        enabled: !page.busy
+        opacity: enabled ? 1.0 : 0.0
+        Behavior on opacity { FadeAnimator { duration: 400 }}
+
+        DeleteButton {
+            id: deleteButton
```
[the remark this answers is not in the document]
jpetrell: Fixed. And couple of similar cases in other pages.

plugin/SpreadsheetPage.qml
jpetrell: Ok sounds like setting the color to clear the OpenGL context has issues. Will investigate.

plugin/PresentationPage.qml
jpetrell: Fixed.

plugin/SpreadsheetPage.qml
jpetrell: Fixed.

jpetrell:
Reinstated the regressed, old cover code back. The cover rendering was really flaky, made image grab async (16ms timer) and postponed the grab until Calligra reports the document status as loaded. It should be now better, but in general am tempted to drop the document previews from the covers due to overall unreliability of having timing-based image grabs/captures.

pvuorela:
Yea the cover state is a bit sad. I do like how it looks and mostly it works despite the flakiness. Then again we shouldn't have code depending on luck. Ideally we should get the content there with some other method than image grab.

For another case notice there's still something wrong on text document page numbering. On content page +1 item is highlighted compared and tapping a page item takes to either page or page-1.

plugin/DeleteButton.qml
pvuorela: Still an image :)

pvuorela:
For minor detail on cover, pdf document now includes the page header and pulldown menu content. Guess I can live with that, though.

jpetrell:
PDF active cover no longer includes pulley or page header, but uses the canvas as the source item. Fixed one offset problem with text document page numbering.

pvuorela:
Page numbering looks good, but pdf cover also has these properties "for cover state" that didn't seem moved here. Pdf cover itself seemed kind of stretched and having some empty areas.

pvuorela:
Ok, see pdf reverting a bit to previous state.

LGTM, can fix the cover details separately.

PR #126, opened 2019-04-11
Enable opening plain text files.

adenexter:


jpetrell:
- Add horizontal page margin on the left and right edge, and padding large at the beginning and the end
- After the new Documents design goes in it will probably gain PageHeader with title showing filename, and delete option also in full-screen mode
- For maximal contrast and to align with other text viewers (e.g. Email plain text viewer) always use white background and black text
- Wonder if we should take similar viewer in use for plain text emails
- Zooming font size is cool, though very inconsistent from how zooming works on other. Need to show to Martin Schuele still and ask his feedback.
- The text in between the pinched fingers should stay visible like with normal zooming (i.e. update contentY during pinching)

jpetrell:
Also missing TextLinking usage, numbers, email addresses, urls, etc. should be highlighted.

pvuorela:
Split view removal in, needs rebase.

jpetrell:
Looks great. :) Add padding large top and bottom margin to text paragraph inside the white background. Will ask Martin for feedback. Will ask Martin for wrap on/off icon. Also would be nice to add text size setting to the toolbar (or through pinch zooming), but I guess could merge the current version already before such additional features.

jpetrell:
LGTM. Remember to add some top and bottom margin inside the white content item before merging.

plugin/PlainTextDocumentPage.qml
```diff
@@ -0,0 +1,130 @@
...
+                    width: parent.width - (2 * x)
+                    wrapMode: documentPage.wrap ? Text.Wrap : Text.NoWrap
+                    color: Theme.darkPrimaryColor
+                    linkColor: Theme.highlightColor
+                    font.pixelSize: Math.round(documentPage.fontSize)
+                    text: lineText
+                    textFormat: Text.StyledText
+
+                    onImplicitWidthChanged: {
+                        if (implicitWidth  > documentPage.maximumWidth) {
```
jpetrell: Two spaces before >

jpetrell:
Spotted one issue: the deletion is broken, both when back-stepping during remorse or waiting for remorse.

plugin/plaintextmodel.h
```diff
@@ -0,0 +1,109 @@
...
+#include <QUrl>
+
+#include <vector>
+
+class PlainTextModel : public QAbstractItemModel
+{
+    Q_OBJECT
+    Q_PROPERTY(QUrl source READ source WRITE setSource NOTIFY sourceChanged)
+    Q_PROPERTY(Status status READ status NOTIFY statusChanged)
+    Q_PROPERTY(int count READ rowCount NOTIFY countChanged)
```
pvuorela: Quite a minor thing, but ponder if naming like lineCount, or lines more emphasized in general, would be clearer.

adenexter: Changed to 'lineCount'. 'lines' is ambiguous, it could return the number of lines or return a list of them.

adenexter:
> Spotted one issue: the deletion is broken, both when back-stepping during remorse or waiting for remorse.

Fixed

PR #127, opened 2019-05-27
[sailfish-office] Add custom error UI for calligra documents. 

pvuorela:
Depends on:
https://git.merproject.org/mer-core/calligra/merge_requests/9

WIP at this point. Wanted to quickly try out at this point how my earlier error reporting gets to the app side. Already a lot better than the existing error reporting.

### DIFF
--- a/plugin/FullscreenError.qml
+++ b/plugin/FullscreenError.qml
@@ -1,0 +1,57 @@
+/*
+ * Copyright (C) 2019 Jolla Ltd.
+ * Contact: Pekka Vuorela <pekka.vuorela@jolla.com>
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; version 2 only.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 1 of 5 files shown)


pvuorela:
Adjusted layout a bit etc. Think good enough now.

plugin/FullscreenError.qml
```diff
@@ -0,0 +1,56 @@
...
+        x: Theme.horizontalPageMargin
+        width: parent.width - 2*x
+        spacing: Theme.paddingMedium
+        anchors.verticalCenter: parent.verticalCenter
+
+        Image {
+            id: warningIcon
+
+            anchors.horizontalCenter: parent.horizontalCenter
+            source: "image://theme/icon-l-attention?" + Theme.highlightColor
```
adenexter: I'm not really sure where we stand going forward with the theme color provider, but maybe a HighlightImage and hightlighted: true instead here?

adenexter:
Is the [sailfish-office] tag in the PR title without a bug number going to enrage CI?

pvuorela:
Don't think PR title matters. Commit has a bug reference.

plugin/FullscreenError.qml
```diff
@@ -0,0 +1,56 @@
...
+    id: root
+
+    property string error
+
+    anchors.fill: parent
+
+    Rectangle {
+        anchors.fill: parent
+        opacity: 0.4
+        color: Theme.highlightDimmerColor
```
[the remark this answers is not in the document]
pvuorela: Not sure, highlightBackgroundOpacity docs talk about usage with highlightBackgroundColor. For another thing isn't dimming kind of opposite to highlighting? In that sense the property name is strange?

plugin/FullscreenError.qml
pvuorela: Sure.

PR #130, opened 2019-07-05
Add translation support for some Calligra strings

pvuorela:
Would be nice if Calligra had its translations available, but they
are pain to work with, being located in huge kde subversion repository,
away from the main code. Also everything included would mean
tens of megabytes of data when only a few string are ever shown
to the user on sailfish-office. And even then we'd be forced to show
what Calligra decided as format and available translations.

So let's just hack the few known strings separately for now.

@adenexter @dcaliste @jpetrell

### DIFF
--- a/plugin/FullscreenError.qml
+++ b/plugin/FullscreenError.qml
@@ -23,8 +23,85 @@ TouchBlocker {
     id: root
 
     property string error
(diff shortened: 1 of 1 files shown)


adenexter:
LGTM.

PR #132, opened 2019-08-13
[documents] Adapt remorse usage to the new design. Contributes to JB#46248

jpetrell:


pvuorela:
> We should optimize for the common case, which is that you tapped delete because you wanted to delete the item and no longer want to see the document lingering on the screen. For the accidents we provide an easy way to undo the action.

I guess my concern is delete, not being a common case to begin with, is being offered so easily available.

But small detail and this works now better, with less code even, so LGTM.

Though one more thing I noticed: open last document in the list, trigger delete inside there and then cancel on the main page -> something flashes at the bottom, but the item gets back visible outside the screen. Ideally could perhaps expect the view to scroll a bit.

And even if the item is restored in the middle of the page, it might be nice if it did some quick blink.

PR #135, opened 2019-08-22
[sailfish-office] Correct note creation not working in PDF documents.  Fixes TJC#212151

dcaliste:
Simple typo making note creation failing, see [TJC](https://together.jolla.com/question/212151/cant-create-notes-in-pdf/).

PR #137, opened 2019-08-26
Various minor corrections in annotation handling

dcaliste:
It's a follow-up of #135 after comment from poster on TJC mentioning that notes where not created at the right place.

Testing this, I've found several minor issue with annotation handling after migration to latest design:
* dismissing a long press context menu was making the flickable scroll to top.
* creation and modification of a note from the context menu was doing nothing because of a wrong id in PDFContextMenu*.qml.
* notes created from toolbar were not positioned properly because the 'at' argument was missing the content offsets that existed before.

Quick tests of annotation creation and modification in a zoomed-in document were not showing additional bugs. Maybe @pvuorela or @jpetrell if you have time to test also, you may find something else. Tell me, I'll try to give a look. Notice, annotation deletion is currently segfaulting… I still need to investigate this further.

### DIFF
--- a/plugin/ContextMenuHook.qml
+++ b/plugin/ContextMenuHook.qml
@@ -27,6 +27,7 @@ Item {
     property alias backgroundOpacity: background.opacity
 
     property real _flickableContentHeight
+    property real _flickableContentYAtOpen
     property bool _opened: _menu ? _menu._open : false
 
     property int _hookHeight
@@ -45,7 +46,7 @@ Item {
 
     function showMenu(menu) {
         _menu = menu
-        menu.show(hook)
+        menu.open(hook)
         _flickableContentHeight = _menu._flickable.contentHeight
     }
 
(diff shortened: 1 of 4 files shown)


dcaliste:
@pvuorela, I've found why deleting an annotation from the edit page pulley menu was crashing. It's coming from the time when page transition has been modified to start as soon as possible and not only when destination page is ready.

My bad, the annotation handling is not locking the pdf document object. And the annotation was removed during rendering… I've swapped the calls as an emergency fix (see the latest commit). The proper fix will come later by adding locking mechanism for annotation handling.

dcaliste:
@pvuorela, reading again the locking mechanism code, I remember why it was done like that… The rendering itself is not locking, since it would mean that the rendering thread would block the main thread each time some rendering occurs (which means most of the time) and the main thread would like to touch the pdf document.

There are two solutions:
- transform the addAnnotation and removeAnnotation methods in pdfrenderthread.cpp to inherit the job class and be treated as a job in the queue, so only one job is executed at a time, and rendering will not happen, by design during annotation modification.
- keep it like that and pay attention when using the addAnnotation and removeAnnotation calls to be done before rendering happens.

The first solution is future proof. I'll try to implement it in the coming days to see if it's not too much additional bloat. In my opinion, the current proposed patch (second solution) is a good quick fix for the issue.

pvuorela:
Deleting a note crashed the app for me, but think this should be better anyway already -> LGTM.

Another issue I noticed is that now if i press and hold for context menu in portrait and then switch to landscape, only existing half of the width is used. Not related to these changes, though.

PR #138, opened 2019-08-28
Correct search bar progress not filling-up toolbar height. Contributes TJC#212568

dcaliste:
As [reported by @rozgwi on TJC](https://together.jolla.com/question/212568/documents-misaligned-toolbar-on-xperia-x/), the search progress bar was not filling properly up the toolbar height. I'm also pushing an additional commit to correct the page indicator not highlighted on press.

@pvuorela and @jpetrell, during the redesigned, you choose to remove BackgroundItem from toolbar items. I guess it's by design to use only IconButton there. I've one question though: the search icon is still a BackgroundItem and when pressed, it is highlighted with a solid coloured rectangle, while all other icon buttons are not. Should we correct this inconsistency by making SearchBarItem to inherit IconButton instead of BackgroundItem ? Or should we add BackgroundItem to all other IconButtons ?

### DIFF
--- a/plugin/ToolBar.qml
+++ b/plugin/ToolBar.qml
@@ -75,8 +75,8 @@ PanelBackground {
         id: contentItem
 
         spacing: Theme.paddingLarge
-        anchors.verticalCenter: parent.verticalCenter
         x: Math.max(0, parent.width/2 - width/2)
+        height: parent.height
     }
 
     Connections {
--- a/plugin/IndexButton.qml
+++ b/plugin/IndexButton.qml
@@ -29,7 +29,7 @@ MouseArea {
     enabled: count > 1 && allowed
     opacity: count > 0 && allowed ? (count > 1 ? 1.0 : Theme.opacityHigh) : 0.0
     width: Theme.itemSizeSmall
-    height: Theme.itemSizeSmall
+    height: parent.height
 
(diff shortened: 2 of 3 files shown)


plugin/ToolBar.qml
```diff
@@ -77,6 +77,7 @@ PanelBackground {
         spacing: Theme.paddingLarge
         anchors.verticalCenter: parent.verticalCenter
```
pvuorela: This is perhaps now unnecessary if height matches?

pvuorela:
For search item background, I'd maybe just remove it. Also to search opened state has strange fulll width background and icon colored.

plugin/ToolBar.qml
dcaliste: I agree, removed.

dcaliste:
> For search item background, I'd maybe just remove it.

Ok, done. At the beginning, I try to make SearchBarItem to inherit from IconButton, but since it has a lot more when expanded, I simply transform the BackgroundItem into a plain Item and adjusted a bit the signals and properties. I think, it's safer, not to introduce regression like that, than trying to make a IconButton being able to cope with the additional TextField and close icon when expanded.

pvuorela:
LGTM, thanks!

PR #139, opened 2019-09-02
[sailfish-office] Increase the size of the page index in toolbar. Contributes to TJC#212568

dcaliste:
Following discussion in [TJC thread](https://together.jolla.com/question/212568/documents-misaligned-toolbar-on-xperia-x/), I propose to expand a bit the size of the page index in PDF toolbar. For documents with many pages (like more than 100), the current label is shrinking to fit into a itemSizeSmall width.

I propose to expand this to itemSizeMedium, and to avoid to have wider space for documents with few pages (< 10), I'm changing a bit the width calculation logic to (label.implicitWidth + padding) limited by itemSizeMedium.

### DIFF
--- a/plugin/IndexButton.qml
+++ b/plugin/IndexButton.qml
@@ -29,12 +29,13 @@ MouseArea {
 
     enabled: count > 1 && allowed
     opacity: count > 0 && allowed ? (count > 1 ? 1.0 : Theme.opacityHigh) : 0.0
-    width: Theme.itemSizeSmall
+    width: Math.min(Theme.itemSizeMedium, label.implicitWidth + Theme.paddingSmall)
     height: parent.height
 
     Label {
+        id: label
         anchors.centerIn: parent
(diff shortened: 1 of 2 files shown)


dcaliste:
@pvuorela or @jpetrell if you have time to give feedback. I'm keeping the limitation of the label width, as done before, even if the toolbar is now more consise in width with icons more packed.

Besides, I've checked that when the back button is visible, the new index page width is not putting things out of the screen. Doing this, I've noticed that the back button was not properly vertically aligned. I've pushed another commit on top.

dcaliste:
@pvuorela did you have time to look at this one? Discard the question if you're busy!

pvuorela:
Looks a lot better now, thanks. LGTM. The page number label could increase width even more if needed, but that could then require more advanced layouting, should be good enough for most cases already.

dcaliste:
Great, thank you @pvuorela

PR #140, opened 2019-09-18
[sailfish-office] Correct russian charset for TXT files. Fixes JB#47341

mphilippov-omp:
Added detecting original file charset with uchardet library
Added opening text file with the original encoding

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,22 @@ find_package(Qt5LinguistTools REQUIRED)
 find_package(QtSparql REQUIRED)
 find_package(Booster REQUIRED)
 
+project(sailfish-office)
+
+cmake_minimum_required(VERSION 2.8.6)
+
+set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
+set(CMAKE_AUTOMOC TRUE)
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+find_package(Qt5Core REQUIRED)
+find_package(Qt5Quick REQUIRED)
+find_package(Qt5Gui REQUIRED)
+find_package(Qt5DBus REQUIRED)
+find_package(Qt5LinguistTools REQUIRED)
+find_package(QtSparql REQUIRED)
+find_package(Booster REQUIRED)
+
 include(cmake/QtTranslationWithID.cmake)
 
 include_directories(
@@ -61,7 +77,7 @@ create_translation(engen_qm_file ${CMAKE_BINARY_DIR}/sailfish-office.ts ${sailfi
 
 add_executable(sailfish-office ${sailfishoffice_SRCS} ${engen_qm_file})
 qt5_use_modules(sailfish-office Widgets Quick DBus)
-target_link_libraries(sailfish-office stdc++ ${QT_LIBRARIES} ${BOOSTER_LIBRARY} ${QTSPARQL_LIBRARY})
+target_link_libraries(sailfish-office stdc++ uchardet ${QT_LIBRARIES} ${BOOSTER_LIBRARY} ${QTSPARQL_LIBRARY})
 
 install(TARGETS sailfish-office DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
 install(FILES
@@ -75,4 +91,3 @@ install(FILES ${CMAKE_BINARY_DIR}/sailfish-office.ts DESTINATION ${CMAKE_INSTALL
 install(FILES org.sailfish.office.service DESTINATION ${CMAKE_INSTALL_PREFIX}/share/dbus-1/services)
 install(FILES org.sailfish.office.xml DESTINATION ${CMAKE_INSTALL_PREFIX}/share/dbus-1/interfaces)
 install(FILES sailfish-office.privileges DESTINATION ${CMAKE_INSTALL_PREFIX}/share/mapplauncherd/privileges.d)
-
--- a/plugin/plaintextmodel.cpp
+++ b/plugin/plaintextmodel.cpp
@@ -24,16 +24,22 @@
 #include <QThread>
 #include <silicatheme.h>
 
+#include <unicode/ucsdet.h>
+
+#include <uchardet/uchardet.h>
+
 class PlainTextModel::FileData : public QSharedData
 {
 public:
     FileData(PlainTextModel *model)
         : fileName(model->m_file.fileName())
         , model(model)
+        , codec(nullptr)
     {
     }
 
     QString fileName;
+    QTextCodec *codec;
     PlainTextModel *model = nullptr;
 };
 
@@ -116,10 +122,18 @@ void PlainTextModel::setSource(const QUrl &source)
                 qmlInfo(this) << "Can't open " << m_source << ": " << m_file.errorString();
                 m_status = Error;
             } else {
+                qint64 limit = 5*1024*1024;
+                qint64 limitSize = qMin(m_file.size(), limit);
+
+                uchar * res = m_file.map(0, limitSize);
+                QTextCodec * detectedCodec = detectCodec(reinterpret_cast<const char *>(res), limitSize);
+
                 m_textStream.setDevice(&m_file);
+                m_textStream.setCodec(detectedCodec);
 
                 if (m_file.size() > maximumSynchronousSize) {
                     m_fileData = new FileData(this);
+                    m_fileData->codec = detectedCodec;
 
                     Reader *const reader = new Reader(m_fileData);
                     reader->start();
@@ -264,6 +278,25 @@ bool PlainTextModel::readLines(
     return false;
 }
 
+QTextCodec *PlainTextModel::detectCodec(const char *encodedCharacters, qint64 length)
+{
+    QTextCodec *codec = QTextCodec::codecForLocale();
+
+    uchardet_t cd = uchardet_new();
+    int res = uchardet_handle_data(cd, encodedCharacters, length);
+    if (!res) {
+        uchardet_data_end(cd);
+
+        const char * name = uchardet_get_charset(cd);
+        codec = QTextCodec::codecForName(name);
+    }
+
+    uchardet_data_end(cd);
+    uchardet_delete(cd);
+
+    return codec;
+}
+
 PlainTextModel::Reader::Reader(const QExplicitlySharedDataPointer<PlainTextModel::FileData> &fileData)
     : m_fileData(fileData)
 {
@@ -280,6 +313,7 @@ void PlainTextModel::Reader::run()
         QCoreApplication::postEvent(this, event);
     } else {
         QTextStream stream(&file);
+        stream.setCodec(m_fileData->codec);
 
         bool atEnd = false;
         for (bool cache = true; m_fileData->model && !atEnd; cache = false) {
--- a/rpm/sailfish-office.spec
+++ b/rpm/sailfish-office.spec
@@ -17,7 +17,7 @@ BuildRequires: cmake
 BuildRequires: qt5-qttools-linguist
 Requires: calligra-components >= 2.7.9+git4
 Requires: calligra-filters >= 2.7.9+git4
-Requires: sailfishsilica-qt5 >= 1.1.46
+Requires: sailfishsilica-qt5 >= 1.1.53
 Requires: sailfish-components-accounts-qt5
 Requires: sailfish-components-textlinking
 Requires: libqt5sparql-tracker
--- a/plugin/CMakeLists.txt
+++ b/plugin/CMakeLists.txt
@@ -13,7 +13,7 @@ set(plugin_SRCS
 
 add_library(sailfishofficeplugin MODULE ${plugin_SRCS})
 qt5_use_modules(sailfishofficeplugin Widgets Quick)
-target_link_libraries(sailfishofficeplugin stdc++ ${QT_LIBRARIES} ${SAILFISHSILICA_LIBRARIES})
+target_link_libraries(sailfishofficeplugin stdc++ uchardet ${QT_LIBRARIES} ${SAILFISHSILICA_LIBRARIES})
 
 install(TARGETS sailfishofficeplugin DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/qt5/qml/Sailfish/Office)
 
--- a/plugin/plaintextmodel.h
+++ b/plugin/plaintextmodel.h
@@ -95,6 +95,7 @@ private:
     inline QString *lineAt(int index);
     inline static bool readLines(
             QTextStream *stream, std::vector<Line> *lines, std::vector<QString> *cache);
+    static QTextCodec * detectCodec(const char *encodedCharacters, qint64 length);
 
     std::vector<Line> m_lines;
     QCache<int, QString> m_textCache { maximumCost + (maximumCost / 20) };


pvuorela:
Includes binaries and headers, that's a no-no. Should either package the dependency separately, to separate repository, or then include a git submodule to be built here.

plugin/plaintextmodel.cpp
```diff
@@ -117,6 +121,11 @@ void PlainTextModel::setSource(const QUrl &source)
                 m_status = Error;
             } else {
                 m_textStream.setDevice(&m_file);
+                QByteArray ba = m_file.readAll();
+
+                QString name = detect(ba.data(), ba.size())->name();
+                m_textStream.setCodec(QTextCodec::codecForName( name.toStdString().data() ));
```
pvuorela: Uhm, so the method returns qtextcodec pointer, which is converted to string which is converted to qbytearray which is converted back to different qtextcodec pointer?

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +273,26 @@ bool PlainTextModel::readLines(
...
 }
 
+QTextCodec *PlainTextModel::detect(const char *encodedCharacters, int length)
+{
+    QTextCodec *codec = QTextCodec::codecForLocale();
+
+    uchardet_t cd = uchardet_new();
+    int res = uchardet_handle_data(cd, encodedCharacters, length);
+    if(!res)
+    {
```
pvuorela: Follow style. If () {

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +273,26 @@ bool PlainTextModel::readLines(
...
+        uchardet_data_end(cd);
+
+        const char * name = uchardet_get_charset(cd);
+        codec = QTextCodec::codecForName(name);
```
pvuorela: Need to check, but hope the name strings are compatible.

plugin/plaintextmodel.cpp
```diff
@@ -280,6 +309,10 @@ void PlainTextModel::Reader::run()
         QCoreApplication::postEvent(this, event);
     } else {
         QTextStream stream(&file);
+        QByteArray ba = file.readAll();
+        QString name = detect(ba.data(), ba.size())->name();
+        stream.setCodec(QTextCodec::codecForName( name.toStdString().data() ));
+        stream.seek(0);
```
pvuorela: Detection done twice?

mphilippov-omp:
> Includes binaries and headers, that's a no-no. Should either package the dependency separately, to separate repository, or then include a git submodule to be built here.

Well, this library has a source code in github, but there is no spec file. Thats why I included binaries and header file in a project

plugin/plaintextmodel.cpp
```diff
@@ -117,6 +121,11 @@ void PlainTextModel::setSource(const QUrl &source)
                 m_status = Error;
             } else {
                 m_textStream.setDevice(&m_file);
+                QByteArray ba = m_file.readAll();
```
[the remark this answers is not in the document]
mphilippov-omp: I think it is better to detect encoding of the whole file content.

plugin/plaintextmodel.cpp
mphilippov-omp: Well, I tried ICU at first, but it detects non-latin text with errors. Btw there is the same case in jolla-notes. Thats why I had to find another charset detector. I

> This isn't really just russian charset, but any non-utf8. Furthermore russian per se should already work, provided it's utf-8, right?

Yes. Russian text with UTF-8 encoding works, but if we have file with russian text with Windows-1251 encoding, it opend with non-human-readable text. The same situation is also with arabic text.

plugin/plaintextmodel.cpp
mphilippov-omp: Yes, It's a mistake.

plugin/plaintextmodel.cpp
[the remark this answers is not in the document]
pvuorela: Why? Some character around 5MB position changing the result?

plugin/plaintextmodel.cpp
mphilippov-omp: What do you mean?

pvuorela:
> Well, this library has a source code in github, but there is no spec file. Thats why I included binaries and header file in a project

You can test that way, sure, but as such this will not be merged.

plugin/plaintextmodel.cpp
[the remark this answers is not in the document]
mphilippov-omp: I think you are right

plugin/plaintextmodel.cpp
pvuorela: There are two places where detect() is called.

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +273,25 @@ bool PlainTextModel::readLines(
     return false;
 }
 
+QTextCodec *PlainTextModel::detect(const char *encodedCharacters, int length)
```
pvuorela: Detect is slightly vague as name, something like detectCodec might be better.

pvuorela:
> Ok. I'll add the git submodule of this library. Will it be ok?

Depends how it goes. Submodule ok, but code from such should be built or linked into office's plugin. No extra libraries (or headers etc) installed to /usr/lib.

plugin/plaintextmodel.cpp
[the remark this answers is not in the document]
adenexter: You could mmap the file rather than streaming the data, that should ensure there's enough data available without potentially allocating megabytes on the heap to scan the first handful of characters.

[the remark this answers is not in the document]
mphilippov-omp: > You could mmap the file rather than streaming the data, that should ensure there's enough data available without potentially allocating megabytes on the heap to scan the first handful of characters.

I need the streaming data to establish the right codec

mphilippov-omp:
> > Ok. I'll add the git submodule of this library. Will it be ok?
> 
> Depends how it goes. Submodule ok, but code from such should be built or linked into office's plugin. No extra libraries (or headers etc) installed to /usr/lib.

Actually it is a separate library and it is better to use it like a separate library. It will be useful in sailfish-office to detect encodings of TXT files and in Calligra for CSV files.

pvuorela:
A separate library makes sense if there's need to implement detection in multiple places, but in that case the packaging needs to go into its own repository. We cannot globally ship random libraries with applications.

plugin/plaintextmodel.cpp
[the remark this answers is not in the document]
dcaliste: Mapping the data, instead of reading them, will allow to use the data (in the detection library for instance) without allocating the memory: https://doc.qt.io/qt-5/qfiledevice.html#map From my understanding, in that case, the actual file for storage is used when accessing the data and not a copy of the data into memory.

mphilippov-omp:
> A separate library makes sense if there's need to implement detection in multiple places, but in that case the packaging needs to go into its own repository. We cannot globally ship random libraries with applications.

Ok. I'll create a repository. Where will it be better? In "sailfishos" or anywhere else?

pvuorela:
Yea, sailfish gitlab should be good. Just create a personal repo there, can then pick up from there to make a global one.

But but, before going too deep there, what exactly is the problem with icu detector? Some specific encoding? A little sceptical if you've been checking all possible non-latin ones :) New packages need to be also considered how much need there is, so if icu only has minor problems it could be a reason not to add more.

mphilippov-omp:
> Yea, sailfish gitlab should be good. Just create a personal repo there, can then pick up from there to make a global one.
> 
> But but, before going too deep there, what exactly is the problem with icu detector? Some specific encoding? A little sceptical if you've been checking all possible non-latin ones :) New packages need to be also considered how much need there is, so if icu only has minor problems it could be a reason not to add more.

What about ICU. I tried to test ICU on different charsets and in 4 times of 8 it gave mistakes. For example Russian text with cp1251 encoding this library detected as ISO-8859-6 (Arabic). Russian text with KOI8-R and IBM866 was also detected with wrong charset

mphilippov-omp:
We had this problem with jolla-notes where ICU library was also used

mphilippov-omp:
> Yea, sailfish gitlab should be good. Just create a personal repo there, can then pick up from there to make a global one.

Ok, I've created personal repository on gitlab, but how can I create the global one? Do I have any permissions to do that?

pvuorela:
> Ok, I've created personal repository on gitlab, but how can I create the global one? Do I have any permissions to do that?

Just link it here so we can review it.

mphilippov-omp:
> > Ok, I've created personal repository on gitlab, but how can I create the global one? Do I have any permissions to do that?
> 
> Just link it here so we can review it.

Here it is
https://git.sailfishos.org/MIkhail/uchardet

pvuorela:
> https://git.sailfishos.org/MIkhail/uchardet

Thanks. Some initial observations:
- %files commonly as last section
- Should split into subpackages, -devel separately as headers shouldn't be installed on devices, /usr/bin stuff into some tools package for the same reason, or alternatively dropped altogether. Man pages we don't really need, so can as well %exclude.
- "cd uchardet", this shouldn't be done. Instead include /uchardet or /%{name} on the %setup line.
- should do ldconfig after installation and removal.

mphilippov-omp:
> > https://git.sailfishos.org/MIkhail/uchardet
> 
> Thanks. Some initial observations:
> 
>     * %files commonly as last section
> 
>     * Should split into subpackages, -devel separately as headers shouldn't be installed on devices, /usr/bin stuff into some tools package for the same reason, or alternatively dropped altogether. Man pages we don't really need, so can as well %exclude.
> 
>     * "cd uchardet", this shouldn't be done. Instead include /uchardet or /%{name} on the %setup line.
> 
>     * should do ldconfig after installation and removal.

Done

plugin/plaintextmodel.cpp
mphilippov-omp: Done

plugin/plaintextmodel.cpp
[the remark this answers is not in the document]
mphilippov-omp: Done

plugin/plaintextmodel.cpp
mphilippov-omp: Done

plugin/plaintextmodel.cpp
mphilippov-omp: Done

mphilippov-omp:
Hello! I've  corrected all mistakes you've matched. Can you review, please?
________________________________________
От: Pekka Vuorela [notifications@github.com]
Отправлено: 19 сентября 2019 г. 15:10
Кому: sailfishos/sailfish-office
Копия: Mikhail Filippov; Author
Тема: Re: [sailfishos/sailfish-office] [sailfish-office] Correct russian charset for TXT files. Fixes JB#47341 (#140)

https://git.sailfishos.org/MIkhail/uchardet

Thanks. Some initial observations:

  *   %files commonly as last section
  *   Should split into subpackages, -devel separately as headers shouldn't be installed on devices, /usr/bin stuff into some tools package for the same reason, or alternatively dropped altogether. Man pages we don't really need, so can as well %exclude.
  *   "cd uchardet", this shouldn't be done. Instead include /uchardet or /%{name} on the %setup line.
  *   should do ldconfig after installation and removal.

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +276,25 @@ bool PlainTextModel::readLines(
     return false;
 }
 
+QTextCodec *PlainTextModel::detectCodec(const char *encodedCharacters, int length)
```
dcaliste: The size of a file returned by `size()` is `qint64` and the length parameter of uchardet_handle_data() is `size_t`. Even if we really don't want to pass gigabytes of data, it's better to respect 64bits length.

plugin/plaintextmodel.cpp
```diff
@@ -116,10 +122,16 @@ void PlainTextModel::setSource(const QUrl &source)
                 qmlInfo(this) << "Can't open " << m_source << ": " << m_file.errorString();
                 m_status = Error;
             } else {
+                uchar * res = m_file.map(0, m_file.size());
```
dcaliste: We should limit the size here I think, something like `min(m_file.size(), 5*1024*1024)` for instance.

CMakeLists.txt
```diff
@@ -61,7 +77,7 @@ create_translation(engen_qm_file ${CMAKE_BINARY_DIR}/sailfish-office.ts ${sailfi
 
 add_executable(sailfish-office ${sailfishoffice_SRCS} ${engen_qm_file})
 qt5_use_modules(sailfish-office Widgets Quick DBus)
-target_link_libraries(sailfish-office stdc++ ${QT_LIBRARIES} ${BOOSTER_LIBRARY} ${QTSPARQL_LIBRARY})
+target_link_libraries(sailfish-office stdc++ uchardet ${QT_LIBRARIES} ${BOOSTER_LIBRARY} ${QTSPARQL_LIBRARY})
```
dcaliste: Is uchardet really linked to the main sailfish-office library ? Is it not required just for the text plugin ?

plugin/plaintextmodel.cpp
```diff
@@ -24,16 +24,22 @@
 #include <QThread>
 #include <silicatheme.h>
 
+#include <unicode/ucsdet.h>
```
dcaliste: Is this a necessary include for uchardet ?

plugin/plaintextmodel.cpp
```diff
@@ -24,16 +24,22 @@
...
 public:
     FileData(PlainTextModel *model)
         : fileName(model->m_file.fileName())
         , model(model)
+        , codec(nullptr)
     {
     }
 
     QString fileName;
+    QTextCodec * codec;
```
dcaliste: Space after start and initialise to nullptr.

CMakeLists.txt
mphilippov-omp: No. I've answered, that it is a separated library and was not linked. I gave sources to Pekka for review

dcaliste: Yes, I know, I've read the history of the excahnge, my question is : is it necessary to link uchardet to sailfish-office library when it is used only by the text plugin ?

mphilippov-omp: Yes because it will be also used in calligra to find charset for csv files

dcaliste: I see, but excuse my insistance, should it be linked by the calligra library in that case (or by the calligra plugin if the charset detection is done before passing the data to calligra itself) ?

mphilippov-omp: Yes, it should. Previosly when a user opens a CSV file, a dialog (developed via QWidgets) appears where user should choose encoding and delimiter himself. 
Now we came to a conclusion with Pekka that this dialog should be removed and a csv file should be opened with an original encoding. Thats why uchardet detecting in calligra for csv files should be necessary.

plugin/plaintextmodel.cpp
mphilippov-omp: What do you mean?

plugin/plaintextmodel.cpp
mphilippov-omp: No, it isn't. I'll remove it

plugin/plaintextmodel.cpp
mphilippov-omp: You are right. I'll do it

CMakeLists.txt
dcaliste: I understand the necessity and concur with the requirement for uchardet, but in my opinion, this is the wrong place to link with uchardet for calligra. The link should be done in calligra project itself since it's where the actual code using uchardet will be.

From my understanding:
- Calligra is used via QML components in sailfish office, so it means that the libcalligracomponent.so is dlopened by the QML engine when a new calligra object is required from QML.
- This calligra component will use uchardet and thus requires to resolve some symbols like `uchardet_handle_data`. If libcalligracomponent.so is not dependent on uchardet, the resolving will work only if uchardet_handle_data is already in memory via a previous dlopen of another library. If you link it with sailfish-office this will actually be the case and everything will work.
- But if you dlopen libcalligracomponent.so from another software for any reason, it will fail with unresolved symbols because of the missing linkage with uchardet library.

That's why I'm suggesting not to link uchardet with sailfish-office library but with the library that is actually using the code and I don't see here that uchardet is used to decode CSV files.

plugin/plaintextmodel.cpp
dcaliste: I mean that if I have a 5GB file named foo.txt, this will crash the code, or have unexpected behaviour because `int length` cannot represent this actual size. That's why `size_t` exists in C and is used to represent file sizes. Qt is not using size_t when calling `size()` for reason I don't know, but they are using an equivalent type which is `qint64`. So in my opinion, you should declare length either `size_t` or `qint64` (maybe the later because the code is mostly using Qt) to be sure not to overflow integers.

CMakeLists.txt
mphilippov-omp: It's not used here to decode CSV files. What about QML...Actually QML part of sailfish-office doesn't know about Calligra a lot. We just set the source file and then wait until it'll be displayed. Sailfish-office doesn't know anything about the internal processes when we call "setSource" method in calligra. We need uchardet for the internal part.

What about using uchardet - actually it is linked in cmakelists in "CSV" folder and is used for csv

pvuorela: > Yes because it will be also used in calligra to find charset for csv files

This will be up to calligra folks to consider. May or may not happen.

In general Damien has a point, module (being it library or binary) that uses some other library should link against that, others don't need to care. If main application doesn't explicitly use uchardet, this should be dropped.

rpm/sailfish-office.spec
```diff
@@ -17,7 +17,7 @@ BuildRequires: cmake
 BuildRequires: qt5-qttools-linguist
 Requires: calligra-components >= 2.7.9+git4
 Requires: calligra-filters >= 2.7.9+git4
-Requires: sailfishsilica-qt5 >= 1.1.46
+Requires: sailfishsilica-qt5 >= 1.1.53
```
pvuorela: Doesn't look like belonging here. Rebase against master.

CMakeLists.txt
```diff
@@ -14,6 +14,22 @@ find_package(Qt5LinguistTools REQUIRED)
...
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+find_package(Qt5Core REQUIRED)
+find_package(Qt5Quick REQUIRED)
+find_package(Qt5Gui REQUIRED)
+find_package(Qt5DBus REQUIRED)
+find_package(Qt5LinguistTools REQUIRED)
+find_package(QtSparql REQUIRED)
+find_package(Booster REQUIRED)
+
```
pvuorela: Hmh?

PR #141, opened 2019-09-20
[sailfish-office] Don't use annotation boundary after removal.

dcaliste:
@pvuorela, sorry I was mistaken with ec21e4eba266f0ad23417970bf2b39c0f4f4211d. It's not fixing the crash issue, I think I was mislead by the fact that it was working in ```gdb``` where I tested. Anyway, the (hopefully) real fix is coming with this PR. Maybe an upgrade in Poppler make that removed annotation are deleted soon than before and we cannot the pointer anymore afterwards. So, I'm saving the bounding box for the signal before removing the annotation.


pvuorela:
Seems a sensical change. Thanks.

PR #142, opened 2019-09-20
[sailfish-office] Correct russian charset for TXT files. Fixes JB#47341

mphilippov-omp:


plugin/CMakeLists.txt
```diff
@@ -9,6 +9,56 @@ set(plugin_SRCS
     fileinfo.cpp
     plaintextmodel.cpp
     sailfishofficeplugin.cpp
+
```
dcaliste: Just a suggestion, please ignore it if too complicated. Since this uchardet has a build system itself based on cmake, would it be nicer to use it ? Like that when uchardet is updated in the future, we don't need to update our own CMakeLists files.

Maybe there is a cmake way top compile sub cmake project in a project, or it can be done in the spec file, starting building it before the sailfish office itself.

mphilippov-omp: Thank you for suggestion, but I think, that it'll be rather complicated now because uchardet's cmake builds it as separate utility with it's own libraries and we will have some troubles with applying them while building sailfish-office itself. Actually in this case it will be better to use it as a separate library as I told you earlier (IMHO).

P.S. Thank you for an attention to uchardet library, but can you review the whole code in the project?

mphilippov-omp:
can you make a review please?
________________________________________
От: Damien Caliste [notifications@github.com]
Отправлено: 21 сентября 2019 г. 11:28
Кому: sailfishos/sailfish-office
Копия: Mikhail Filippov; Author
Тема: Re: [sailfishos/sailfish-office] [sailfish-office] Correct russian charset for TXT files. Fixes JB#47341 (#142)

@dcaliste commented on this pull request.

________________________________

In plugin/CMakeLists.txt<https://github.com/sailfishos/sailfish-office/pull/142#discussion_r326852507>:

> @@ -9,6 +9,56 @@ set(plugin_SRCS
     fileinfo.cpp
     plaintextmodel.cpp
     sailfishofficeplugin.cpp
+


Just a suggestion, please ignore it if too complicated. Since this uchardet has a build system itself based on cmake, would it be nicer to use it ? Like that when uchardet is updated in the future, we don't need to update our own CMakeLists files.

Maybe there is a cmake way top compile sub cmake project in a project, or it can be done in the spec file, starting building it before the sailfish office itself.

dcaliste:
> can you make a review please?

Sorry, I'm not working for Jolla. I took the initiative to give my opinion, because, well, I already contributed  to some extend to the code. You addressed my concerns in the previous MR, so mainly this one, I agree with, thus myself having no new remark. But I cannot give you a green flag, you should wait for @pvuorela or @jpetrell.

plugin/CMakeLists.txt
pvuorela: I'm also a little concerned about replicating the file list here. Updating the checkout gets more complicated if the list here needs to be modified too, let alone if there starts to be compilation options. Ideally it could be built as a static lib by its own build system and just linked here. Or maybe we should just have that separate lib? Though also pondering if the size increase is worth the better detection over icu, which also could get better with updates.

For another thing noticed that the checkout is not a released version but rather the current master. Is there anything needed not included in latest 0.0.6?

plugin/plaintextmodel.cpp
```diff
@@ -24,6 +24,8 @@
 #include <QThread>
 #include <silicatheme.h>
 
+#include "uchardet/src/uchardet.h"
```
pvuorela: Would be nice if c++ side was agnostic of the library placement, i.e. include as in how the lib is commonly used, build system setting up the flags so that proper directions are used for lookup.

plugin/CMakeLists.txt
mphilippov-omp: > I'm also a little concerned about replicating the file list here. Updating the checkout gets more complicated if the list here needs to be modified too, let alone if there starts to be compilation options. Ideally it could be built as a static lib by its own build system and just linked here. Or maybe we should just have that separate lib? Though also pondering if the size increase is worth the better detection over icu, which also could get better with updates.
> 
What about ICU, we had bugs with non-UTF encoding for russian text. For example if we receive a txt file with cp1251 encoding by email we will not read it correctly. I've told it to you in previous PR.
IMHO it will be better used as a separate library with its own RPM and linked in a project, as I insisted previously

> For another thing noticed that the checkout is not a released version but rather the current master. Is there anything needed not included in latest 0.0.6?

what do you mean?

plugin/plaintextmodel.cpp
mphilippov-omp: > Would be nice if c++ side was agnostic of the library placement, i.e. include as in how the lib is commonly used, build system setting up the flags so that proper directions are used for lookup.
Well actually you are right

plugin/CMakeLists.txt
dcaliste: > IMHO it will be better used as a separate library with its own RPM and linked in a project, as I insisted previously

If uchardet is only used here, it makes sense not to package it separately. I've done the same for various KF5 packages for Calligra (see the calligra spec file). I think you can follow the same road here:
- compile the uchardet in the `%build` section of the spec file. Use a `-DBUILD_STATIC_LIBS=ON` and a `-DBUILD_SHARED_LIBS=OFF` flags for the cmake command. Install it in a tmp directory.
- when calling cmake for sailfish-office in the spec file, set the `CMAKE_PREFIX_PATH` variable to your tmp directory, so it can find the uchardet package. Modify the CMakeLists.txt files of plugin to look for the uchardet package and link to it accordingly.

Like that, you can use "normal" include for uchardet, like `#include <uchardet/uchardet.h>`, and if once it is decided to move uchardet in a dedicated package, just the spec file will have to be modified.

If you don't see how to do this, or have issue, I can set it up for you early next week.

mphilippov-omp:
It is also used in calligra to detect encoding for CSV files. Not just here.
See MR https://git.merproject.org/mer-core/calligra/merge_requests/14

And yes, I agree that it will be better to use include like #include <uchardet/uchardet.h>
________________________________________
От: Damien Caliste [notifications@github.com]
Отправлено: 25 сентября 2019 г. 10:28
Кому: sailfishos/sailfish-office
Копия: Mikhail Filippov; Comment
Тема: Re: [sailfishos/sailfish-office] [sailfish-office] Correct russian charset for TXT files. Fixes JB#47341 (#142)

@dcaliste commented on this pull request.

________________________________

In plugin/CMakeLists.txt<https://github.com/sailfishos/sailfish-office/pull/142#discussion_r327969344>:

> @@ -9,6 +9,56 @@ set(plugin_SRCS
     fileinfo.cpp
     plaintextmodel.cpp
     sailfishofficeplugin.cpp
+


IMHO it will be better used as a separate library with its own RPM and linked in a project, as I insisted previously

If uchardet is only used here, it makes sense not to package it separately. I've done the same for various KF5 packages for Calligra (see the calligra spec file). I think you can follow the same road here:

  *   compile the uchardet in the %build section of the spec file. Use a -DBUILD_STATIC_LIBS=ON and a -DBUILD_SHARED_LIBS=OFF flags for the cmake command. Install it in a tmp directory.
  *   when calling cmake for sailfish-office in the spec file, set the CMAKE_PREFIX_PATH variable to your tmp directory, so it can find the uchardet package. Modify the CMakeLists.txt files of plugin to look for the uchardet package and link to it accordingly.

Like that, you can use "normal" include for uchardet, like #include <uchardet/uchardet.h>, and if once it is decided to move uchardet in a dedicated package, just the spec file will have to be modified.

If you don't see how to do this, or have issue, I can set it up for you early next week.

monich:
To me is sounds like this uchardet thing needs to be packaged separately (it's a shared library, isn't it).

mphilippov-omp:
> To me is sounds like this uchardet thing needs to be packaged separately (it's a shared library, isn't it).

Actually yes

monich:
> > To me is sounds like this uchardet thing needs to be packaged separately (it's a shared library, isn't it).
> 
> Actually yes

Package it then so that it builds in OBS :) There are plenty of examples at https://git.sailfishos.org/mer-core how to do it right. I _think_ it should be at https://git.sailfishos.org, although I'm not 100% sure about that.

mphilippov-omp:
Well, I've changed Uchardet to ICU charset detection. Can you review my code please?

rpm/sailfish-office.spec
```diff
@@ -15,6 +15,7 @@ BuildRequires: poppler-qt5-devel poppler-qt5 poppler-devel poppler
 BuildRequires: mapplauncherd-qt5-devel
 BuildRequires: cmake
 BuildRequires: qt5-qttools-linguist
+BuildRequires: libicu
```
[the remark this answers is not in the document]
mphilippov-omp: Done

plugin/plaintextmodel.cpp
```diff
@@ -116,10 +119,18 @@ void PlainTextModel::setSource(const QUrl &source)
                 qmlInfo(this) << "Can't open " << m_source << ": " << m_file.errorString();
                 m_status = Error;
             } else {
+                qint64 limit = 5*1024*1024;
+                qint64 limitSize = qMin(m_file.size(), limit);
+
+                uchar * res = m_file.map(0, limitSize);
```
monich: Where does it get unmapped?

mphilippov-omp: > Where does it get unmapped?

nowhere. As @dcaliste told, mapping is better for reading big data than file.readAll()

mphilippov-omp: Anyway you are right. I'll do it

mphilippov-omp: Done

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +277,47 @@ bool PlainTextModel::readLines(
...
+QTextCodec *PlainTextModel::detectCodec(const char *encodedCharacters, qint64 length)
+{
+    QLocale locale;
+    QTextCodec *codec = QTextCodec::codecForLocale();
+
+    UErrorCode error = U_ZERO_ERROR;
+    if (UCharsetDetector * const detector = ucsdet_open(&error)) {
+        ucsdet_setText(detector, encodedCharacters, length, &error);
+
+
```
monich: ETOOMANYEMPTYLINES :)

mphilippov-omp: Yes, though these lines of code need separation, there are so many empty lines. I've done it.

plugin/plaintextmodel.cpp
mphilippov-omp: Ok

dcaliste: > I would also prefer to use read-only mapping

I think it's already mapped read-only because the `m_file` member has been set read-only itself around line 115.

rpm/sailfish-office.spec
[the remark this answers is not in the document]
dcaliste: Since using pkg_config, something like `pkg_check_modules(ICU icu-i18n REQUIRED)` should be added in the `CMakeLists.txt` in my opinion. Then add the include and libs to the required variables. It is compiling by luck that unicode/ucsdet.h is in the path and that libicu is already in memory, because the plugin is not linked against.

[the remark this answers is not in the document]
dcaliste: Something like:
```
find_package(PkgConfig REQUIRED)
pkg_check_modules(SAILFISHSILICA sailfishsilica REQUIRED)
pkg_check_modules(ICU icu-i18n REQUIRED)

include_directories(
    ${SAILFISHSILICA_INCLUDE_DIRS}
    ${ICU_INCLUDE_DIRS}
)

set(plugin_SRCS
    fileinfo.cpp
    plaintextmodel.cpp
    sailfishofficeplugin.cpp
)

add_library(sailfishofficeplugin MODULE ${plugin_SRCS})
qt5_use_modules(sailfishofficeplugin Widgets Quick)
target_link_libraries(sailfishofficeplugin stdc++ ${QT_LIBRARIES} ${SAILFISHSILICA_LIBRARIES} ${ICU_LIBRARIES})
```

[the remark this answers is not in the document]
mphilippov-omp: Done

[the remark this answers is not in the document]
dcaliste: Thanks.

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +277,46 @@ bool PlainTextModel::readLines(
...
+        int32_t count = 0;
+        if (U_FAILURE(error)) {
+            qWarning() << "Unable to detect text encoding" << u_errorName(error);
+        } else if (const UCharsetMatch ** const matches = ucsdet_detectAll(detector, &count, &error)) {
+            QTextCodec *bestMatch = nullptr;
+
+            for (int32_t i = 0; i < count; ++i) {
+                if (QTextCodec *detectedCodec = QTextCodec::codecForName(ucsdet_getName(matches[i], &error))) {
+                    if (detectedCodec->name() == codec->name() || locale.name().left(locale.name().indexOf("_")) == QString(ucsdet_getLanguage(matches[i], &error))) {
+                        bestMatch = detectedCodec;
```
pvuorela: As detected codecs are ordered as best ones first, this will in effect choose the worst one that matches language? Or if there's a language based codec with high confidence, this will override it with the default utf-8 if found later in the array.

Maybe could get better results if just giving language based detection some confidence boost (say, multiply with 1.5 or something), and picking the highest resulting value?

Could also split that long lines.

mphilippov-omp: I've done it. Please, take a look at it.

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +277,46 @@ bool PlainTextModel::readLines(
...
+
+        int32_t count = 0;
+        if (U_FAILURE(error)) {
+            qWarning() << "Unable to detect text encoding" << u_errorName(error);
+        } else if (const UCharsetMatch ** const matches = ucsdet_detectAll(detector, &count, &error)) {
+            QTextCodec *bestMatch = nullptr;
+            QString localeName = locale.name();
+
+            for (int32_t i = 0; i < count; ++i) {
+                if(QString(ucsdet_getLanguage(matches[i], &error)) == localeName.left(localeName.indexOf("_"))) {
```
pvuorela: if ()

mphilippov-omp: Oops, sorry. Corrected

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +277,46 @@ bool PlainTextModel::readLines(
...
+
+            for (int32_t i = 0; i < count; ++i) {
+                if (QString(ucsdet_getLanguage(matches[i], &error)) == localeName.left(localeName.indexOf("_"))) {
+                    bestMatch = QTextCodec::codecForName(ucsdet_getName(matches[i], &error));
+                    break;
+                }
+            }
+
+            if (bestMatch) {
+                codec = bestMatch;
```
pvuorela: Now with utf-8 confidence 100 and language specific locale with confidence 1 will pick the latter, which doesn't feel like the correct thing to do.

mphilippov-omp: Well it seems that previous variant was better because because it made a choice according to the codec and to the language. So what was wrong?

mphilippov-omp: Corrected

pvuorela: Ok, now it handles utf-8 and language specific whichever just has better confidence.

But then again latin1 confidence 100, utf8 or language specific confidence 1 will pick the latter one. Quite heavy bias?

mphilippov-omp: Actually the language has better priority because the language of the interface is the native user's language or the language the user can speak anyway. So IMHO language should have better priority then the codec.

mphilippov-omp: So I don't understand you. What do you mean?

mphilippov-omp: IMHO it would be better to compare with the language (and give the best codec of the locale language)or give just the most confident codec otherwise, because, as I told, the most confident codec could be wrong, so it would be better to mention language

plugin/plaintextmodel.cpp
mphilippov-omp: > Maybe could get better results if just giving language based detection some confidence boost (say, multiply with 1.5 or something), and picking the highest resulting value?
> 

What do you mean?

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +277,49 @@ bool PlainTextModel::readLines(
...
+        int32_t count = 0;
+        if (U_FAILURE(error)) {
+            qWarning() << "Unable to detect text encoding" << u_errorName(error);
+        } else if (const UCharsetMatch ** const matches = ucsdet_detectAll(detector, &count, &error)) {
+            QTextCodec *bestMatch = nullptr;
+            QString localeName = locale.name();
+
+            for (int32_t i = 0; i < count; ++i) {
+                if (QTextCodec *detectedCodec = QTextCodec::codecForName(ucsdet_getName(matches[i], &error))) {
+                    if (detectedCodec->name() == codec->name() ||
```
monich: `QTextCodec::codecForName` can return NULL. Check the return value.

mphilippov-omp: I see

mphilippov-omp: I see it.

plugin/plaintextmodel.cpp
pvuorela: First utf8 vs latin1. The latter will be picked up only if detection says there's no chance the content is utf8. If if thinks there's some chance, if only that 1 confidence level versus 100, it will pick now utf8. I don't think there should be any bias between these two, whatever has more confidence should be just chosen. 

And why utf8 is because QTextCodec::codecForLocale() is in practice that for everything these days. In a way would be better not to use the whole method, but instead explicitly create utf8 codec to avoid misunderstandings.

Then the language, this will now in similar way pick encoding that matches the locale, now matter how low confidence there is for it if better guesses don't return the current language. Latin-1with 100 confidence level will still lose against codec with confidence of 1 but matching the language and locale.

Testing this with some files, the language is a bit random: fr, no, ro, tr, he. But the api docs do state that the language might end up guessed as side-effect.

pvuorela:
Commit message includes quite a bit of repeated lines and other noise.

plugin/plaintextmodel.cpp
mphilippov-omp: Done it

plugin/plaintextmodel.cpp
mphilippov-omp: Done

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +277,49 @@ bool PlainTextModel::readLines(
...
+        int32_t count = 0;
+        if (U_FAILURE(error)) {
+            qWarning() << "Unable to detect text encoding" << u_errorName(error);
+        } else if (const UCharsetMatch ** const matches = ucsdet_detectAll(detector, &count, &error)) {
+            QTextCodec *bestMatch = nullptr;
+            QString localeName = locale.name();
+
+            for (int32_t i = 0; i < count; ++i) {
+                if (QTextCodec *detectedCodec = QTextCodec::codecForName(ucsdet_getName(matches[i], &error))) {
+                    if (detectedCodec) {
```
pvuorela: This now checks null twice?

mphilippov-omp: Oops, sorry. Corrected

plugin/plaintextmodel.cpp
mphilippov-omp: line 297 checks if the value is NULL

plugin/plaintextmodel.cpp
pvuorela: Seeing this made the utf-8 now explicit, but all the other problems still remain. Or actually this now seems maybe even worse, e.g. never picks up latin1 no matter how high confidence there is for it? (At least unless the halfway random language guess happens to match current locale.)

mphilippov-omp: So if current language is English (e.g) Latin1 with confidence=100 can be detected according to the current language match

pvuorela: Sorry, missed these earlier comments, maybe best responding these ones:

> What do you mean? Just language based detection? And why should we multiply confidence which is already calculated and sorted?
> Why cannot I take the best encoding with the language of locale?
> And what should I do if there is no proper language? Perhaps I should take the most confident encoding, shouldn't I? Or the encoding of locale?

So suppose we have latin1 file and the detection returns codec/confidence pairs <latin1, 100>, [...something random], <windows-1251, 3>, <utf8, 1>. What we really want is that latin1, but the code here would return either windows-1251 or utf8 based on the language, both which have extremely low confidence.

However, if the list had <latin1, 100>, <windows-1251, 90> and the current language was russian, a little bias towards the latter could be justified. That's what I meant by multiplying. (Alternatively could be adding something too, but then we'd assume the returned range which didn't seem documented(?)). 90 * 1.5=>135 would lean towards 1251 on this case, but the earlier low confidence 3 * 1.5=> 4.5 would still be low confidence.

So all in all, I'm ok with just the simplistic solution of always just using the first codec returned, but if there's need to have bias towards current language, it should be a reasonable bias and not a forceful override.

Btw, the script of the current locale could probably be more correct way to bias, i.e. giving all the cyrillic languages boost for cyrillic codecs, but that might have a bit more complexity.

mphilippov-omp: Do you mean that it would be better to analyse codecs with confidence>100% after multiplying?

pvuorela: I'm not talking about any threshold or such. I'm not even sure are these percentages or just plain numbers.  I'm just talking about giving some results a confidence boost if the language matches, and after that picking the codec that has the highest confidence.

mphilippov-omp: So what is the purpose of this multiplication? The sequence of codecs won't change and I can find the best language-matched codec with the highest confidence without this multiplication.

mphilippov-omp: > However, if the list had <latin1, 100>, <windows-1251, 90> and the current language was russian, a little bias towards the latter could be justified. That's what I meant by multiplying. (Alternatively could be adding something too, but then we'd assume the returned range which didn't seem documented(?)). 90 * 1.5=>135 would lean towards 1251 on this case, but the earlier low confidence 3 * 1.5=> 4.5 would still be low confidence.
> 
Yes, Windows-1251 will go now with 135 confidence, but latin1 will go with 150 confidence. As you see, the sequence of encodings will not change, so what's the purpose? I can find the most confident language-matched encoding without this multiplication

pvuorela: Purpose is giving a confidence boost. Making that 90 cyrillic win the 100 latin1. You can find the highest matching already, but you won't have a bias towards anything on it.

Or if you're happy with just the highest confidence, just pick that and ditch all this language matching.

pvuorela: > Yes, Windows-1251 will go now with 135 confidence, but latin1 will go with 150 confidence.

So that idea was to give the boost _only_ if the language matches to current locale. If this detects latin1 file containing Russian text, surely it's then better. (Which shouldn't be even possible)

mphilippov-omp: Ok. That seems I understood you

mphilippov-omp: Done

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +277,65 @@ bool PlainTextModel::readLines(
...
+
+            if(bestMatchICU) {
+                QString localeName = locale.name();
+
+                int bestConfidenceLan = 0;
+                QTextCodec * bestMatchLan = nullptr;
+
+                for (int32_t i = 0; i < count; ++i) {
+                    if (localeName.left(localeName.indexOf("_")) == QString(ucsdet_getLanguage(matches[i], &error))) {
+                        if(QTextCodec * detectedCodec = QTextCodec::codecForName(ucsdet_getName(matches[i], &error))) {
```
pvuorela: if ()

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +277,65 @@ bool PlainTextModel::readLines(
...
+                            bestConfidenceLan = ucsdet_getConfidence(matches[i], &error) * 1.5;
+                            bestMatchLan = detectedCodec;
+
+                            break;
+                        }
+                    }
+                }
+
+                if (bestMatchLan) {
+                    bestConfidenceICU > bestConfidenceLan ? codec = bestMatchICU : codec = bestMatchLan;
```
pvuorela: Assignments in the operands is uncommon and also unnecessary. Could be just codec = bestConfidence_etc ? bestMatchICU : bestMatchLan;

pvuorela:
Please squash the commits.

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +277,65 @@ bool PlainTextModel::readLines(
...
+
+                    break;
+                }
+            }
+
+            if(bestMatchICU) {
+                QString localeName = locale.name();
+
+                int bestConfidenceLan = 0;
+                QTextCodec * bestMatchLan = nullptr;
```
pvuorela: I'd usually avoid abbreviations.

mphilippov-omp: What do you mean? There are no acronyms in the name

plugin/plaintextmodel.cpp
```diff
@@ -264,6 +277,65 @@ bool PlainTextModel::readLines(
...
+            for (int32_t i = 0; i < count; ++i) {
+                if (QTextCodec *detectedCodec = QTextCodec::codecForName(ucsdet_getName(matches[i], &error))) {
+                    bestConfidenceICU = ucsdet_getConfidence(matches[i], &error);
+                    bestMatchICU = detectedCodec;
+
+                    break;
+                }
+            }
+
+            if(bestMatchICU) {
```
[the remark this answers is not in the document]
mphilippov-omp: What do you mean? Acronym ICU?

plugin/plaintextmodel.cpp
mphilippov-omp: Ok. I'll correct

plugin/plaintextmodel.cpp
mphilippov-omp: Done

plugin/plaintextmodel.cpp
mphilippov-omp: Done

plugin/plaintextmodel.cpp
pvuorela: "lan" is not a word, or well could be used for local area network. -> bestMatchLanguage or something like that.

plugin/plaintextmodel.cpp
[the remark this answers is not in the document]
mphilippov-omp: Done

plugin/plaintextmodel.cpp
mphilippov-omp: Done

plugin/plaintextmodel.cpp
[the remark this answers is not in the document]
pvuorela: If ()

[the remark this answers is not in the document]
mphilippov-omp: Done

plugin/plaintextmodel.cpp
monich: `lang` could be an appropriate abbreviation for `language` but `lan` is definitely too short, I second that.

mphilippov-omp: Yes. I've already corrected it

pvuorela:
LGTM

PR #143, opened 2019-10-17
Ensure zooming is always possible, even on a large screen

llewelld:
The maximum zoom level was previously set to 10. On a large screen, a
small file could have a minimum zoom level larger than this. As a
result, the bounds check (min <= zoom <= max) in Calligra's
ViewController::zoom() method always failed, and the zoom became locked
to some initial value.

The overlay tool icons for Spreadsheets are shown only when the document
is at minimum zoom level. Since the minimum zoom level was never
reached, as a side effect the tool icons would never appear.

This change ensures that the maximum zoom level is always larger than
the minimum zoom level, ensuring that zooming is always possible and
that the tools can be made visible.

### DIFF
--- a/plugin/PDFDocumentPage.qml
+++ b/plugin/PDFDocumentPage.qml
@@ -1,5 +1,7 @@
 /*
- * Copyright (C) 2013-2014 Jolla Ltd.
+ * Copyright (c) 2013 - 2019 Jolla Ltd.
+ * Copyright (c) 2019 Open Mobile Platform LLC.
+ *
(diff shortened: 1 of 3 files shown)


pvuorela:
LGTM per se, but should the TextDocumentPage also get a similar thing? Both originally added by commit 12f3f92fac6c307

pvuorela:
LGTM

llewelld:
Thanks for pointing out the additional usage @pvuorela.

PR #146, opened 2019-10-30
[sailfish-office] Add name, date and type sort in file list. Fixes JB#47911

mphilippov-omp:
- Added a dialog which allows user to select sort parameter: name, type or date

- Added user, name and date sort for file list items

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ set(sailfishoffice_QML_SRCS
     qml/FileListPage.qml
     qml/Main.qml
     qml/CoverFileItem.qml
+    qml/SortTypeSelectionPage.qml
 )
 
 file(GLOB plugin_QML_SRCS plugin/*.qml)
--- a/models/documentlistmodel.cpp
+++ b/models/documentlistmodel.cpp
@@ -1,4 +1,5 @@
 /*
+ * Copyright (C) 2019 Open Mobile Platform LLC
  * Copyright (C) 2013-2014 Jolla Ltd.
  * Contact: Robin Burchell <robin.burchell@jolla.com>
  *
@@ -44,6 +45,7 @@ public:
         roles.insert(FileReadRole, "fileRead");
         roles.insert(FileMimeTypeRole, "fileMimeType");
         roles.insert(FileDocumentClass, "fileDocumentClass");
+        roles.insert(FileTypeAndNameRole, "fileTypeAndNameRole");
     }
     QList<DocumentListModelEntry> entries;
     QHash<int, QByteArray> roles;
@@ -69,7 +71,7 @@ QVariant DocumentListModel::data(const QModelIndex &index, int role) const
     case FilePathRole:
         return d->entries.at(index.row()).filePath;
     case FileTypeRole:
-        return QString().append(d->entries.at(index.row()).fileType).append(d->entries.at(index.row()).fileRead.toString(Qt::ISODate));
+        return d->entries.at(index.row()).fileType;
     case FileSizeRole:
         return d->entries.at(index.row()).fileSize;
     case FileReadRole:
@@ -78,6 +80,8 @@ QVariant DocumentListModel::data(const QModelIndex &index, int role) const
         return d->entries.at(index.row()).mimeType;
     case FileDocumentClass:
         return d->entries.at(index.row()).documentClass;
+    case FileTypeAndNameRole:
+        return QString().append(d->entries.at(index.row()).fileType).append(d->entries.at(index.row()).fileName);
     default:
         break;
     }
--- a/models/documentlistmodel.h
+++ b/models/documentlistmodel.h
@@ -1,4 +1,5 @@
 /*
+ * Copyright (C) 2019 Open Mobile Platform LLC
  * Copyright (C) 2013-2014 Jolla Ltd.
  * Contact: Robin Burchell <robin.burchell@jolla.com>
  *
@@ -46,7 +47,8 @@ public:
         FileSizeRole,
         FileReadRole,
         FileMimeTypeRole,
-        FileDocumentClass
+        FileDocumentClass,
+        FileTypeAndNameRole
     };
 
     DocumentListModel(QObject *parent = 0);
--- a/models/filtermodel.cpp
+++ b/models/filtermodel.cpp
@@ -1,4 +1,5 @@
 /*
+ * Copyright (C) 2019 Open Mobile Platform LLC
  * Copyright (C) 2015 Caliste Damien.
  * Contact: Damien Caliste <dcaliste@free.fr>
  *
@@ -24,6 +25,8 @@ FilterModel::FilterModel(QObject *parent)
     : QSortFilterProxyModel(parent)
 {
     this->setFilterRole(DocumentListModel::Roles::FileNameRole);
+    setSortParameter(SortParameter::Date);
+    setSortCaseSensitivity(Qt::CaseInsensitive);
 }
 
 FilterModel::~FilterModel()
@@ -33,6 +36,40 @@ FilterModel::~FilterModel()
 void FilterModel::setSourceModel(DocumentListModel *model)
 {
     QSortFilterProxyModel::setSourceModel(static_cast<QAbstractItemModel*>(model));
+}
+
+int FilterModel::sortParameter() const
+{
+    return m_sortParameter;
+}
+
+void FilterModel::setSortParameter(int sortParameter)
+{
+    if (m_sortParameter == sortParameter) {
+        return;
+    }
+    
+    m_sortParameter = sortParameter;
+    Qt::SortOrder order = Qt::AscendingOrder;
+
+    switch (m_sortParameter) {
+    case Name:
+        setSortRole(DocumentListModel::Roles::FileNameRole);
+        break;
+    case Type:
+        setSortRole(DocumentListModel::Roles::FileTypeAndNameRole);
+        break;
+    case Date:
+        setSortRole(DocumentListModel::Roles::FileReadRole);
+        order = Qt::DescendingOrder;
+        break;
+    default:
+        break;
+    }
+
+    emit sortParameterChanged();
+    
+    sort(0, order);
 }
 
 DocumentListModel* FilterModel::sourceModel() const
--- a/models/filtermodel.h
+++ b/models/filtermodel.h
@@ -1,4 +1,5 @@
 /*
+ * Copyright (C) 2019 Open Mobile Platform LLC
  * Copyright (C) 2015 Caliste Damien.
  * Contact: Damien Caliste <dcaliste@free.fr>
  *
@@ -27,18 +28,35 @@ class FilterModel : public QSortFilterProxyModel
 {
     Q_OBJECT
     Q_PROPERTY(DocumentListModel *sourceModel READ sourceModel WRITE setSourceModel NOTIFY sourceModelChanged)
+    Q_PROPERTY(int sortParameter READ sortParameter WRITE setSortParameter NOTIFY sortParameterChanged)
 public:
     FilterModel(QObject *parent = 0);
     ~FilterModel();
+    
+    enum SortParameter {
+        None,
+        Name,
+        Type,
+        Date
+    };
+
+    Q_ENUM(SortParameter)
 
 public:
     DocumentListModel* sourceModel() const;
+    
+    int sortParameter() const;
+    void setSortParameter(int sortParameter);
 
 public Q_SLOTS:
     void setSourceModel(DocumentListModel *model);
+        
+private:
+    int m_sortParameter;
 
 Q_SIGNALS:
     void sourceModelChanged();
+    void sortParameterChanged();
 };
 
 #endif // FILTERMODEL_H
--- a/qml/FileListPage.qml
+++ b/qml/FileListPage.qml
@@ -1,4 +1,5 @@
 /*
+ * Copyright (C) 2019 Open Mobile Platform LLC
  * Copyright (C) 2013-2014 Jolla Ltd.
  * Contact: Robin Burchell <robin.burchell@jolla.com>
  *
@@ -58,6 +59,21 @@ Page {
         if (!searchEnabled) {
             searchField.text = ""
         }
+    }
+    
+    function getSortParameterName(parameter) {
+        if (parameter === FilterModel.Name) {
+            //% "name"
+            return qsTrId("jolla-email-me-sort_by_name")
+        } else if (parameter === FilterModel.Type) {
+            //% "type"
+            return qsTrId("jolla-email-me-sort_by_type")
+        } else if (parameter === FilterModel.Date) {
+            //% "date"
+            return qsTrId("jolla-email-me-sort_by_date")
+        }
+
+        return ""
     }
 
     FilterModel {
@@ -145,6 +161,20 @@ Page {
                                              //% "Hide search"
                                            : qsTrId("sailfish-office-me-hide_search")
                 onClicked: page.searchEnabled = !page.searchEnabled
+            }
+            
+            MenuItem {
+                //% "Sort by: %1"
+                text: qsTrId("sailfish-office-me-sort_by").arg(getSortParameterName(filteredModel.sortParameter))
+                onClicked: {
+                    var obj = pageStack.animatorPush("SortTypeSelectionPage.qml")
+                    obj.pageCompleted.connect(function(page) {
+                        page.sortSelected.connect(function(sortParameter) {
+                            filteredModel.sortParameter = sortParameter
+                            pageStack.pop()
+                        })
+                    })
+                }
             }
         }
 
--- a/qml/SortTypeSelectionPage.qml
+++ b/qml/SortTypeSelectionPage.qml
@@ -1,0 +1,56 @@
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+import Sailfish.Office.Files 1.0
+
+Page {
+    id: root
+
+    signal sortSelected(int sortType)
+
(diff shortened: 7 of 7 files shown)


CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ set(sailfishoffice_QML_SRCS
     qml/FileListPage.qml
     qml/Main.qml
     qml/CoverFileItem.qml
+    qml.SortTypeSelectionDialog.qml
```
pvuorela: Bad character, doesn't even compile without fixing it.

models/filtermodel.cpp
```diff
@@ -35,6 +37,37 @@ void FilterModel::setSourceModel(DocumentListModel *model)
     QSortFilterProxyModel::setSourceModel(static_cast<QAbstractItemModel*>(model));
 }
 
+void FilterModel::sortByParameter()
+{
+    switch(m_sortParameter) {
```
pvuorela: switch () by convention.

models/filtermodel.cpp
```diff
@@ -35,6 +37,37 @@ void FilterModel::setSourceModel(DocumentListModel *model)
...
+
+    sort(0);
+}
+
+FilterModel::SortParameter FilterModel::sortParameter() const
+{
+    return m_sortParameter;
+}
+
+void FilterModel::setSortParameter(const SortParameter &sortParameter)
```
pvuorela: Pointless reference. It's good if copying is expensive, but this is just an int.

qml/FileListPage.qml
```diff
@@ -146,6 +147,12 @@ Page {
                                            : qsTrId("sailfish-office-me-hide_search")
                 onClicked: page.searchEnabled = !page.searchEnabled
             }
+            
+            MenuItem {
+                //% "Sort"
+                text: qsTrId("sailfish-office-me-sort")
+                onClicked: pageStack.push("SortTypeSelectionDialog.qml", {filterModel: filteredModel})
```
pvuorela: Could use animatorPush()

qml/SortTypeSelectionDialog.qml
```diff
@@ -0,0 +1,92 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import QtQuick 2.4
+import Sailfish.Silica 1.0
+import Sailfish.Office.Files 1.0
+
+Dialog {
```
pvuorela: Please implement this like jolla-email has: pulldown menu having "Sort by: <type" and then resulting page allowing to just click on sorting label. We don't really need separate acceptance for this and the switches look strange too.

qml/SortTypeSelectionDialog.qml
```diff
@@ -0,0 +1,92 @@
...
+
+    property FilterModel filterModel
```
pvuorela: This doesn't really need the model for anything else than the sort property. Could consider doing something like signal sortSelected(int) and in the FileList connect that to assigning the value to the actual model.

models/filtermodel.h
```diff
@@ -27,18 +28,37 @@ class FilterModel : public QSortFilterProxyModel
...
 public:
     DocumentListModel* sourceModel() const;
+    
+    SortParameter sortParameter() const;
+    void setSortParameter(const SortParameter &sortParameter);
 
 public Q_SLOTS:
     void setSourceModel(DocumentListModel *model);
+    
+    void sortByParameter();
```
pvuorela: Ends up similar, but this doesn't get used as slot, rather as Q_INVOKABLE. Then again is there reason to even have the method instead of just sorting when sort type is set?

CMakeLists.txt
mphilippov-omp: Sorry. Corrected

models/filtermodel.cpp
mphilippov-omp: Done

qml/FileListPage.qml
mphilippov-omp: Why is  it better to use instead of push()? Is there any issue in documentation about this method?

pvuorela: Starts push already before the page is created so the UI feels faster. Not documented at the moment due to being still slightly work in progress.

mphilippov-omp: Undocumented method usage is a bad way, IMHO. Because Lord knows what will happen to it further, how it will be change or whether it will exist

pvuorela: Besides it's already used on nearly all other places here. One instance of push() but i recall there was some specific detail related to it.

models/filtermodel.h
mphilippov-omp: Yes, I agree, that it is used mostly as Q_INVOKABLE and it is not correct to use it as a slot, but what do you mean under "is there reason to even have the method instead of just sorting when sort type is set"? I need it to match sort role and  call sort() of QSortFilterProxyModel to sort items

qml/SortTypeSelectionDialog.qml
mphilippov-omp: Well, ok. I'll try it

models/filtermodel.cpp
```diff
@@ -35,6 +37,37 @@ void FilterModel::setSourceModel(DocumentListModel *model)
...
+{
+    m_sortParameter = sortParameter;
```
dcaliste: We should emit signal only if value is actually changed and call sortByParameter() here, not from JS in QML.

models/filtermodel.cpp
```diff
@@ -22,6 +23,7 @@
 
 FilterModel::FilterModel(QObject *parent)
     : QSortFilterProxyModel(parent)
+    , m_sortParameter(SortParameter::None)
```
dcaliste: Currently, it's the case. But in fact, tracker seems to provide document sorted by date in fact. Can we reflect this by initialising this at sorted by date ?

models/filtermodel.h
pvuorela: I mean now you have an api that you can use someModel.sortParameter = FilterModel.Date, but which won't be sorted according to time until the method here is explicitly called. So instead the sort mode setter could already trigger the new sorting.

models/documentlistmodel.cpp
```diff
@@ -78,6 +80,8 @@ QVariant DocumentListModel::data(const QModelIndex &index, int role) const
         return d->entries.at(index.row()).mimeType;
     case FileDocumentClass:
         return d->entries.at(index.row()).documentClass;
+    case FileDateRole:
+        return d->entries.at(index.row()).fileRead;
```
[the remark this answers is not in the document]
pvuorela: Also seems strange that there are two roles returning the same thing now.

What comes to lastAccessDateTime, or atime, type of properties, they shouldn't really ever be used. File systems are mounted as relatime or noatime for a reason. Think we should change this to last modified time.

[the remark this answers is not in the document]
dcaliste: Well, in fact, it's coming from tracker database values, not directly read from file system info. And tracker is advertizing this as "lastAccessed"… Whatever this last access time means in term of file system (as you said, it's pretty sure it's mtime and not atime).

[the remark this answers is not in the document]
dcaliste: Besides, you're right, the "ReadRole" already exist and should be used instead of adding a new one (and renamed also in my opinion).

models/filtermodel.cpp
mphilippov-omp: Done

qml/SortTypeSelectionDialog.qml
mphilippov-omp: Done

qml/SortTypeSelectionDialog.qml
mphilippov-omp: Done

models/filtermodel.h
mphilippov-omp: Done

models/documentlistmodel.cpp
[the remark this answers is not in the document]
mphilippov-omp: Done

models/filtermodel.cpp
mphilippov-omp: Modified

models/filtermodel.cpp
mphilippov-omp: Modified

mphilippov-omp: What do you mean? Sorting by date by default?

models/documentlistmodel.h
```diff
@@ -1,4 +1,5 @@
 /*
+ * Copyright (C) 2019 Open Mobile Platform LLC
```
pvuorela: No changes to this and above file.

models/filtermodel.cpp
```diff
@@ -35,6 +37,34 @@ void FilterModel::setSourceModel(DocumentListModel *model)
...
 }
 
+int FilterModel::sortParameter() const
+{
+    return m_sortParameter;
+}
+
+void FilterModel::setSortParameter(int sortParameter)
+{
+    m_sortParameter = sortParameter;
```
pvuorela: For minor detail this could have the common change check, if (m_sortParameter == sortParameter) return;

qml/FileListPage.qml
```diff
@@ -146,6 +159,20 @@ Page {
                                            : qsTrId("sailfish-office-me-hide_search")
                 onClicked: page.searchEnabled = !page.searchEnabled
             }
+            
+            MenuItem {
+                //% "Sort by: %1"
+                text: qsTrId("sailfish-office-me-sort_by").arg(getSortParameterName(filteredModel.sortParameter))
```
pvuorela: By default you now get "Sort by:" without any actual sorting noted. Should be always sorted by something.

qml/SortTypeSelectionPage.qml
```diff
@@ -0,0 +1,69 @@
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+//import Sailfish.Office 1.0
```
pvuorela: Remove if not needed.

qml/SortTypeSelectionPage.qml
```diff
@@ -0,0 +1,69 @@
...
+        console.warn("Unknown sort type")
+        return ""
+    }
+
+    SilicaListView {
+        anchors.fill: parent
+        model: sortModel
+
+        header: PageHeader {
+            //% "Sort by:"
```
pvuorela: Remove semicolon, email has it in code but translations skip it. We can just skip here.

qml/SortTypeSelectionPage.qml
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        delegate: BackgroundItem {
+            Label {
+                x: Theme.horizontalPageMargin
+                anchors.verticalCenter: parent.verticalCenter
+                text: sortTypeText(sortType)
+                font.pixelSize: Theme.fontSizeMedium
+                color: highlighted ? Theme.highlightColor : Theme.primaryColor
+                elide: Text.ElideRight
```
pvuorela: Is like this in email, but don't think this would be needed and font size is medium by default.

qml/SortTypeSelectionPage.qml
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            onClicked: root.sortSelected(sortType)
+        }
+        VerticalScrollDecorator {}
+    }
+
+    ListModel {
```
pvuorela: Small detail, but as this doesn't need any modifications to the list, could just do  model: [FilterModel.Name,  FilterModel.Type,  FilterModel.Date] and refer to model as modelData on the delegate.

qml/FileListPage.qml
mphilippov-omp: Ok. As @dcaliste told, I'll sort it by date by default

models/filtermodel.cpp
mphilippov-omp: Ok, I'll do it

models/filtermodel.cpp
mphilippov-omp: Done

qml/FileListPage.qml
mphilippov-omp: Done

qml/SortTypeSelectionPage.qml
mphilippov-omp: Done

models/filtermodel.cpp
mphilippov-omp: Done

models/documentlistmodel.h
mphilippov-omp: Removed

models/filtermodel.cpp
mphilippov-omp: Done

qml/FileListPage.qml
mphilippov-omp: Done

qml/SortTypeSelectionPage.qml
mphilippov-omp: Done

qml/SortTypeSelectionPage.qml
mphilippov-omp: Done

qml/SortTypeSelectionPage.qml
mphilippov-omp: Ok, removed

qml/SortTypeSelectionPage.qml
mphilippov-omp: Done

models/filtermodel.cpp
```diff
@@ -24,6 +25,7 @@ FilterModel::FilterModel(QObject *parent)
     : QSortFilterProxyModel(parent)
 {
     this->setFilterRole(DocumentListModel::Roles::FileNameRole);
+    setSortParameter(SortParameter::Date);
```
pvuorela: Looks like this gets the oldest first. Think it should be newest.

mphilippov-omp: What do you mean?

qml/SortTypeSelectionPage.qml
mphilippov-omp: Done

models/filtermodel.cpp
mphilippov-omp: Ok. I'll do it

mphilippov-omp: Done

models/filtermodel.cpp
```diff
@@ -35,6 +37,40 @@ void FilterModel::setSourceModel(DocumentListModel *model)
...
+    
+    m_sortParameter = sortParameter;
+    Qt::SortOrder order = Qt::AscendingOrder;
+
+    switch (m_sortParameter) {
+    case Name:
+        setSortRole(DocumentListModel::Roles::FileNameRole);
+        break;
+    case Type:
+        setSortRole(DocumentListModel::Roles::FileTypeRole);
```
pvuorela: This also is a bit surprising now. Choosing file type will get big groups for specific extension, but inside the file types the items are sorted, as it appears to me, oldest first. Should get secondary sorting by name.

Alternatively we could first merge the name and date sorting and add this one later.

qml/FileListPage.qml
```diff
@@ -59,6 +60,18 @@ Page {
...
         }
     }
+    
+    function getSortParameterName(parameter) {
+        if (parameter === FilterModel.Name) {
+            return qsTrId("jolla-email-me-sort_name")
+        } else if (parameter === FilterModel.Type) {
+            return qsTrId("jolla-email-me-sort_type")
+        } else if (parameter === FilterModel.Date) {
+            return qsTrId("jolla-email-me-sort_date")
```
pvuorela: Actually could have separate translations for these as the result is used inside a string. Might want to capitalize menu item but not these.

mphilippov-omp: Explain please

models/filtermodel.cpp
mphilippov-omp: Explain please. You told that you want sorting by date in type sorting groups. But why is the  name sorting secondary?

pvuorela: So sorting by name and sorting by date will almost never have items that are considered equal (i.e. name would need to be same for two files etc). While as for file type you will get big groups that are equal by the file type, like all the pdf files, and inside the group there is either random order or order by some secondary property. Now the secondary property seems to be date and that by showing the oldest first.

Switching the secondary order of FileType sorting to be alphabetical would make more sense. The file type part is already getting sorted alphabetically too.

mphilippov-omp: Ok. In documentlistmodel.cpp:72 `d->entries.at(index.row()).fileRead.toString(Qt::ISODate)` is added to the file extension. Will it be better to change it to file name? Or will it be better to create another role?

qml/FileListPage.qml
mphilippov-omp: Done

models/filtermodel.cpp
pvuorela: Right, that's a strange thing to append there. File type should be file type and nothing more.

Actually trying removing the ISODate part I'm getting secondary ordering to be newest first, which is good enough for me.

mphilippov-omp: Ok, I'll create another role so

mphilippov-omp: Done

pvuorela:
One more thing: the name sortings could be better case insensitive.

mphilippov-omp:
> One more thing: the name sortings could be better case insensitive.

Done

pvuorela:
LGTM

PR #148, opened 2019-11-07
[sailfish-office] Use left aligned details page items. Fixes JB#41622

pvuorela:
Also now using full file path as there is more space for it.

FileInfo adjusted a little not to confuse paths and urls.

@jpetrell @saukko


jpetrell:
LGTM

PR #149, opened 2019-11-11
Translate page title

pvuorela:
@jpetrell

### DIFF
--- a/plugin/TextDocumentToCPage.qml
+++ b/plugin/TextDocumentToCPage.qml
@@ -41,28 +41,14 @@ Page {
         delegate: BackgroundItem {
             property bool isCurrentItem: model.contentIndex + 1 == page.document.currentIndex
             Label {
-                anchors {
-                    left: parent.left
-                    leftMargin: Theme.horizontalPageMargin + (Theme.paddingLarge * model.level)
-                    right: pageNumberLabel.left
-                    rightMargin: Theme.paddingLarge
-                    verticalCenter: parent.verticalCenter
-                }
-                text: model.title
+                x: Theme.horizontalPageMargin
+                width: parent.width - 2*x
+                anchors.verticalCenter: parent.verticalCenter
+                //% "Page %1"
+                text: qsTrId("sailfish_office-la-page_number").arg(model.contentIndex + 1)
                 color: highlighted || isCurrentItem ? Theme.highlightColor : Theme.primaryColor
                 truncationMode: TruncationMode.Fade
             }
-            Label {
-                id: pageNumberLabel
-                anchors {
-                    right: parent.right
-                    rightMargin: Theme.horizontalPageMargin
-                    verticalCenter: parent.verticalCenter
-                }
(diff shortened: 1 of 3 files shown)


pvuorela:
From first commit: 
    Calligra TextContentsModelImpl.cpp translates this with i18n(), being
    only instance under models subdirectory, but we don't ship calligra
    translations. Since the title is nothing more complicated, let's just
    translate the page number ourselves.
    
    Nesting level is also hard-coded 0 at this point.

plugin/TextDocumentToCPage.qml
```diff
@@ -41,28 +41,14 @@ Page {
...
-                    verticalCenter: parent.verticalCenter
-                }
-                text: model.title
+                x: Theme.horizontalPageMargin
+                width: parent.width - 2*x
+                anchors.verticalCenter: parent.verticalCenter
+                //% "Page %1"
+                text: qsTrId("sailfish_office-la-page_number").arg(model.contentIndex + 1)
                 color: highlighted || isCurrentItem ? Theme.highlightColor : Theme.primaryColor
                 truncationMode: TruncationMode.Fade
```
jpetrell: Shouldn't fade. Luckily not much change of that happening with short label and whole display width.

plugin/TextDocumentToCPage.qml
```diff
@@ -41,28 +41,14 @@ Page {
         delegate: BackgroundItem {
             property bool isCurrentItem: model.contentIndex + 1 == page.document.currentIndex
             Label {
-                anchors {
-                    left: parent.left
-                    leftMargin: Theme.horizontalPageMargin + (Theme.paddingLarge * model.level)
```
jpetrell: Dropping hierarchy indentation?

plugin/TextDocumentToCPage.qml
pvuorela: Hm, not sure if I'd call it necessarily wrong here even if we had proper page titles. Anyhow as it now should be always just one word + number I think the simplicity of delegate is better than making it wrap.

jpetrell:
LGTM

plugin/TextDocumentToCPage.qml
jpetrell: UI labels should never fade, user content can fade as long as there is a way to see it in full.

Fading out UI labels is especially problematic on cases where the differentiating part is at the end, the page number.

jpetrell: UI label = translatable system label

pvuorela: Might be translatable, but from user point of view there's no much difference if it was coming from the document itself in this case. If we make wrapping/eliding changes, we should then do it for all the document type table of content pages. 

Then again it should be often possible to view the content, if nothing else then clicking on the item and seeing where it takes.

Granted that the page number as last part here is slightly problematic, but don't think in practice it should ever trigger.

PR #150, opened 2019-11-14
[sailfish-office] Change file read DateTime format after it was chaned in jolla settings. Fixes JB#48127

mphilippov-omp:
1. Change date-time format in jolla-settings from 24-hours to 12-hours format
2. File read DateTime format in file list should be changed too

### DIFF
--- a/qml/FileListPage.qml
+++ b/qml/FileListPage.qml
@@ -21,6 +21,7 @@ import QtQuick 2.0
 import Sailfish.Silica 1.0
 import Sailfish.Office 1.0
 import Sailfish.Office.Files 1.0
+import Nemo.Configuration 1.0
 
 Page {
     id: page
@@ -234,9 +235,18 @@ Page {
                     }
                     Label {
                         anchors.right: parent.right
-                        text: Format.formatDate(model.fileRead, Format.Timepoint)
+                        text: Format.formatDate(model.fileRead, Format.DateMedium)
+                              + " "
+                              + Format.formatDate(model.fileRead, timeFormatConfig.value === "24"
+                                                                  ? Format.TimeValueTwentyFourHours
+                                                                  : Format.TimeValueTwelveHours)
                         font.pixelSize: Theme.fontSizeExtraSmall
                         color: listItem.highlighted ? Theme.secondaryHighlightColor : Theme.secondaryColor
+                    }
+                    
+                    ConfigurationValue {
+                        id: timeFormatConfig
+                        key: "/sailfish/i18n/lc_timeformat24h"
                     }
                 }
             }


qml/FileListPage.qml
```diff
@@ -234,10 +235,19 @@ Page {
                     }
                     Label {
                         anchors.right: parent.right
-                        text: Format.formatDate(model.fileRead, Format.Timepoint)
+                        text: Format.formatDate(model.fileRead, Format.DateMedium)
+                              + " "
+                              + Format.formatDate(model.fileRead, timeFormatConfig.value === "24"
+                                                                  ? Format.TimeValueTwentyFourHours
+                                                                  : Format.TimeValueTwelveHours)
```
pvuorela: Please don't add a second level of time formatting where such is not really needed. High chance that it will end up just worse. 

What you might want here is binding the value to configuration, i.e. change there will trigger re-evalutation, but even that would be sort of a local workaround for a wider problem. The time format configuration shouldn't really be read by random apps.

mphilippov-omp: Actually I don't understand what you mean. I don't change the configuration, I just read it's value. I just want application date time format to match the date time format in configuration

mphilippov-omp: > The time format configuration shouldn't really be read by random apps.

Ok. Do you agree that if you change time format in jolla-settings, this format should be changed in all the applications? I.e if you select 12-hour format, file read date-time should be not 14 Nov 2019 15:56, but 14 Nov 2019 3:56 PM

pvuorela: The bug is real, but the fix for that shouldn't be worked around separately in each app that might be affected. Ideally the formatting would just work, but in general there should be some holistic way to handle these.

mphilippov-omp: > Ideally the formatting would just work, but in general there should be some holistic way to handle these.

Do you mean that there should be a global solution for all the applications?
What is the better solution in your opinion?

pvuorela: > Do you mean that there should be a global solution for all the applications?

Yes.

> What is the better solution in your opinion?

Not sure. Could be a some sort of element that had formatted text as property and proper change signaling. Or could be some javascript wrapper to the function making sure the value gets an implicit binding to config. Or since this often applies to full lists, could be an option getting all the delegates recreated without needing extra handling for each delegate separately.

Bunch of different ways to got forward, needing some though. And then again the whole formatter is a bit ad-hoc, needing to be widely reconsidered.

mphilippov-omp: Is it ok to create a new sailfish-silica component that implements formatting? Or where should it be situated (if not in sailfish-silica)?

pvuorela: Silica would be a natural place, but be warned that all this is box of worms department.

mphilippov-omp: Box of worms? And where should it be better?

pvuorela: Meaning that there can be lots of details involved.

mphilippov-omp: I meant, what is the better place for this new component?

pvuorela: Said that silica would be a natural place.

mphilippov-omp: Ok

PR #152, opened 2019-11-18
[sailfish-office] Display toolbar icons while and after zooming. Contributes JB47527

mphilippov-omp:


pvuorela:
The overlay hiding is there on purpose, compare to e.g. zooming images in gallery. In a way slightly agree that it might not be optimal for bigger spreadsheet not to be able to get the actions shown without zooming back, but then again this way of not doing anything for overlay visibility looks strange too. 

Another place to compare with could be the pdf viewer which has an opaque toolbar that gets visible when scrolling the view down, though that would be slightly bigger UI change for spreadsheet. Maybe other document types too.

cc @jpetrell

PR #156, opened 2019-12-10
[sailfish-office] Fix crash when adding pdf annotation below the document. Fixes JB#48393

pvuorela:
On a very short document it's possible to try adding an annotation
below the document area. Doing that made the app crash.

Now the operation fails silently on the UI side, but should be quite much
better already.

@dcaliste @jpetrell

### DIFF
--- a/pdf/pdfdocument.cpp
+++ b/pdf/pdfdocument.cpp
@@ -174,7 +174,11 @@ void PDFDocument::setAutoSavePath(const QString &filename)
 void PDFDocument::addAnnotation(Poppler::Annotation *annotation, int pageIndex,
                                 bool normalizeSize)
 {
-    d->thread->addAnnotation(annotation, pageIndex, normalizeSize);
+    if (pageIndex < 0) {
+        qWarning() << "Invalid page index for annotation";
+    } else {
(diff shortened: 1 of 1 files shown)


dcaliste:
It's ok as you did. It's fixing the issue.

Looking at the code, I'm wondering if we should modify the `attachOnce()` routine to return `false` when failing, so callers can quick return and not call the failing addAnnotation routine. What do you think ?

pvuorela:
Was also considering making some protection there, maybe more as "if (m_page >= 0) ...->addAnnotation(..)", but there were multiple places calling addAnnotation() so only a single check here seemed simpler.

dcaliste:
> but there were multiple places calling addAnnotation() so only a single check here seemed simpler.

Indeed, this LGTM as is.

In fact, it seems to me that all places addAnnotation() is called, a attachOnce call is done before, allowing to short-circuit by quick return if the latter returns false (in addition to the protection you already added).

pvuorela:
I'll merge this now to get rid of a crash. In longer term we could perhaps consider do we want to either show some notification if an annotation fails, or in this case prevent the dialog from appearing when clicking outside the pdf content.

PR #159, opened 2020-02-13
[office] Fix pdf annotation boundary. Fixes JB#48952

pvuorela:
QRectF::united() is const and the code didn't seem to do much.

@dcaliste @llewelld @jpetrell

### DIFF
--- a/pdf/pdfannotation.cpp
+++ b/pdf/pdfannotation.cpp
@@ -97,7 +97,7 @@ void PDFAnnotation::attach(PDFDocument *document, PDFSelection *selection)
     
     QRectF bounds = word.second;
     for (int i = 1; i < selection->count(); i++)
-        bounds.united(selection->rectAt(i).second);
+        bounds |= selection->rectAt(i).second;
 
     d->m_annotation->setBoundary(bounds);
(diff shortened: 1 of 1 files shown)


pdf/pdfannotation.cpp
```diff
@@ -97,7 +97,7 @@ void PDFAnnotation::attach(PDFDocument *document, PDFSelection *selection)
     
     QRectF bounds = word.second;
     for (int i = 1; i < selection->count(); i++)
-        bounds.united(selection->rectAt(i).second);
+        bounds = bounds.united(selection->rectAt(i).second);
```
dcaliste: I think you can use |= like on line 336.

This `::attach()` is overloaded for the HighlightAnnotation, and in UI, only highlight annotations are calling the `::attach()` method. That's why this bug has been left unnoticed, I think.

dcaliste:
Thanks for the fix by the way.

pdf/pdfannotation.cpp
pvuorela: Oh, I'll switch to that for consistency.

PR #163, opened 2020-05-28
[office] Hide the application background in the homescreen when viewing document pages. Contributes to JB#49705

adenexter:
Use the Page backgroundColor property which is capable of a few more
optimisations instead of the background rectangle in pages.


blammit:
LGTM

PR #167, opened 2020-09-24
[sailfish-office] Display DOC files in application. Fixes JB#51338

mphilippov-omp:


pvuorela:
Could you explain why this would be needed here? We've had plenty of word documents already shown long time already. 

Checked the bug, copied the file to device. Works already without any changes. (Except that it's huge and loading chokes)

mphilippov-omp:
> Could you explain why this would be needed here? We've had plenty of word documents already shown long time already.
> 
> Checked the bug, copied the file to device. Works already without any changes. (Except that it's huge and loading chokes)

Well, I should rename it correctly. *.doc files are not displayed in the list of documents

pvuorela:
.doc files have worked ok for years. If there is some special case that isn't working, I'd be interested in what it is. It doesn't seem to be the one reported.

mphilippov-omp:
> .doc files have worked ok for years. If there is some special case that isn't working, I'd be interested in what it is. It doesn't seem to be the one reported.

Have you checked it? Because *.doc files are not displayed now

dcaliste:
As far as I can tell, it's working on my side:
doc

But maybe it's related to latest tracker version change ? My tracker database of document is old now. But adding a doc file in the `~/Documents` makes it immediately appear in the list.

PR #171, opened 2020-12-09
[sailfish-office] Add notifications permission.

dcaliste:
Notifications are used as information bubble to provide
feedback when adding annotations in PDF files.

Open a PDF and in the toolbar tap on the pen for instance. Doing it should open a minor notification saying that you can highlight text with the finger.

@rainemak and @pvuorela, sorry to be slow ;)


rainemak (approved):

rainemak:
You're fast :+1:

dcaliste:
Well to slow to notice that Notifications was missing also when I noticed that Sharing was not there. Anyway, thanks for the quick merges.

PR #173, opened 2021-01-12
[sailfish-office] Update application and organisation name. Fixes JB#52708

dcaliste:
As discussed from sailfishos/sailjail-permissions#33, I'm changing application name and organisation name for the sailfish-office executable also, so that QML storage can be accessible via sailjail automatically.

I'm not proposing to migrate old database because it's just view settings per document. Not a big deal to loose them during transition.

@Tomin1 and @pvuorela what's your opinion ?


pvuorela:
Works better again. LGTM.

PR #176, opened 2021-01-27
[sailfish-office] Replace internal FileInfo type with one from Nemo.FileManager. Contributes to JB#52722

adenexter:


qml/Main.qml
```diff
@@ -53,7 +54,7 @@ ApplicationWindow {
 
     // file = file or url
     function openFile(file) {
-        fileInfo.source = file
+        fileInfo.url = file
```
rainemak, line 57: Comment says that file or url. Also FileInfo::Private::resolvePath handled relative paths. Is that realistic to assume in this case?

adenexter: The QML engine will resolve a string against the current context base URL when assigning to a URL property and in the current circumstances that will resolve to a file:// URL. Relative paths can't work because they require context, as it is now they will be resolved against the path of the QML file which is a guaranteed nonsense result but so is resolving against the working directory of a non-command line application.

rainemak (approved):

PR #179, opened 2021-04-23
Migrate to Tracker3 

pvuorela:
Switched to libtracker-sparql based driver and implemented a simple graph update api: https://git.sailfishos.org/mer-core/libqtsparql/merge_requests/10

One potentially problematic detail now is that the metadata extraction with Tracker3 seems to start only when needed, i.e. the Documents app launched the first time. At the moment this doesn't recognize the ongoing mining and will say for some time there are no documents on the device. Not sure at the moment what to do with that, but think I'll first adapt the remaining parts to the new tracker, and perhaps could first merge this sort of version anyway.

@llewelld @rainemak @okodron @abranson

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -2,8 +2,6 @@
 #define CONFIG_H
 
 #define QML_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/share/sailfish-office/"
-#define IMAGES_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/share/sailfish-office/images/"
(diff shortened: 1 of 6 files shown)


pvuorela:
Rebased and adjusted the sparql query: now declaring better from which graph the predicate functions get their data, using fileSize from Documents (needs the patch I have on tracker-miners or upstream PR merged), switched lastAccessed to lastModified as that is better available and better supported.

adenexter (approved):

okodron (approved):

abranson (approved):

PR #180, opened 2021-04-26
[sailfish-office] Adapt to sharing UI and API changes. JB#53987 OMP#JOLLA-75

blammit:
The sharing UI is now launched out-of-process.

### DIFF
--- a/plugin/PDFDocumentPage.qml
+++ b/plugin/PDFDocumentPage.qml
@@ -1,5 +1,6 @@
 /*
- * Copyright (C) 2013-2014 Jolla Ltd.
+ * Copyright (C) 2013-2021 Jolla Ltd.
+ * Copyright (C) 2021 Open Mobile Platform LLC.
  * Contact: Robin Burchell <robin.burchell@jolla.com>
  *
  * This program is free software; you can redistribute it and/or
@@ -18,6 +19,7 @@
 
 import QtQuick 2.0
 import Sailfish.Silica 1.0
+import Sailfish.Share 1.0
 import Sailfish.Office.PDF 1.0 as PDF
 import org.nemomobile.configuration 1.0
 import org.nemomobile.notifications 1.0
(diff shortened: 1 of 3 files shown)


blammit:
Note: associated API changes need to be merged before this can be merged.

qml/FileListPage.qml
```diff
@@ -292,8 +293,14 @@ Page {
...
                         text: qsTrId("sailfish-office-la-share")
-                        onClicked: pageStack.animatorPush("Sailfish.TransferEngine.SharePage",
-                                                          { serviceFilter: ["e-mail"], mimeType: model.fileMimeType, source: model.filePath })
+                        onClicked: {
+                            shareAction.fileResources = [model.filePath]
+                            shareAction.trigger()
+                        }
+                        ShareAction {
+                            id: shareAction
+                            capabilitiesFilter: model.fileMimeType
```
chriadam, line 302: It previously used a serviceFilter "e-mail" as well as a mime type filter.  Was this change intentional?  Similarly for the other cases above.

blammit: Yes, it's okay to remove that now as that feature was only used to configure the in-app "add account" feature, but the user will now be sent to the Settings app so it's no longer necessary.

plugin/PDFDocumentPage.qml
```diff
@@ -196,12 +198,12 @@ DocumentPage {
                 //% "Share"
                 text: qsTrId("sailfish-office-me-share")
                 onClicked: {
-                    pageStack.animatorPush("Sailfish.TransferEngine.SharePage",
-                                           {
-                                               "source": page.source,
-                                               "mimeType": page.mimeType,
-                                               "serviceFilter": ["e-mail"]
-                                           })
+                    shareAction.resources = [page.source.toString()]
```
pvuorela, line 201: Seems to work, not fully sure why, but the sharing action by such only handling file paths, not file urls. But should be easy to add more explicit url support, i.e. if it starts with / or file://

pvuorela:
LGTM assuming that the share api officially supports also urls.

plugin/PDFDocumentPage.qml
blammit: Yes this was incorrect. Fixed now.

pvuorela (approved):

okodron (approved):

PR #182, opened 2021-05-18
[sailfish-office] Remove separate .desktop files for MIME type handling. OMP#JOLLA-138 JB#54217

adenexter:
A single file per application makes it possible to inspect the
permissions of the application handling a MIME type without having to do
a secondary resolution to find the file with the permissions.


okodron (approved):

rainemak (approved):
LGTM

PR #186, opened 2022-01-04
Include %license file and adjust naming slightly

pvuorela:
Tracker3 migration switched from (badly working) lastAccessed to lastModified, thus the fileRead/lastRead type of naming doesn't make sense anymore. "Date" matches the sorting option and is more abstract.

@Tomin1 @mlehtima

### DIFF
--- a/models/documentlistmodel.h
+++ b/models/documentlistmodel.h
@@ -45,7 +45,7 @@ public:
         FilePathRole,
         FileTypeRole,
         FileSizeRole,
-        FileReadRole,
+        FileDateRole,
         FileMimeTypeRole,
         FileDocumentClass,
         FileTypeAndNameRole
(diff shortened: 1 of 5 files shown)


models/documentlistmodel.h
```diff
@@ -62,7 +62,8 @@ class DocumentListModel : public QAbstractListModel
     virtual QHash<int, QByteArray> roleNames() const;
 
     void setAllItemsDirty(bool status);
-    void addItem(QString name, QString path, QString type, int size, QDateTime lastRead, QString mimeType);
+    void addItem(const QString &name, const QString &path, const QString &type, int size, QDateTime date,
```
Tomin1, line 65: Just wondering out loud: could `date` be also const reference?

Tomin1 (approved):
The change looks correct to me, I just left one comment but it probably won't make much of a difference anyway.

models/documentlistmodel.h
pvuorela: QDateTime instances should be small so the cost of the reference should be bigger than making a copy so left it this way on purpose.

PR #187, opened 2022-01-13
Use freedesktop.org API for window activation and ExecDBus for D-Bus activation

Tomin1:
Eat our own dog food.

Requires https://github.com/sailfishos/lipstick/pull/17 and lipstick-jolla-home change to work.


dbusadaptor.cpp
```diff
@@ -22,19 +21,27 @@
...
+    , m_view(view)
 {
 }
 
 DBusAdaptor::~DBusAdaptor()
 {
 }
 
-void DBusAdaptor::openFile(const QStringList &files)
+void DBusAdaptor::Open(const QStringList &uris, const QVariantMap &platformData)
```
pvuorela, line 33: Can be done separately, but pondering would it make sense to offer some common component for this adaptor.

dbusadaptor.cpp
```diff
@@ -22,19 +21,27 @@
...
 {
-    if (files.count() > 0) {
-        QMetaObject::invokeMethod(m_view->rootObject(), "openFile", Q_ARG(QVariant, files.at(0)));
+    Q_UNUSED(platformData)
+    if (!uris.isEmpty()) {
+        QMetaObject::invokeMethod(m_view->rootObject(), "openFile", Q_ARG(QVariant, uris.at(0)));
     } else {
-        QMetaObject::invokeMethod(m_view->rootObject(), "activate");
+        Activate(platformData);
```
spiiroin, line 39: As platformData is used / passed onwards from here, it is not really Q_UNUSED, or?

spiiroin (approved):

dbusadaptor.cpp
Tomin1: It was forgotten there. Originally these had separate implementations but I thought that calling `Activate` would make sense here. I'll remove that `Q_UNUSED` from this function. Well spotted. :+1:

spiiroin (approved):

pvuorela (approved):

PR #190, opened 2022-02-22
[sailfish-office] Add missing pdfannotation ctor implementation. Fixes JB#57554

pvuorela:
The plain qobject ctor is defined but not implemented. If this was
uncreatable type it could be skipped, but then the
"property PDFAnnotation annotation" type of declarations in qml side
wouldn't work.

There was also partial m_annotation null protection, but not everywhere.

@Tomin1 @rainemak  @dcaliste

### DIFF
--- a/pdf/pdfannotation.cpp
+++ b/pdf/pdfannotation.cpp
@@ -42,6 +42,11 @@ public:
     int m_page;
 };
 
+PDFAnnotation::PDFAnnotation(QObject *parent)
+    : QObject(parent), d(new PDFAnnotation::Private(nullptr))
+{
+}
+
 PDFAnnotation::PDFAnnotation(Poppler::Annotation *annotation, QObject *parent)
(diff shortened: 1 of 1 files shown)


pdf/pdfannotation.cpp
```diff
@@ -89,6 +97,9 @@ void PDFAnnotation::attachOnce(PDFDocument *document, int page)
 
 void PDFAnnotation::attach(PDFDocument *document, PDFSelection *selection)
```
Tomin1, line 98: This seems to guard against most things but what if `d->m_document == nullptr && document == nullptr`? I think this would then call `addAnnotation()` on `nullptr`.

Tomin1 (approved):
Just one comment about something loosely related. Otherwise LGTM though.

pdf/pdfannotation.cpp
pvuorela: I'll add null check for the document parameter too.

